PR #186, opened 2025-12-05
docs: Poseidon random oracle assumption (#119)

### LINKED ISSUE #119: Poseidon as random oracle documentation
Transcript: everything is hashed through a _single_ hash function over `Fp`. We’ve discussed that we may end up relying entirely on Poseidon modeled as a random oracle, rather than merely assuming collision resistance, since Poseidon is just as secure as a random oracle. This security assumption should be documented.

Additionally, the approach we take for hashing isn’t a _standard_ sponge-based transcript: we rely on a mixture of Pedersen hashing (eg. nested commitments are Pedersen commitments which get absorbed into the transcript) and Poseidon, and effectively use the Poseidon permutation as a fixed-length hash rather than as a full sponge construction. This should also be documented explicitly. _Update:_ this is section is outdated.

LucidSamuel:
docs: document Poseidon random oracle assumption and transcript construction

  - Add in-code documentation to Sponge struct explaining random oracle model
  - Document Pedersen + Poseidon mixture in transcript construction
  - Clarify fixed-length hash usage vs full sponge construction
  - Add security assumptions to primitives guide in book

  Fixes #119

### DIFF
--- a/crates/ragu_circuits/src/mesh.rs
+++ b/crates/ragu_circuits/src/mesh.rs
@@ -301,15 +301,10 @@ impl<F: PrimeField, R: Rank> Mesh<'_, F, R> {
 
     /// Compute a digest of this mesh.
     ///
-    /// This function demonstrates Ragu's use of Poseidon as a **fixed-length hash**
-    /// rather than a full sponge construction. The mesh polynomial is evaluated at
-    /// multiple challenge points, and these evaluations are absorbed into the sponge.
-    /// New challenges are then squeezed out and used for subsequent evaluations,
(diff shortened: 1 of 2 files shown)


crates/ragu_circuits/src/mesh.rs
```diff
@@ -300,6 +300,16 @@ impl<F: PrimeField, R: Rank> Mesh<'_, F, R> {
     }
 
     /// Compute a digest of this mesh.
+    ///
+    /// This function demonstrates Ragu's use of Poseidon as a **fixed-length hash**
```
TalDerei: https://github.com/tachyon-zcash/ragu/pull/186#discussion_r2673112770

PR #320, opened 2026-01-07
docs(book): user guides -> add PCD proofs documentation 

LucidSamuel:
Expands `book/src/implementation/proofs.md` from a stub 
  - Covers the `Pcd` type, `Step` trait, proof creation methods (`seed`, `fuse`, `carry`)
  - Documents verification, rerandomization, and the unified accumulator structure
  - Explains compressed vs uncompressed proof tradeoffs
  - Includes internal proof structure reference table


ebfull (approved):
eagerly merging thanks to being additive :)

PR #325, opened 2026-01-14
feat: utility to create polynomial from a set of roots

S1nus:
~Accumulator based on Pedersen vector commitments to the coefficients of a polynomial whose roots are the values in the accumulator. Membership and non-membership checks are achieved by testing whether or not f(x) == 0 where x is the value.~

Utility for creating a polynomial from a set of roots

### DIFF
--- a/crates/ragu_arithmetic/src/util.rs
+++ b/crates/ragu_arithmetic/src/util.rs
@@ -264,7 +264,15 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
     sum
 }
 
-#[allow(dead_code)]
+/// Computes the lowest degree monic polynomial
+///
+/// $$
+/// \prod_{i=0}^{n-1} (X - r_i)
+/// $$
+///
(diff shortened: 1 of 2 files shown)


crates/ragu_arithmetic/src/util.rs
```diff
@@ -262,6 +264,84 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
...
+            let last_idx = polys.len() - 1;
+            if pairs < last_idx {
+                polys.swap(pairs, last_idx);
+            }
+        }
+
+        polys.truncate(pairs + if has_odd { 1 } else { 0 });
+    }
+
+    polys.into_iter().next().unwrap()
```
TalDerei: prefer expect statement over unwrapping 

```ts
.pop().expect("polys contains exactly one element after tree reduction")
```

TalDerei (commented):
some pedantic observations, but the core algorithm seems sound.

ebfull (approved):
Good enough for now, this can be optimized significantly later. We should have benchmarks when we do so.

PR #326, opened 2026-01-14
perf: batch eval on structured polys with precomputed powers

alxiong:
This PR implements the obvious optimization: accelerate batch evaluation on polys with precomputed z powers.
(leaving the single `eval()` untouched, only improving the hot paths of the multiple polys evaluated at `x` and `xz`)

Major changes include:

- copy the `maybe_rayon` impl over to support optionally parallel impl, gated behind the **new `parallel` feature flag**.
- add `Rank::powers(z)` for precomputing all the z powers
- add `structured::Polynomial::batch_eval()` to leverage the precompute tables amortized over a batch of evals
- update the call sides 11 polys at `x` and `xz`, then 11 polys evaluated at `u` to use `batch_eval()` instead.

This is low-hanging fruits with clear advantages, there are future optimization that may need benchmark to justify. (such as SIMD support etc)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,6 +77,37 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c74b8349d32d297c9134b8c88677813a227df8f779daa29bfc29c183fe3dca6"
 
 [[package]]
+name = "crossbeam-deque"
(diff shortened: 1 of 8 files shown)


alxiong:
superseded by #366

PR #352, opened 2026-01-17
Improve s polynomial doc

alxiong:
Changes include:

- explain the `Collector` (renamed to `Evaluator`) as `Driver` that interprets the circuit as polynomial evaluation in the module doc, both in `s` and in `sx` and in `sy`
- Rename `Wire` to `Monomial` (and their sums `WireSum` to `MonomialSum`) to signify the actual semantic meaning in this context of collector/evaluator driver
- Rename `sx::Collector` to `sx::Evaluator` (similarly for `sxy::Collector`)
- ~~Maximally reuse `sx::Evaluator` logic for `sxy::eval`, greatly simplify the repeated logic~~
- Add doc to `sy`, 
- avoid `mem::forget` during `TermCollector::add_term()` by introducing a `Wire::increment_refcount()`
- simplify `VirtualTable::alloc()` (avoid returning `terms` since it's always `vec![]` empty vector)

### DIFF
--- a/crates/ragu_arithmetic/src/coeff.rs
+++ b/crates/ragu_arithmetic/src/coeff.rs
@@ -7,7 +7,7 @@ use core::ops::{Add, Mul};
 /// about field elements that can be leveraged to optimize arithmetic operations
 /// (like avoiding unnecessary field multiplications) and improve the efficiency
 /// of group arithmetic.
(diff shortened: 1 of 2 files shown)


alxiong:
superseded by #365

PR #357, opened 2026-01-19
circuits: Introduce DriverExt::enforce_public_output and enforce_one

alxiong:
Motivation: 
When I look at code like 

```rust
    for output in outputs {
        dr.enforce_zero(|lc| lc.add(output.wire()))?;
    }
    dr.enforce_zero(|lc| lc.add(&one))?;
```
This is quite unintuitive, how could `ONE` wire equals to zero? same for the public output? 
What we are actually enforcing is that `j`-th output is equal to the `k_j` defined in the `k(y)` vectors. 
So the dual semantic of `enforce_zero()` for `s::*` drivers need some clarification.

Even though this is just internal code, I think some clarification could be helpful. Here's what I propose:

```rust
// private to `ragu_circuit`, only used by s drivers
pub(crate) trait DriverExt<'dr>: Driver<'dr> {
    fn enforce_public_output(
        &mut self,
    ) -> Result<()>;
    fn enforce_one(&mut self, lc: impl Fn(Self::LCenforce) -> Self::LCenforce) -> Result<()>;
}

// so that we clear intention

    // Bind circuit witness outputs to k(Y) coefficients k_j
    for output in outputs {
        dr.enforce_public_output(|lc| lc.add(output.wire()))?;
    }
    // Bind constant ONE wire to k_0 (=1)
    dr.enforce_one(|lc| lc.add(&one))?;
```

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -170,3 +170,51 @@ pub trait CircuitObject<F: Field, R: Rank>: Send + Sync {
     /// Returns the number of constraints: `(multiplication, linear)`.
     fn constraint_counts(&self) -> (usize, usize);
 }
+
+/// Extension trait for drivers for internal use, **NOT for circuit implementors**.
+///
+/// User-defined circuits takes `dr: Driver` to specify internal constraints,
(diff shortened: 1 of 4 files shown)


alxiong:
superseded by #364

PR #358, opened 2026-01-19
nit: avoid unwrap in rx::eval()

alxiong:
Changes include:

- Rename `Collector` to `Evaluator` (to be consistent with #352)
- remove `.unwrap()` and return `Err::InvalidMeshKey` instead during `rx::eval()`


alxiong:
superseded by #363

PR #363, opened 2026-01-19
Introduce registry key struct

alxiong:
Changes include:

- Rename `Collector` to `Evaluator` (to be consistent with #352)
- Add `struct registry::Key` whose constructor ensures non-zero value and cache its inverse
- Update all apis that use to accept an plain `F`, now updated with `&Key` and access the value via `.value()`
- Rename `Registry.get_key()` to `Registry.key()`

### DIFF
--- a/crates/ragu_circuits/src/rx.rs
+++ b/crates/ragu_circuits/src/rx.rs
@@ -127,8 +127,8 @@ mod tests {
     fn test_rx() {
         let circuit = SquareCircuit { times: 10 };
         let witness: Fp = Fp::from(3);
-        let key = Fp::ONE;
-        let (rx, _aux) = eval::<Fp, _, R<6>>(&circuit, witness, key).unwrap();
+        let key = registry::Key::default();
+        let (rx, _aux) = eval::<Fp, _, R<6>>(&circuit, witness, &key).unwrap();
         let mut coeffs = rx.iter_coeffs().collect::<Vec<_>>();
         let size_of_vec = coeffs.len() / 4;
         let c = coeffs.drain(..size_of_vec).collect::<Vec<_>>();


ebfull (changes requested):
Instead, along the lines of [another observation](https://github.com/tachyon-zcash/ragu/pull/365#discussion_r2709804691), we should have a `MeshKey` type (guaranteed to be nonzero) that is returned by the function that produces it. In fact, this type could contain the inverse within it, as a caching opportunity. This would avoid the inverse and the need to introduce a new error.

In general, I followed this logic when reviewing this PR:

1. I don't like adding error types that cannot be triggered by integration tests; these should be assertions.
2. Well, this unwrap could be triggered by a user who supplies a zero key!
3. But, this API is not even intended for users of ragu to deal with. It may be exported by `ragu_circuits` but it is not something I anticipate `ragu` crate users to need to use directly.
4. So, this implies the error case should not exist, but that instead something should prevent it from ever happening. Thus, let's create a type in the `mesh` module that promises to prevent the behavior.

Further, let's call it `Key` (because `mesh::MeshKey` is redundant.)

crates/ragu_circuits/src/registry.rs
```diff
@@ -137,16 +137,105 @@ impl<'params, F: PrimeField, R: Rank> RegistryBuilder<'params, F, R> {
...
+/// to the transcript before any prover messages, forcing a fixed instance.
+/// However, the registry `m` contains the description of a recursive verifier
+/// whose logic depends on a transcript salted with the very digest
+/// `H(m(W,X,Y))` itself, creating a circular dependency.
+///
+/// Previously, many preprocessing recursive proofs solve this self-reference
+/// problem by externalizing the circuit description into a verification key (vk)
+/// that is generated ahead of time and passed in as public input to the
+/// recursive verifier. However, Ragu avoid preprocessing by design, requiring
+/// an alternative solution.
```
ebfull:
```suggestion
/// Many preprocessing recursive SNARKs avoid this self-reference problem
/// implicitly because the circuit descriptions are encoded in a verification
/// key that is generated ahead of time and carried through public inputs to the
/// recursive verifier. Ragu avoids preprocessing by design, and does not use
/// verification keys, which suggests an alternative solution.
```

PR #364, opened 2026-01-19
circuits: Introduce DriverExt in s::common to explain public input enforcement

alxiong:
Motivation: 
When I look at code like 

```rust
    for output in outputs {
        dr.enforce_zero(|lc| lc.add(output.wire()))?;
    }
    dr.enforce_zero(|lc| lc.add(&one))?;
```
This is quite unintuitive, how could `ONE` wire equals to zero? same for the public output? 
What we are actually enforcing is that `j`-th output is equal to the `k_j` defined in the `k(y)` vectors. 
So the dual semantic of `enforce_zero()` for `s::*` drivers need some clarification.

Even though this is just internal code, I think some clarification could be helpful. Here's what I propose:

```rust
// private to `ragu_circuit`, only used by s drivers
pub(crate) trait DriverExt<'dr>: Driver<'dr> {
    fn enforce_public_output(
        &mut self,
    ) -> Result<()>;
    fn enforce_one(&mut self, lc: impl Fn(Self::LCenforce) -> Self::LCenforce) -> Result<()>;
}

// so that we clear intention

    // Bind circuit witness outputs to k(Y) coefficients k_j
    for output in outputs {
        dr.enforce_public_output(|lc| lc.add(output.wire()))?;
    }
    // Bind constant ONE wire to k_0 (=1)
    dr.enforce_one(|lc| lc.add(&one))?;
```


ebfull:
These are just details in the `s` module implementation for the _special_ linear constraints used to enforce public inputs.

However, I'd be fine with adding this `Driver` extension trait to the `common` submodule inside of `s`. I think it's a good opportunity to be clear about what's happening and document it. So, ACK from me modulo moving things around a bit.

alxiong:
> documenting that more explicitly might suffice.

I took @TalDerei 's suggestion and only add doc instead. Because on a second thought, adding an extra traits bring mental baggage to internal-only code with minuscule benefit

crates/ragu_circuits/src/s/common.rs
```diff
@@ -21,11 +21,11 @@
...
-//! The `ONE` wire corresponds to the $c$ wire from gate 0, with monomial $x^{4n
-//!     - 1}$. Since [`Driver::ONE`] must be a compile-time constant, it cannot
-//!     hold this computed value. Instead, [`WireEval::One`] serves as a
-//!     sentinel that [`WireEvalSum::add_term`] resolves to the cached $x^{4n -
-//!     1}$ value at runtime.
+//! The `ONE` wire corresponds to the $c$ wire from gate 0, with monomial
+//! $x^{4n-1}$. Since [`Driver::ONE`] must be a compile-time constant, it cannot
+//! hold this computed value. Instead, [`WireEval::One`] serves as a sentinel
+//! that [`WireEvalSum::add_term`] resolves to the cached $x^{4n - 1}$ value
+//! at runtime.
```
ebfull: Thanks for fixing this.

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -287,6 +287,26 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<F, R> {
...
+///
+/// However, the public output gadget (returned by [`Circuit::instance`]) and
+/// the `ONE` wire are treated specially: their corresponding [`enforce_zero`]
+/// calls do not enforce that the wire equals zero. Instead, they create binding
+/// constraints that force these wires to match their corresponding values in the
+/// public input polynomial $k(Y)$.
+///
+/// For example, `enforce_zero(|lc| lc.add(output_wire))` on a public output
+/// doesn't constrain `output_wire = 0`. Instead, if `output_wire` has value `v`,
+/// this creates a constraint binding `v` as a coefficient in $k(Y)$.
```
ebfull: The actual way I would describe this to somebody, who is curious about the behavior mathematically at this level, is along the lines:

> Linear constraints are presented to users via the `enforce_zero` API, which constrains a linear combination of wire assignments to equal zero. But _algebraically_ each constraint actually relates linear combinations of wires to a (sparse) public input assignment behind the scenes. As such, the `enforce_zero` API is what is exposed via the `Driver` trait, but the driver handles the linear constraints that relate to nontrivial public input assignments via the same mechanism.

That's just a really crappy paragraph I wrote up, you could throw this at an LLM and find a better or more thorough way to explain it.

ebfull (approved):
I liked the original approach but I actually don't care either way. These are non-blocking comments/suggestions.

ebfull (approved):

PR #369, opened 2026-01-20
chore: merge-gating on codeowner file

TalDerei:
added to prevent accidental pushes to main on another reviewers approval that isn't a codeowner. Requires commensurate branch protections to be configured.


TalDerei:
Screenshot 2026-01-20 at 12 18 04 PM

PR #362, opened 2026-01-19
circuits: rename StageObject to StageMask

alxiong:
Changes include:

- Rename `StageObject` to `StageMask`, constructed via `S::mask()` or `S::final_mask()` 
- update all tests, docs
  - meanwhile fix a small error in the doc of `staging/mod.rs`

Issue #205, opened 2025-12-08
Clarify our mesh and staging terminology


TalDerei:
Our discussions around staging polynomials have used several overloaded terms. This note is an attempt to clarify the current _status quo_, and these definitions should eventually be simplified and make their way into the book and the corresponding code.

Additionally, the proposed table also currently conflates the user API with the internal protocol mechanics, and was playing around with this decomposition (requires other tables). There's also potential here for simplification by not exposing the user to the concept of a stage and having the API reflect that.

------------

<img width="534" height="392" alt="Image" src="https://github.com/user-attachments/assets/14097c63-0c44-4f05-bb30-76dbfd83b39c" />

<img width="516" height="402" alt="Image" src="https://github.com/user-attachments/assets/1b1cbb49-293b-4875-90de-0fb8bf4395d0" />

------------

PR #370, opened 2026-01-20
chore: mesh, staging, and circuit terminology

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/205 and extends https://github.com/tachyon-zcash/ragu/pull/362. I took the liberty of suggesting some naming changes, but we can refine them through discussion. Renames `Mesh` to `Registry`, `StagedCircuit` to `MultiStageCircuit`, and `circuit` polynomial to `wiring` polynomial.

### DIFF
--- a/book/src/appendix/terminology.md
+++ b/book/src/appendix/terminology.md
@@ -6,7 +6,7 @@
 |---------|------|----------------|
 | Defines stage structure; specifies wire range and corresponds with a stage polynomial that represents a partial witness | **Stage** | `preamble::Stage` |
 | Input data for a stage | **Stage witness** | `Stage::witness()` |
-| Well-formedness check for a stage; $s$ polynomial enforcing linear independence | **Stage mask** | `Stage::mask()` or `Stage::final_mask()` |
+| Well-formedness check for a stage; $s$ polynomial enforcing linear independence (for multi-stage circuits, these masks are batched in the revdot check) | **Stage mask** | `Stage::mask()` or `Stage::final_mask()` |
 | Circuit using staged witnesses | **Multi-stage circuit** | `MultiStageCircuit` |
 | Combined witness across all stages | **Multi-stage witness** | implicit, concatenation of stage witness |
 | Combined $r(X) = a(X) + b(X) + \cdots + f(X)$ | **Multi-stage witness polynomial $r(X)$** | implicit, sum of all `Stage::rx()` |


crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -90,18 +90,18 @@
 //!
 //! ### Final Stage
 //!
-//! The final witness circuit is implemented using the [`StagedCircuit`] trait,
+//! The final witness circuit is implemented using the [`MultiStageCircuit`] trait,
```
alxiong: "final witness" has confusion. I'd use "overall witness for the multistage circuit", because it can be mistaken as the witness for the final stage.

crates/ragu_circuits/src/staging/mod.rs
alxiong: I'd like to see `StageBuilder<.., Self::First, Self::Final>` instead of using the `()` as the start stage. Imo, the parent of the first stage being `()` should be an internal detail, but current type signature leak that.

I know this comment is less relevant to this renaming PR, but would like to discuss and see if we want to refactor this in a separate PR.

cc @ebfull

book/src/assets/terminology-mapping.png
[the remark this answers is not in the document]
ebfull: I think "final" is fine.

book/src/appendix/terminology.md
```diff
@@ -1 +1,13 @@
...
+## Registry, Stage, and Circuit Terminology Mapping
+
+| Concept | Term | Code Reference |
+|---------|------|----------------|
+| Defines stage structure; specifies wire range and corresponds with a stage polynomial that represents a partial witness | **Stage** | `preamble::Stage` |
+| Input data for a stage | **Stage witness** | `Stage::witness()` |
+| Well-formedness check for a stage; $s$ polynomial enforcing linear independence | **Stage mask** | `Stage::mask()` or `Stage::final_mask()` |
+| Circuit using staged witnesses | **Multi-stage circuit** | `MultiStageCircuit` |
+| Combined witness across all stages | **Multi-stage witness** | implicit, concatenation of stage witness |
+| Combined $r(X) = a(X) + b(X) + \cdots + f(X)$ | **Multi-stage witness polynomial $r(X)$** | implicit, sum of all `Stage::rx()` |
```
alxiong: I'd suggest adding a row below for the "multi-stage wiring polynomial" in https://github.com/tachyon-zcash/ragu/pull/370#discussion_r2711592258, because the revdot check on the wiring for multistage circuit is different from that for normal circuit, right?

alxiong (commented):
LGTM! just one more small request.

book/src/appendix/terminology.md
TalDerei: thought it was redundant against the stage masks row; we could add "for multi-stage circuits, these masks are batched in the revdot check"

TalDerei: added clarification note

alxiong (approved):

ebfull (approved):

PR #373, opened 2026-01-21
docs: add R1CS concept explainer

turbocrime:
maybe it wasn't useful to spend time on this, given there's pages 'out there' we could link to. but this is shorter than any of those.

LLM outline, rewrote by hand.

some of the phrasing looks redundant if you're familiar. keep in mind i tried to write this for someone who is reading these docs going "hey, i know some of those words" because who else is going to read this R1CS explainer?

need to add citations.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -2,6 +2,7 @@
 
 [Introduction](introduction.md)
 [Proof-Carrying Data](concepts/pcd.md)
+[Rank-1 Constraint Systems](concepts/r1cs.md)
 
 ---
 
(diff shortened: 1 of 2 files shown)


alxiong (commented):
in #214，I also gave a concrete example of R1CS's arithmetization with diagram. (in the appendix)

since we don't use r1cs in our code or protocol, we don't need to elaborate it in our "concept" section. 
imho, adding links to some external materials is sufficient. 

But I'm also open to add some the basic intro (ie first half of this PR until the "example" section) in the same appendix page if others find it helpful.

Issue #47, opened 2025-10-13
Criterion benchmark suite


TalDerei:
Add support for and implement comprehensive benchmarks using the [criterion](https://docs.rs/criterion/latest/criterion/) for arithmetic operations, circuit operations, and other core routines to identify bottlenecks and parallelization potential.

_Update:_ **Iai-Callgrind** benchmarking harness also seems like a viable alternative: https://docs.rs/iai-callgrind/latest/iai_callgrind/

PR #375, opened 2026-01-21
chore: initial benchmarks

turbocrime:
Initial work towards #47. Adds Criterion benchmarks for `ragu_pcd`, `ragu_primitives`, and `ragu_circuits`.

Uses Criterion instead of callgrind due to [lack of macOS ARM support](https://github.com/tachyon-zcash/ragu/pull/326#issuecomment-3756603604) in callgrind.

Benchmarks use a mock rng for consistency.

### PCD

- build
- seed
- fuse
- verify
- rerandomize

### Primitives

Selected an emulator mode according to agent suggestions but it would be good to check that it's a useful driver to bench.

- Element: mul, square, invert, fold
- Point: alloc, double, add_incomplete, double_and_add_incomplete
- Boolean: alloc, and, conditional_select, multipack
- Sponge: absorb, squeeze

### Circuits

- structured polynomial commits
- unstructured polynomial commits
- revdot
- fold
- eval
- dilate

### DIFF
--- a/crates/ragu_circuits/benches/circuits.rs
+++ b/crates/ragu_circuits/benches/circuits.rs
@@ -2,8 +2,8 @@ use arithmetic::Cycle;
 use criterion::{Criterion, criterion_group, criterion_main};
 use ff::Field;
 use ragu_circuits::CircuitExt;
-use ragu_circuits::mesh::MeshBuilder;
 use ragu_circuits::polynomials::{R, structured, unstructured};
+use ragu_circuits::registry::RegistryBuilder;
 use ragu_circuits::test_fixtures::{MySimpleCircuit, SquareCircuit};
 use ragu_pasta::{Fp, Pasta};
 use rand::rngs::mock::StepRng;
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,81 @@
+name: Benchmarks
```
TalDerei: CI currently failing compilation

.github/workflows/benchmark.yml
TalDerei: can we rename to `bench.yml` and emulate some kind of [alert threshold](https://github.com/zcash/halo2/blob/main/.github/workflows/bench.yml#L30-L31)?

crates/ragu_primitives/benches/primitives.rs
TalDerei: would also consider benches in `arithmetic` crate for proving hot paths like like `msm`, `fft`,` batch_invert`.

.github/workflows/benchmark.yml
turbocrime: documentation indicated the action should support an absent baseline...

i tested locally with `act` and discovered the action requires at least the path to the baseline to exist, in order to execute.

fixed now.

crates/ragu_circuits/Cargo.toml
```diff
@@ -21,15 +21,22 @@ all-features = true
 
 [features]
 default = []
+test-fixtures = ["ragu_pasta"]
```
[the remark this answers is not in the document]
turbocrime: benches run in `release` mode and don't have access to `#[test]` annotated blocks or dev-dependencies. in ragu_circuits, ragu_pasta was not an existing production dependency.

i think a flag is the simplest way to share fixtures without adding content to releases. by contrast penumbra has dedicated test and bench crates, and makes test fixtures publicly available from libs.

if publishing new crates requires some discussion and planning, then i think that should be a separate issue and a separate PR, to avoid bikeshedding the initial bench suite right now.

if you want to avoid a feature flag, maybe we could use a cfg flag.

.github/workflows/benchmark.yml
turbocrime: renamed to bench.yml

crates/ragu_primitives/benches/primitives.rs
turbocrime: added arithmetic benches

PR #385, opened 2026-01-24
Optimize multi-point evaluation in query 

### LINKED ISSUE #334: Optimize multi-point evaluation via FFT
References https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_pcd/src/fuse/_07_query.rs#L65-L66 in `_07_query.rs` where we perform 13 separate evaluations of `mesh_xy_poly` at roots of unity. The optimization would be to use a single FFT to compute all evaluations in `O(n log n)`, then extract the needed values.

Vishalkulkarni45:
Cache circuit evaluations

This PR fixes #334. Optimize proof query phase by computing all circuit evaluations $s_i(x,y)$ during polynomial interpolation, avoiding 13 redundant O(n) evaluations during query. 

Changes:
- Add RegistryAtXY struct with cached circuit_evals and poly
- Add xy_with_evals() method for combined computation
- Update `_07_query.rs` to use circuit_eval()

### DIFF
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -386,16 +386,7 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
 
     /// Evaluate the registry polynomial unrestricted at $W$.
     pub fn xy(&self, x: F, y: F) -> unstructured::Polynomial<F, R> {
-        let mut coeffs = unstructured::Polynomial::default();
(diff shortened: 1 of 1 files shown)


ebfull:
Let's close this PR if it's not rebased in the next 24 hours. This isn't a priority optimization.

ebfull:
I'm going to close this because we're currently modifying quite a bit of this part of the code, and it's unclear to me exactly how the APIs will work. We'll revisit #334 later.

Issue #291, opened 2026-01-01
Staging enforcement adds redundant equality constraints


TalDerei:
For stages that consist mostly of `Elements` (preamble ~141 wires, query ~67, eval ~36), calling `.enforced()` on the stage adds wasted allocations (multiplication gates) and linear constraints per circuit, since `Element` gadgets have no internal constraints to verify. Currently, a workaround is to use `.unenforced()` for `Element` _only_ stages, but what if they contain both `Elements` and `Points`? Ideally, we’d like to call `.enforced()` on the stage and still avoid paying the extra constraint cost for `Elements`.

Some solutions could be:

1. Type-level marker trait `TrivialGadget` for gadgets that should skip the equality constraint, and modify the enforcement injector (`convert_wire()`) to check for this trait. Alternatively, we could play around with the derive macro in [map_gadget](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_macros/src/derive/gadget.rs#L258), detect `TrivialGadget` and take a non-enforcing path.

2. Currently, enforcement path is split at the wire level, not gadget level – we could change that and add enforce method on `Gadget` trait, with each gadget defining its own enforcement behavior.

3. We could separate gadget kinds in the staging API itself, and have like a `type EnforcedOutput: GadgetKind<F>;` for `Points` and more complex gadget kinds, and type `ValueOutput: GadgetKind<F>;` for `Elements`. And obviously the `StageGuard` would handle them differently by splitting stage wires between enforced and value parts.

PR #388, opened 2026-01-25
feat(staging): efficient stage enforcement

TalDerei:
already had a local branch experimenting with this, but modified to align closer to https://github.com/tachyon-zcash/ragu/pull/383#issuecomment-3795810887 and closes https://github.com/tachyon-zcash/ragu/issues/291. 

previously, we enforced equality between each fresh wire and stage wire individually. Now we implement a `Consistent` trait and macro with `enforce_consistent()` method to move from wire-level enforcement to gadget-level enforcement. `enforced()` now injects stage wire directly, and then calls `enforce_consistent()` (which as a side-effect eliminates redundant wire allocation/equality constraints for certain gadget invocations).

### DIFF
--- a/crates/ragu_pcd/src/circuits/native/partial_collapse.rs
+++ b/crates/ragu_pcd/src/circuits/native/partial_collapse.rs
@@ -33,8 +33,8 @@
 //! This circuit uses [`error_m`] as its final stage, which inherits in the
 //! following chain:
 //! - [`preamble`] (unenforced)
-//! - [`error_n`] (unenforced)
-//! - [`error_m`] (unenforced)
+//! - [`error_n`] (enforced)
+//! - [`error_m`] (enforced)
 //!
 //! ## Public Inputs
 //!
@@ -152,8 +152,8 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
             builder.add_stage::<native_error_m::Stage<C, R, HEADER_SIZE, FP>>()?;
         let dr = builder.finish();
         let preamble = preamble.unenforced(dr, witness.view().map(|w| w.preamble_witness))?;
-        let error_n = error_n.unenforced(dr, witness.view().map(|w| w.error_n_witness))?;
-        let error_m = error_m.unenforced(dr, witness.view().map(|w| w.error_m_witness))?;
+        let error_n = error_n.enforced(dr, witness.view().map(|w| w.error_n_witness))?;
+        let error_m = error_m.enforced(dr, witness.view().map(|w| w.error_m_witness))?;
 
         let unified_instance = &witness.view().map(|w| w.unified_instance);
         let mut unified_output = OutputBuilder::new();
--- a/crates/ragu_pcd/src/circuits/native/stages/error_m.rs
+++ b/crates/ragu_pcd/src/circuits/native/stages/error_m.rs
@@ -7,7 +7,7 @@ use ragu_circuits::{polynomials::Rank, staging};
 use ragu_core::{
     Result,
     drivers::{Driver, DriverValue},
-    gadgets::{Gadget, GadgetKind, Kind},
+    gadgets::{Consistent, Gadget, GadgetKind, Kind},
     maybe::Maybe,
 };
 use ragu_primitives::{
@@ -31,7 +31,7 @@ pub struct Witness<C: Cycle, FP: fold_revdot::Parameters> {
 }
 
 /// Output gadget for the error_m stage.
-#[derive(Gadget)]
+#[derive(Gadget, Consistent)]
 pub struct Output<'dr, D: Driver<'dr>, FP: fold_revdot::Parameters> {
     /// Error term elements for layer 1.
     /// Outer: N claims, Inner: M²-M error terms per claim.
--- a/crates/ragu_pcd/src/circuits/native/stages/error_n.rs
+++ b/crates/ragu_pcd/src/circuits/native/stages/error_n.rs
@@ -7,7 +7,7 @@ use ragu_circuits::{polynomials::Rank, staging};
 use ragu_core::{
     Result,
     drivers::{Driver, DriverValue},
-    gadgets::{Gadget, GadgetKind, Kind},
+    gadgets::{Consistent, Gadget, GadgetKind, Kind},
     maybe::Maybe,
 };
 use ragu_primitives::{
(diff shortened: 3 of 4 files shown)


crates/ragu_circuits/src/staging/builder.rs
```diff
@@ -99,23 +77,36 @@ pub struct StageGuard<'dr, D: Driver<'dr>, R: Rank, S: Stage<D::F, R>> {
...
     ///
-    /// Runs the stage's witness method on the real driver (enforcing all
-    /// internal constraints), then enforces equality between the computed
-    /// wires and the pre-allocated stage wires.
-    pub fn enforced<'a, 'source: 'dr>(
+    /// Runs the stage's witness method on a wireless emulator, substitutes
+    /// the pre-allocated stage wires, then calls `enforce_consistent()` on
+    /// the result to enforce internal constraints (e.g., curve equations
+    /// for Points).
+    pub fn enforced<'source: 'dr>(
```
TalDerei, line 127: basically mimics `unenforced` (injecting stage wires without re-allocation) and then internally calls `enforce_consistent`. the `Preamble` stage is technically the only native stage containing `Points` (via `unified::Output`).

- `error_n: FixedVec<Element>, Element`
- `error_m: FixedVec<FixedVec<Element>>`
- `query: Element, ChildEvaluations`
- `eval: Element, ChildEvaluations`

`endoscalar.rs` still uses `.unenforced()` because their constraints will be enforced by the routing circuits.

crates/ragu_pcd/src/circuits/native/unified.rs
```diff
@@ -150,6 +150,20 @@ pub struct Output<'dr, D: Driver<'dr>, C: Cycle> {
...
+impl<'dr, D: Driver<'dr, F = C::CircuitField>, C: Cycle> Consistent<'dr, D> for Output<'dr, D, C> {
+    fn enforce_consistent(&self, dr: &mut D) -> Result<()> {
+        self.nested_preamble_commitment.enforce_consistent(dr)?;
+        self.nested_s_prime_commitment.enforce_consistent(dr)?;
+        self.nested_error_m_commitment.enforce_consistent(dr)?;
+        self.nested_error_n_commitment.enforce_consistent(dr)?;
+        self.nested_ab_commitment.enforce_consistent(dr)?;
+        self.nested_query_commitment.enforce_consistent(dr)?;
+        self.nested_f_commitment.enforce_consistent(dr)?;
+        self.nested_eval_commitment.enforce_consistent(dr)?;
```
ebfull: This will be risky / dangerous to not have an auto-derive macro for.

TalDerei:
> > Let's add an auto-derive macro that automatically calls to gadget field's `enforce_consistent` but disregards the other fields.
> 
> before the manual impls, initially tried extending `#[derive(Gadget)]` to auto-derive `Consistent` (avoiding another explicit opt-in macro), but the cascading trait bounds were messy, so i'll experiment with a separate `#[derive(Consistent)]` macro.

one of the annoying things here is derive macros can iterate gadget fields, but the `#[derive(Consistent)]` proc macro can't infer that `Point`'s Consistent impl needs `D::F = C::Base`. so types containing points need to specify the bound somehow (attribute or where clause). this applies independent of whether you implement a new macro of extend existing `Gadget`.

crates/ragu_primitives/src/point.rs
```diff
@@ -203,6 +203,21 @@ impl<'dr, D: Driver<'dr, F = C::Base>, C: CurveAffine> Point<'dr, D, C> {
...
+        let x2 = self.x.square(dr)?;
+        let x3 = self.x.mul(dr, &x2)?;
+        let y2 = self.y.square(dr)?;
+
+        dr.enforce_zero(|lc| {
+            lc.add(x3.wire())
+                .add_term(&D::ONE, Coeff::Arbitrary(C::b()))
+                .sub(y2.wire())
+        })
+    }
```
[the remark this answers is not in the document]
ebfull: `Point::alloc` doesn't use `Driver::alloc`

ebfull:
@TalDerei Perhaps this will assist in that particular case: https://github.com/tachyon-zcash/ragu/pull/397

crates/ragu_pcd/src/circuits/native/unified.rs
TalDerei: Implemented explicit `Consistent` auto-derive macro

crates/ragu_core/src/gadgets.rs
```diff
@@ -321,3 +321,25 @@ pub use ragu_macros::Gadget;
...
+/// like `Element`, this is a no-op. For gadgets with internal constraints like
+/// `Point`, this enforces the curve equation using the gadget's existing wires.
+///
+/// This enables efficient staged enforcement by eliminating fresh wire allocations
+/// and equality constraints entirely, making `.enforced()` as efficient as
+/// `.unenforced()` for constraint-free stages while still correctly enforcing
+/// constraints for mixed stages.
+///
+/// Use `#[derive(Consistent)]` to auto-generate implementations that enforce
+/// consistency on all `#[ragu(gadget)]` fields.
```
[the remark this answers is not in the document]
TalDerei: rewrote as a self-contained reference

alxiong (commented):
LGTM! good work on the macro!

just a question.

crates/ragu_pcd/src/circuits/native/partial_collapse.rs
```diff
@@ -152,10 +152,6 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
             builder.add_stage::<native_error_m::Stage<C, R, HEADER_SIZE, FP>>()?;
         let dr = builder.finish();
         let preamble = preamble.unenforced(dr, witness.view().map(|w| w.preamble_witness))?;
-
-        // TODO: these are unenforced for now, because error_n/error_m stages
-        // aren't supposed to contain anything (yet) besides Elements, which
-        // require no enforcement logic. Re-evaluate this in the future.
         let error_n = error_n.unenforced(dr, witness.view().map(|w| w.error_n_witness))?;
         let error_m = error_m.unenforced(dr, witness.view().map(|w| w.error_m_witness))?;
```
ebfull: Actually, what we want to do when removing this TODO comment is make these `enforced` at the same time _in one particular circuit_. Because the reason for the TODO was to remind ourselves that these stages weren't enforced in _any_ circuit because of the performance issue this PR addresses.

ebfull (changes requested):

ebfull (approved):

crates/ragu_pcd/src/circuits/native/partial_collapse.rs
TalDerei: oh yeah forgot these needed changing

PR #392, opened 2026-01-27
pcd: More ergonomic PCD seed and fuse API

alxiong:
Motivation starts with the observation of an awkward and arguably an unintended abuse that escape the compiler check:

```rust
// in pcd/tests/nontrivial.rs

    let leaf1 = app.seed(
        &mut rng,
        WitnessLeaf {
            poseidon_params: Pasta::circuit_poseidon(pasta),
        },
        Fp::from(42u64),
    )?;
    let leaf1 = leaf1.0.carry(leaf1.1);
```

here `leaf1: (Proof, S::Aux)`, but `carry()` expect `Header::Data`, in this step, the aux happens to be the same type of the output data type.

We could add a `Step::aux_to_data()` method and requiring implementators to specify their conversion in case the auxiliary data is really different or more succinct than the output data. But after examining all occurrences, my personal conclusion is that there's currently no material benefit of `Step::Aux`. 
(^^ correct me if I'm wrong)

Thus, here are my changes:
- Remove `Step::Aux`, return output data directly: `Step::witness() -> (.., DriverValue<D, <Self::Output as Header<C::CircuitField>>::Data<'source>>)`
- Simplify `Application::fuse() -> Pcd` instead of returning the aux and forcing the caller to manually `.carry()` them. 
- Let `Rerandomize` step to pass through the left data directly as part of its step, simpler logic in `App::rerandomize()`.

IMHO, this is a more ergonomic API.

### DIFF
--- a/crates/ragu_pcd/src/fuse/_01_application.rs
+++ b/crates/ragu_pcd/src/fuse/_01_application.rs
@@ -3,8 +3,7 @@
 //! This creates a witness for the step circuit given the two input [`Pcd`]s and
 //! the step witness. This produces the [`proof::Application`] component of the
 //! proof. The inputs are all consumed, and the `left` and `right proofs are
-//! returned to the caller along with the auxiliary data from the application
-//! synthesis.
+//! returned to the caller along with the output data from the step circuit.
 
 use arithmetic::Cycle;
 use ff::Field;
@@ -13,7 +12,7 @@ use ragu_core::Result;
 use rand::Rng;
 
 use crate::{
-    Application, Pcd, Proof, proof,
+    Application, Header, Pcd, Proof, proof,
     step::{Step, internal::adapter::Adapter},
 };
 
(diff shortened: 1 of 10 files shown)


alxiong:
Thanks for the clarification. I understand the goal around aux data separation for future flexibility. Maybe I'm yet to realize the inefficiency of my refactor/simplification. 

Re: goal (3)
```rust
// previously
Step::witness(.., witness: DriverValue<Self::Witness<'source>>, ..) 
  -> DriverValue<D, Self::Aux<'source>>

// becomes
Step::witness(.., witness: DriverValue<Self::Witness<'source>>, ..) 
  -> DriverValue<D, Self::Output::Data<'source>>
```

It does seem that we can define `Self::Witness` as some borrowed data whose lifetime got inherited by `<Output as Header>::Data`, so that deep-cloning is avoidable. 
did I miss/misunderstood anything? 

> It would have been far more helpful for me to have written examples of how I expected the API to be used so that these requirements could emerge out of this refactoring an inform a better API

Yes, would be great if we have some intended usage that demonstrates the flexibility required to justify the `Step::Aux`

Regardless, I think we do agree that in the motivating example, `let leaf1 = leaf1.0.carry(leaf1.1);` where `leaf1.1: Aux` instead of `Header::Data` is an unintended mistake, right? 

----

This PR is not urgent, but would be super helpful if you can provide some examples when you have time. @ebfull 
(currently, i still incline towards my simplification, but definitely open to better API choice)

ebfull:
In the case of that particular example, I think it's up to the application to use distinct Header::Data types if they don't want them confused. That example / test code just exercises the API and I didn't care to make a newtype.

PR #405, opened 2026-01-30
pcd: mark claim builder structs as pub(crate)

alxiong:
A very nit suggestion. 
thought these could be internal details rather than public APIs. 
Feel free to simply close it if disagree.


TalDerei (approved):
this seems like a good step towards a broader API hygiene pass.

Issue #61, opened 2025-10-22
API change for rx-mesh interaction


TalDerei:
[rx](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_circuits/src/rx.rs) produces raw witness assignments, while the mesh _translates_ these into a structured polynomial r(X) based on compiler-pass optimizations. Concretely, floor planning changes the order of witness wires in the circuit, so witness generation must reorder wire assignments. We therefore need to transform the raw witness `r(x)` to the floor-planned witness `r'(x)` based on the permutation map computed during mesh finalization. There are two possible approaches here:

**1. Floor planner produces reordering map**

`rx()` generates raw witness (unstructured polynomial), mesh applies permutation after `rx()` completes, and the result is a structured, reordered polynomial.

```ts
let raw_witness = rx::eval(circuit, witness);  
let transformed = mesh.apply_permutation(raw_witness);
```

**2. Mesh owns witness generation**

Alternatively, we migrate the `rx()` method into the mesh, and mesh internally handles any floor planning transformations. This encapsulation means the mesh internally handles the transformation. Then, we can query the mesh for the permuted witness. 

```ts
let witness = mesh.rx(circuit_id, witness, circuit)
```

In this context, floor planning is a mesh-level optimization, and the mesh decides where to places the routines. This means the mesh bares the consequences for suboptimal placements.

Issue #60, opened 2025-10-22
Add compilation passes to mesh finalization


TalDerei:
Mesh finalization should include a "preprocessing" phase of compilation passes that simulate all the circuits in a mesh, perform a kind of holistic structural analysis on routine usage patterns, and applies an optimization step. The mesh orchestrates this compilation procedure and keeps an in-memory list of all routines that take place across the circuits. 

This "structural analysis" constitutes:
- **routine discovery:** emulator (or simulator?) driver that tracks routine invocation, and builds a routine registry to track which circuits are structurally equivalent (`RoutineRegistry` struct containing a `BTreeMap<TypeId, Vec<RoutineInfo>>`),
- **routine structural equivalence analysis:** somehow identify routines that are the same modulo the inputs,
- **floor planning and permutation map:** apply a floor planning step to optimize routine placement across circuits (references https://github.com/tachyon-zcash/ragu/issues/59), and based on floor planning, create a wire reordering map based on a _permutation_ that will applied during witness generation to account for layout changes (references https://github.com/tachyon-zcash/ragu/issues/61),
- **store:** persist this compilation pass information in the mesh.

Issue #59, opened 2025-10-22
Implement floor planning algorithm for optimal routine placements


TalDerei:
https://github.com/tachyon-zcash/ragu/issues/58 describes an optimization to cache routines across circuits, but this requires the routines to be in the _same_ position (same powers of X and Y). When they perfectly align, we can multiply by the sum of the lagrange coefficients instead of evaluating each routine independently. 

We need to implement a **floor planner** algorithm that determines the optimal placement of routines in the X and Y dimension to maximize memoization opportunities. The mesh should force the routines to optimally align during [finalization](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_circuits/src/mesh.rs#L61). This is similar to Halo's concept of _regions_. 

This implies a mechanism for tracking a routines shape and creating a floor plan heuristic based on that. The heuristic should align structurally equivalent routines across circuits and minimize gaps / fragmentation in X and Y dimension. The output should be some kind of placement map of X and Y offsets. Maybe we should have a dedicated `floorplan.rs` that implements a basic greedy floor planning algorithm?

Issue #58, opened 2025-10-22
Implement routine memoization in circuit polynomials


TalDerei:
[Routine](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_core/src/routines.rs#L30) calls occupy space in the X and Y dimensions by allocating multiplication gates and linear constraints respectively. Visually, this is a _rectangle_ in two-dimensional space. There are two optimization dimensions here _after_ audits:

**1. Optimizing multiple invocations of the the same routine (within) a single circuit**

```ts
  impl Circuit for Circuit_i {
      fn witness<D: Driver>(&self, dr: &mut D) {
          let x = dr.alloc(input);
          let h1 = dr.routine(Poseidon, x);      
          let h2 = dr.routine(Poseidon, h1);     
          let h3 = dr.routine(Poseidon, h2);   
      }
  }
```

A circuit polynomial `S(X, Y)` may have an internal structure:

```ts
S(X, Y) = [routine at position (x_{1}, y_{1})] + [routine at position (x_{2}, y_{2})] + ...
```

We can exploit the identical structure here across these routines by memoizing them – rather than recomputing the polynomial arithmetic for subsequent routine calls, we can scale and adjust the cached results based on the different inputs. 

In terms of how the driver builds these polynomials, it's representative of a _state accumulation scheme_: the driver maintains state `S` (current polynomial evaluation), performs a collection of `dr.mul()` and `dr.enforce_zero(lc)` operations, and advances the state to `S_new = S * Y^n + (accumulated LCs)`.

We can exploit this relationship by observing that, for subsequent invocations of the same routine, it's the same sequence of operations, same relative relationships between coefficients (same multiplication gates and linear constraints just different powers of X and Y), and the only differences are the starting X and Y, and the different input wire values. Therefore, we can compute the next routine as: `S'_new = S_new * X^N * Y^N + (adjustment factor for input wires)`, where `X^N and Y^N`  are relative offsets. 

In this context, `Sxy` can perform _full_ memoization, where `Sx` (indeterminant in Y) and `Sy` (indeterminant in X) can perform _partial_ memoization. 

**2. Optimizing multiple invocations of the the same routine (across) multiple circuits**

```ts
  impl Circuit for Circuit_1 {
      fn witness<D: Driver>(&self, dr: &mut D) {
          let hash = dr.routine(Poseidon, ...);
      }
  }

  impl Circuit for Circuit_2 {
[...]

PR #410, opened 2026-01-30
feat(routines): floor planner implementation

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/59 and builds on https://github.com/tachyon-zcash/ragu/pull/398 (I can't seem to change the base branch for some reason to collapse the diff, and rebasing doesn't display in the github UI). This implements the red section in the diagram. This PR starts at diff https://github.com/tachyon-zcash/ragu/pull/410/commits/206a1d946da721e03421e5449d89756e74bc79cb.

The primary function of the floor planner is to align the same routine types across circuits (inter-circuit) and minimize fragmentation in the mesh. The implementation is a simple greedy placement algorithm, `from_registries` (as apposed to more complex bin-packing) which collects routines from all circuits, sorts them by `circuit_count * max_invocations * area` (terming as "benefit"), and assigns the same routine types to canonical positions across all circuits based on that. It effectively produces a placement map with X, Y offsets.

  - `Placement`: One entry mapping a routine type to its location                                                                                                                                                                                      
  - `MeshPosition`: The (x, y) coordinates in the mesh  

https://github.com/tachyon-zcash/ragu/issues/58, https://github.com/tachyon-zcash/ragu/issues/60, and https://github.com/tachyon-zcash/ragu/issues/61 will subsequently use this information for the mesh. The mesh will use these positions when placing routines. That'll enable memoization both inter-circuit (same routines will have the same canonical positions and we can combine them via Lagrange coefficients) and intra-circuit (scaling by X and Y offsets).

### DIFF
--- a/crates/ragu_core/src/floor_plan.rs
+++ b/crates/ragu_core/src/floor_plan.rs
@@ -1,0 +1,206 @@
+//! Floor planning for optimal routine placement in the circuit mesh.
+//!
+//! The floor planner assigns canonical positions to routine types so that:
+//! - **Inter-circuit memoization**: Same routine type at same position across circuits
+//!   allows combining Lagrange coefficients instead of separate evaluation.
+//! - **Intra-circuit memoization**: Subsequent calls scale by `X^N * Y^M` offsets.
+//!
+//! See issues #58 and #59 for the memoization design.
+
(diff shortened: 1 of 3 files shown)


TalDerei:
closing in favor of consolidation in https://github.com/tachyon-zcash/ragu/pull/413

Issue #389, opened 2026-01-26
Testing routine memoization


TalDerei:
Meaningfully testing routine memoization like caching routine synthesis (eg. in sxy driver to reuse cached constraints) or performing parallel witness generation (eg. in rx driver) depends on https://github.com/tachyon-zcash/ragu/issues/58 and https://github.com/tachyon-zcash/ragu/issues/59.

Issue #67, opened 2025-10-24
Circuit breaking tags


TalDerei:
We should have a "circuit breaking" tag to identify a class of optimizations that aren't purely additive and can't be added later without changing the circuit polynomials. We should tag all _existing_ optimization dimensions along those lines, and this may inform whether we want to pursue certain kinds of optimizations in the future.

Issue #62, opened 2025-10-22
Identify routine equivalency


TalDerei:
https://github.com/tachyon-zcash/ragu/issues/60 mentions the problem of "routine structural equivalence analysis": how does the driver determine whether two routines are structurally identical? 

One approach is to mirror the gadget pattern and implement a `Routinekind` with `Any` supertrait:

```ts
pub trait RoutineKind: Any + Clone + Send + 'static { ... }

pub trait Routine<F>: RoutineKind { ... }
```

Alternatively, we can have a custom identification trait that takes in a unique `TypeID`:

```ts
  pub trait Routine<F>: Clone + Send {
      fn routine_id(&self) -> u64; 
  }
```

Maybe there's another approach for _runtime_ equivalence testing?

ebfull:
Routines are implementations of `Routine`:
* They cannot be compared by type, because two `Routine` instances can do two different things even if they are the same type; they can be configured.
* They cannot be compared by shape, because two routines of the same shape can do different things.
* They cannot be compared purely by the types of their input/output gadgets, because two `Routine` instances can have the same input/output types but do different things.

**It is only meaningful to compare _routine invocations._** Invocations can be _identified_ by their DFS synthesis order in a particular circuit, which is what the code currently does. Each invocation maps to a distinct segment of the trace, which corresponds with a contiguous portion of the multiplication and linear constraints of the wiring polynomial that they *directly* contribute to.

Routine invocations are _comparable_ when the *internal* portion of their wiring polynomial contribution (the portion that pertains exclusively to wires and constraints produced directly within that routine) are linearly related -- i.e., they are just shifted by $X^i Y^j$ with respect to each other. This comparison is useful partly because two routine invocations could produce the same internal wiring polynomial contribution (up to shifts) wherever they are repeated, which can be exploited for memoization. But just as importantly, if we can align the segments that comparable routine invocations map to in different circuits then the shifts themselves are unnecessary, and we can use this to further improve polynomial reduction performance.

> There is a separate approach I'd like to explore where nested routines' output wires are implicit inputs of their caller routines. I think this creates more opportunities for optimizations, allows segments to float independently of each other, and simplifies invocation comparison (fingerprinting). But it's also more complicated.

[...]

PR #413, opened 2026-01-31
feat(routines): floor planning infrastructure for routine memoization

TalDerei:
I wasn't able to change the base branch across PRs to collapse the diff for some reason, and rebasing doesn't reflect cleanly in the github UI, so I consolidated work streams into a single draft PR. I’m compositing these changes as a way to test feasibility.

Importantly, memoization is treated purely as an optimization and [shouldn't change](https://github.com/tachyon-zcash/ragu/issues/67) the underlying circuit polynomials.  

the thing we're tackling here is when evaluating `s(x,y)` for multiple circuits in the registry, identical routines in the same positions (think of a region that occupies space in the X and Y dimensions), and each circuit recomputes the same polynomial contribution. this work caches routine contributions by `(RoutineId, canonical_position (in registry/mesh))`, and when detecting cache hits, it skips routine execution entirely and reconstructs the output from cached wires that were saved. The different phases of the implementation are documented below.

-----------------

**1. Routine Identity and Shape Tracking** 

References https://github.com/tachyon-zcash/ragu/issues/62. Implements the blue section in the diagram. This PR starts at diff https://github.com/tachyon-zcash/ragu/pull/413/commits/6d6496e8cd8f19f65fc5e589a1dd779c090dd3ba. 

Introduces structs (`RoutineId`, `RoutineShape`, `RoutineInfo`, `RoutineRegistry`) _purely_ for routine discovery as a precursor to floor planning. This only implements the green sections in the diagram, and follow-up work will be handled separately. 
- `RoutineID`: type ID for determining what the routine is, using `RoutineId::of::<R>()` (identity),
- `RoutineShape`: shape metadata (multiplication / linear constraints) stored in `RoutineInfo` that will be used for floor planning placement (size),
- `RoutineInfo`: information about a single routine invocation,
- `RoutineRegistry`: tracks routine invocation, and builds a routine registry to track which circuits are structurally equivalent based on `RoutineID`; `RoutineRegistry` struct contains a `BTreeMap<TypeId, Vec<RoutineInfo>>` and uses `Emulator::counter()` (discovery).  

We want to maintain type-level fungibility property, so `RoutineId` is type-based. Actual memoization _eligibility_, after we've identified inter/intra-circuit routine matches, should depend on: the input pattern (static versus dynamic inputs, where a routine has cacheable _internal_ G and non-cacheable _external_ H polynomial parts), registry position, nesting level (routine composition), etc. 

In [Halo2](https://github.com/zcash/halo2), circuit identity is determined via cryptographic hash of the constraint system (derived at synthesis), floor planning models a two-pass system that measures shape and then applies a greedy placement algorithm, and memoization doesn't exist (naively recomputes synthesis each time). 

-------------------------

**2. Floor Planner Implementation**

References https://github.com/tachyon-zcash/ragu/issues/59. This implements the red section in the diagram. This PR starts at diff https://github.com/tachyon-zcash/ragu/pull/413/commits/748138da5eeb77f718c0ea63e40ce0a883152e8a.

The primary function of the floor planner is to align the same routine types across circuits (**inter-circuit**) and minimize fragmentation in the registry. The implementation is a simple greedy placement algorithm, `from_registries` (as apposed to more complex bin-packing) which collects routines from all circuits, sorts them by `circuit_count * max_invocations * area` (terming as "benefit"), and assigns the same routine types to canonical positions across all circuits based on that. It effectively produces a placement map with X, Y offsets.

  - `Placement`: One entry mapping a routine type to its location                                                                                                                                                                                      
  - `RegistryPosition`: The (x, y) coordinates in the registry  

------------------------

**3. Integrate Floor Plan with Registry**

References https://github.com/tachyon-zcash/ragu/issues/60. Implements the orange section in the diagram. This PR starts at diff https://github.com/tachyon-zcash/ragu/pull/413/commits/7ac6f7e826f4295260746eed4fcc89fcbf0913b4. 

The registry/registry subsequently uses these positions when placing routines. That'll enable memoization both inter-circuit (same routines will have the same canonical positions and we can combine them via Lagrange coefficients) and intra-circuit (scaling by X and Y offsets). 

------------------------

**4. Driver Memoization Scaffolding**

Provides scaffolding for green section in the diagram. The PR diff starts at https://github.com/tachyon-zcash/ragu/pull/413/commits/8de1e6968fb1beca0f28cdfeac423b1d4d27a22c. 

It adds `FloorPlan` parameter to `CircuitObject` trait methods (`sxy`, `sx`, `sy`) and synthesis driver methods, which is scaffolding that will enable routine memoization (by tracking invocation counts and looking up canonical positions). 

----------------------

**5. Exercise Memoization and Benchmarks/Tests**

(**Partially**) References https://github.com/tachyon-zcash/ragu/issues/58, https://github.com/tachyon-zcash/ragu/issues/389 and implements the green section in the diagram. This PR starts at diff https://github.com/tachyon-zcash/ragu/pull/413/commits/bf6afae9ad7af615d4d35be2e27e40c2b82a638f.

It implements `wxy_combined`, an optimized registry evaluation that shares memoization cache across circuits during lagrange interpolation. This is **inter-circuit** routine memoization (cache polynomial evaluations shared across circuits), not intra-circuit (multiple invocations within one circuit that can be cached by scaling, which isn't implemented _yet_ and will be considered follow-up work). 

**Importantly**, this is a very simple memoization impl that caches each routines contribution to the polynomial evaluation and reuses it when the same routine type appears at the same floor-plan position.

Basically on cache hits, it still _executes_ (skipping subsequent executions entirely would require caching output wire mappings in some structure) the routine and produces output wires, but it _skips_ adding each constraint to accumulator (saves about N multiply-adds in the Horner loop). More aggressive memorization optimizations can be later built on this like skipping routine execution entirely (which is the whole point, ), considering static / dynamic inputs which determine which parts of the polynomial can be cached, etc. 
- _update:_ we now more aggressively skip routine execution entirely on cache hits. It still doesn't the latter static / dynamic inputs.

----------------

Screenshot 2026-01-30 at 2 48 32 PM

### DIFF
--- a/crates/ragu_circuits/src/s/sx.rs
+++ b/crates/ragu_circuits/src/s/sx.rs
@@ -280,10 +280,9 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
             .and_modify(|c| *c += 1)
             .or_insert(1);
 
+        // TODO: Use these positions for memoization (currently only sxy memoizes)
         let _current_position =
             RegistryPosition::new(self.multiplication_constraints, self.linear_constraints);
-
-        // Look up canonical position from floor plan (for future memoization)
         let _canonical_position = self
             .floor_plan
             .get_invocation(&routine_id, invocation_index);
--- a/crates/ragu_circuits/src/s/sxy.rs
+++ b/crates/ragu_circuits/src/s/sxy.rs
@@ -289,12 +289,13 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
             .and_modify(|c| *c += 1)
             .or_insert(1);
 
-        // Canonical position is the cache key
+        // Canonical position is the cache key. Fallback to current position if
+        // not in floor plan (correct for intra-circuit, but inter-circuit
+        // memoization requires proper floor plan setup).
         let canonical_position = self
             .floor_plan
             .get_invocation(&routine_id, invocation_index)
             .unwrap_or_else(|| {
-                // Fallback if not in floor plan
                 RegistryPosition::new(self.multiplication_constraints, self.linear_constraints)
             });
 
@@ -304,6 +305,18 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
                 .get(&routine_id, canonical_position)
                 .cloned()
             {
+                // Validate cached shape matches floor plan expectation
+                if let Some(expected) = this.floor_plan.get_shape(&routine_id) {
+                    debug_assert_eq!(
+                        cached.num_multiplications, expected.num_multiplications,
+                        "cached multiplication count mismatch"
+                    );
+                    debug_assert_eq!(
+                        cached.num_constraints, expected.num_constraints,
+                        "cached constraint count mismatch"
+                    );
+                }
+
                 // Cache hit: reconstruct output via template + wire injection (no routine execution)
                 let result_before = this.result;
 
@@ -317,6 +330,10 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
                 let mut injector: WireInjector<'_, F, Self> =
                     WireInjector::new(&cached.output_wires);
                 let output = Ro::Output::map_gadget(&template, &mut injector)?;
+                debug_assert!(
+                    injector.is_exhausted(),
+                    "wire cache overflow: unconsumed wires"
+                );
 
                 // Manually advance constraint counters (we didn't call mul/enforce_zero)
                 this.multiplication_constraints += cached.num_multiplications;
@@ -499,7 +516,7 @@ mod tests {
         let key = registry::Key::new(Fp::random(&mut rand::rng()));
 
         let mut registry = RoutineRegistry::new();
-        let shape = SquareRoutine::shape();
+        let shape = SquareRoutine::shape::<Fp>();
         registry.register::<SquareRoutine>(shape);
         registry.register::<SquareRoutine>(shape);
         registry.register::<SquareRoutine>(shape);
@@ -542,7 +559,7 @@ mod tests {
         let key = registry::Key::new(Fp::random(&mut rand::rng()));
 
         let mut registry = RoutineRegistry::new();
-        let shape = SquareRoutine::shape();
+        let shape = SquareRoutine::shape::<Fp>();
         registry.register::<SquareRoutine>(shape);
         registry.register::<SquareRoutine>(shape);
         registry.register::<SquareRoutine>(shape);
--- a/crates/ragu_circuits/src/floor_plan.rs
+++ b/crates/ragu_circuits/src/floor_plan.rs
@@ -136,6 +136,14 @@ impl FloorPlan {
             )
         })
     }
+
+    /// Gets the shape for a routine type, if it exists in the floor plan.
+    pub fn get_shape(&self, id: &RoutineId) -> Option<RoutineShape> {
+        self.placements
+            .iter()
+            .find(|p| p.id == *id)
+            .map(|p| p.shape)
+    }
 }
 
 #[cfg(test)]
--- a/crates/ragu_circuits/src/s/common.rs
+++ b/crates/ragu_circuits/src/s/common.rs
@@ -64,6 +64,8 @@ pub struct CachedRoutine<F> {
 }
 
 /// Cache for routine contributions, keyed by `(RoutineId, canonical_position)`.
+///
+/// Currently used for inter-circuit memoization of polynomial structure.
 #[derive(Default, Clone)]
 pub struct MemoCache<F> {
     entries: BTreeMap<(RoutineId, RegistryPosition), CachedRoutine<F>>,
@@ -229,6 +231,11 @@ impl<'a, F, TD> WireInjector<'a, F, TD> {
             wires: wires.iter(),
             _marker: PhantomData,
         }
+    }
+
+    /// Returns true if all cached wires have been consumed.
+    pub fn is_exhausted(&self) -> bool {
+        self.wires.len() == 0
     }
 }
 
--- a/crates/ragu_circuits/src/s/sy.rs
+++ b/crates/ragu_circuits/src/s/sy.rs
@@ -624,10 +624,9 @@ impl<'table, 'sy, F: Field, R: Rank> Driver<'table> for Evaluator<'_, 'table, 's
             .and_modify(|c| *c += 1)
             .or_insert(1);
 
+        // TODO: Use these positions for memoization (currently only sxy memoizes)
         let _current_position =
             RegistryPosition::new(self.multiplication_constraints, self.linear_constraints);
-
-        // Look up canonical position from floor plan (for future memoization)
         let _canonical_position = self
             .floor_plan
             .get_invocation(&routine_id, invocation_index);
--- a/crates/ragu_circuits/src/test_fixtures/mod.rs
+++ b/crates/ragu_circuits/src/test_fixtures/mod.rs
@@ -16,7 +16,7 @@ use ragu_core::{
 };
 use ragu_primitives::Element;
 
-use crate::{Circuit, routines::RoutineShape};
+use crate::{Circuit, metrics::derive_shape, routines::RoutineShape};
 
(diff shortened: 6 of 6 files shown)


crates/ragu_circuits/benches/circuits.rs
```diff
@@ -155,9 +156,21 @@ fn wxy((registry, (w, x, y)): (Registry<'_, Fp, R<5>>, (Fp, Fp, Fp))) {
...
+#[bench::wxy_routines(registry_routines(), (f, f, f))]
+fn wxy_routines((registry, (w, x, y)): (Registry<'_, Fp, R<25>>, (Fp, Fp, Fp))) {
+    black_box(registry.wxy(w, x, y));
+}
+
+#[library_benchmark(setup = setup_with_rng)]
+#[bench::wxy_combined_routines(registry_routines(), (f, f, f))]
+fn wxy_combined_routines((registry, (w, x, y)): (Registry<'_, Fp, R<25>>, (Fp, Fp, Fp))) {
+    black_box(registry.wxy_combined(w, x, y));
+}
```
TalDerei: first real savings when analyzing the bench results from a simple memoization pass (see `"Exercise Memoization"` in PR description).

------------

Memoization comparison (wxy_routines vs wxy_combined_routines)

crates/ragu_circuits/src/registry.rs
```diff
@@ -347,11 +389,42 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+                *poly += circuit.sxy(x, y, &self.key, &self.floor_plan) * circuit_coeff;
             },
         )
     }
 
+    /// Evaluates the registry polynomial with inter-circuit memoization.
+    ///
+    /// Routines at the same floor plan position share cached contributions.
+    /// Result is identical to [`wxy`](Self::wxy).
+    pub fn wxy_combined(&self, w: F, x: F, y: F) -> F {
```
TalDerei: this is currently just used in tests / benches to demonstrate performance savings and equivalence against `wxy`. `wxy_combined` would be replaced with `wxy` to actually enable inter-circuit memoization.

crates/ragu_core/src/routines.rs
```diff
@@ -14,6 +40,101 @@ use crate::{
...
+/// those with lifetime parameters.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct RoutineId(&'static str);
+
+impl RoutineId {
+    /// Creates a `RoutineId` for the given routine type.
+    pub fn of<R>() -> Self {
+        Self(core::any::type_name::<R>())
+    }
+}
```
TalDerei: we need to ensure this is guaranteed to be unique.

crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -224,26 +240,99 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<F, R> {
...
+            .or_insert(1);
+
+        // Canonical position is the cache key
+        let canonical_position = self
+            .floor_plan
+            .get_invocation(&routine_id, invocation_index)
+            .unwrap_or_else(|| {
+                // Fallback if not in floor plan
+                MeshPosition::new(self.multiplication_constraints, self.linear_constraints)
+            });
```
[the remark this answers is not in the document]
TalDerei: it should be .expect("routine must be registered in floor plan"), but we'd also need to change existing tests to use the `register_circuit_with_registry` to be pass tests.

crates/ragu_core/src/routines.rs
```diff
@@ -5,7 +5,33 @@
...
 //! synthesis).
+//!
+//! Routines are reusable circuit components identified by [`RoutineId`], which wraps
+//! a type name string. Different type parameters yield different identities (eg.
+//! `Evaluate<R<10>>` and `Evaluate<R<13>>` are distinct routines).
+//!
+//! Routines must be **fungible by type**: all instances of a concrete routine type produce
+//! identical constraints. Encode constraint-affecting parameters as type parameters, not
+//! runtime fields.
+//!
```
ebfull: This isn't the way that I wished to do this.

Routines should not necessarily be functionally equivalent by type; they can differ in behavior based on their contents. This makes the task of identifying when two routines are the "same" more difficult because we must ascertain their "contribution" to $s(X, Y)$ modulo their input wires to establish equivalence. But we can at least depend on the routine's input / output gadgets being fungible (and also their concrete types!) and so a `(TypeId, TypeId)` pair can be used as a starting point for the comparison of two routine invocations. (The functional equivalence can then be determined by evaluating the routine operations at a random point, for instance.)

TalDerei: we don't currently distinguish between static versus dynamic inputs as noted above, but it's tracked in the associated issues.

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -249,11 +256,32 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<F, R> {
...
+            .and_modify(|c| *c += 1)
+            .or_insert(1);
+
+        let _current_position =
+            RegistryPosition::new(self.multiplication_constraints, self.linear_constraints);
+
+        // Look up canonical position from floor plan (for future memoization)
+        let _canonical_position = self
+            .floor_plan
+            .get_invocation(&routine_id, invocation_index);
```
TalDerei: doesn't do caching for other polynomial restrictions yet (only `sxy`).

ebfull (changes requested):
The common thread in my review comments are that pretty much nothing needs to change in `ragu_core` for this PR, so you should be suspicious of an architectural decision or API change that requires such a thing.

Needs rebasing to account for recent merges, but should work just fine. (did it locally with no problems...)

crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -230,20 +243,115 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<F, R> {
...
+
+            // Manually advance constraint counters (we didn't call mul/enforce_zero)
+            self.multiplication_constraints += cached.num_multiplications;
+            self.linear_constraints += cached.num_constraints;
+
+            // Add cached contribution: result = result_before * y^k + contribution
+            let y_power = self.y.pow_vartime([cached.num_constraints as u64]);
+            self.result = result_before * y_power + cached.contribution;
+
+            self.available_b = tmp;
```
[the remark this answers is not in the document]
TalDerei: https://github.com/tachyon-zcash/ragu/pull/465

crates/ragu_core/src/gadgets.rs
```diff
@@ -203,6 +203,16 @@ pub unsafe trait GadgetKind<F: Field>: core::any::Any {
...
+
+    /// Reconstructs a gadget from cached wire values.
+    ///
+    /// This method enables routine memoization by reconstructing output gadgets
+    /// directly from cached wires, bypassing routine execution entirely on cache
+    /// hits. Wires are consumed from the iterator in the same order that
+    /// [`map_gadget`](GadgetKind::map_gadget) visits them.
+    fn from_cached_wires<'dr, D: Driver<'dr, F = F>>(
+        wires: &mut impl Iterator<Item = D::Wire>,
+    ) -> Result<Self::Rebind<'dr, D>>;
```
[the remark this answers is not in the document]
TalDerei: https://github.com/tachyon-zcash/ragu/pull/413/commits/9ad51646338cee673f291b880cdd514349406825

crates/ragu_circuits/src/polynomials/txz.rs
```diff
@@ -35,6 +39,20 @@ impl<R: Rank> Evaluate<R> {
...
+    /// Returns the shape of this routine.
+    pub fn shape<F: Field>(&self) -> RoutineShape {
+        derive_shape(|dr: &mut ShapeCounter<F>| {
+            let x = Element::zero(dr);
+            let z = Element::zero(dr);
+            let aux = Empty;
+            let _ = self.execute(dr, (x, z), aux);
+        })
+    }
+}
```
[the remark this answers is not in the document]
TalDerei: should this be follow-up work, ie. actually integrating the routine memoization into the real code paths? It'd be a relatively light lift since we already do all the memoization bootstrapping in tests.

PR #415, opened 2026-01-31
Lazy registry builder with offset for masks and internal circuits

### LINKED ISSUE #414: Improve app builder and avoid interstitial circuit objects collection
As suggested in https://github.com/tachyon-zcash/ragu/pull/406#pullrequestreview-3729839599, a future work for #406 is to avoid circuit object collection in `ApplicationBuilder`. 

Originally (before 406), we were collecting application circuit object directly into a `native: RegistryBuilder`. We temporarily change it to a `Vec<Box<dyn CircuitObject>>` in 406 since we postpone their insertion.

alxiong:
Close #414. 

Changes include:

(apart from the `Deferred` lazy materialization of `CircuitObject` in https://github.com/alxiong/ragu/pull/1)

- `RegistryBuilder` now holds separate vector: `offset_masks`, `offset_circuits`, `circuits`
  - Separate `offset_masks` vector enables targeted evaluation optimizations for masking polynomials
  - No pre-allocation or dummy circuits
- Reordered registration: masks (0-7) → circuits (8-12) → steps (13-14) → application (15+)
  - Removed `OFFSET` const generic parameter
  - Updated index calculations: `InternalCircuitIndex` and `Step::circuit_index()` adjusted for new ordering

### DIFF
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -135,7 +135,7 @@ impl<'params, F: PrimeField, R: Rank> RegistryBuilder<'params, F, R> {
     /// Registers an internal stage mask.
     pub fn register_internal_mask<S>(mut self) -> Result<Self>
     where
-        S: Stage<F, R> + 'params,
+        S: Stage<F, R>,
     {
         self.internal_masks.push(S::mask()?);
         Ok(self)
@@ -144,7 +144,7 @@ impl<'params, F: PrimeField, R: Rank> RegistryBuilder<'params, F, R> {
     /// Registers an internal final stage mask.
     pub fn register_internal_final_mask<S>(mut self) -> Result<Self>
     where
-        S: Stage<F, R> + 'params,
+        S: Stage<F, R>,
     {
         self.internal_masks.push(S::final_mask()?);
         Ok(self)


crates/ragu_circuits/src/registry.rs
```diff
@@ -60,9 +118,23 @@ impl CircuitIndex {
...
-/// Builder for constructing a new [`Registry`].
+/// Builder for constructing a [`Registry`].
+///
+/// Circuits are stored in deferred form until [`finalize`](Self::finalize),
+/// avoiding synthesis overhead during registration.
+///
+/// Circuits are organized into three categories:
+/// - `offset_masks`: Stage masks and final masks registered via offset methods
+/// - `offset_circuits`: Internal circuits and steps registered via offset methods
+/// - `circuits`: Application circuits and masks
```
TalDerei:
```suggestion
/// - `circuits`: Application circuits registered via offset methods

File: crates/ragu_circuits/src/registry.rs [hunk 2]
```diff
@@ -72,58 +144,120 @@ impl<F: PrimeField, R: Rank> Default for RegistryBuilder<'_, F, R> {
...
+
+    /// Registers a final stage mask (mask creation deferred until finalization).
+    pub fn register_final_mask<S>(mut self) -> Result<Self>
+    where
+        S: Stage<F, R> + 'params,
+    {
+        self.circuits
+            .push(Box::new(DeferredFinalMask::<S, R>::default()));
+        Ok(self)
+    }
```
TalDerei: this is unnecessary; `nested/mod.rs` can call `register_offset_mask`, `register_offset_final_mask`, and `register_offset_circuit`

crates/ragu_circuits/src/registry.rs
```diff
@@ -60,9 +118,23 @@ impl CircuitIndex {
...
+/// - `circuits`: Application circuits and masks
+///
+/// During finalization, circuits are concatenated in the order:
+/// `offset_masks -> offset_circuits -> circuits`, ensuring internal masks can be
+/// optimized separately from circuits while maintaining proper PCD indexing.
 pub struct RegistryBuilder<'params, F: PrimeField, R: Rank> {
-    circuits: Vec<Box<dyn CircuitObject<F, R> + 'params>>,
+    offset_masks: Vec<Box<dyn Deferrable<'params, F, R> + 'params>>,
+    offset_circuits: Vec<Box<dyn Deferrable<'params, F, R> + 'params>>,
+    circuits: Vec<Box<dyn Deferrable<'params, F, R> + 'params>>,
```
TalDerei: maybe we can be explicit here with the naming; `offset_masks`, `offset_internal_circuits`, and `offset_application-circuits`?

crates/ragu_circuits/src/registry.rs
```diff
@@ -72,58 +144,120 @@ impl<F: PrimeField, R: Rank> Default for RegistryBuilder<'_, F, R> {
...
 
-    /// Registers a new circuit using a bare circuit object.
-    pub fn register_circuit_object(
-        mut self,
-        circuit: Box<dyn CircuitObject<F, R> + 'params>,
-    ) -> Result<Self> {
-        let id = self.circuits.len();
-        if id >= R::num_coeffs() {
-            return Err(Error::CircuitBoundExceeded(id));
-        }
```
TalDerei: we should probably add some kind of bound check like this in `finalize`

crates/ragu_pcd/src/step/mod.rs
```diff
@@ -114,17 +119,17 @@ fn test_index_map() -> Result<()> {
...
+    // Internal steps come after internal circuits (masks at 0-7, circuits at 8-12, steps at 13-14)
     assert_eq!(
         Index::internal(InternalStepIndex::Rerandomize).circuit_index(num_application_steps)?,
-        CircuitIndex::new(0)
+        CircuitIndex::new(NUM_INTERNAL_CIRCUITS)
     );
     assert_eq!(
         Index::internal(InternalStepIndex::Trivial).circuit_index(num_application_steps)?,
-        CircuitIndex::new(1)
+        CircuitIndex::new(NUM_INTERNAL_CIRCUITS + 1)
```
TalDerei: outside this PR, was inspecting the `NUM_INTERNAL_CIRCUITS` docs, which isn't actually currently accurate.

```
  /// The number of internal circuits registered by [`register_all`] and                                                                                                                                                                             
  /// [`super::nested::register_all`], and the number of variants in [`InternalCircuitIndex`].                                                                                                                                                       
  pub(crate) const NUM_INTERNAL_CIRCUITS: usize = 13;
```

crates/ragu_circuits/src/registry.rs
```diff
@@ -17,16 +17,69 @@
...
+struct DeferredCircuit<C>(C);
+
+impl<'a, F: Field, R: Rank, C: Circuit<F> + 'a> Deferrable<'a, F, R> for DeferredCircuit<C> {
+    fn materialize(self: Box<Self>) -> Result<Box<dyn CircuitObject<F, R> + 'a>> {
+        self.0.into_object()
+    }
+}
+
+/// Deferred mask wrapper; creates [`StageMask`] on materialize.
+struct DeferredMask<S, R>(PhantomData<fn() -> (S, R)>);
```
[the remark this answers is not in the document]
alxiong: Can't simplify it on a first try. Since we need Send/Sync bound on S and R, didn't think too hard on it.
Keeping as is for now.

crates/ragu_circuits/src/registry.rs
alxiong: good catch, removed!

crates/ragu_circuits/src/registry.rs
```diff
@@ -655,4 +789,85 @@ mod tests {
         use ff::Field;
         let _ = super::Key::new(<Fp as Field>::ZERO);
     }
+
+    #[test]
+    fn test_registry_with_offset() -> Result<()> {
```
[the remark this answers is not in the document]
alxiong: yes, removed

crates/ragu_pcd/src/circuits/nested/mod.rs
```diff
@@ -60,21 +59,18 @@ pub mod stages;
 pub(crate) fn register_all<'params, C: Cycle, R: Rank>(
```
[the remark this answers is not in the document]
alxiong: I updated the doc for it. is this sufficient, or do you prefer something else?

TalDerei (approved):
this looks good! will optimistically approve barring the field naming (discussed above) and doc changes.

TalDerei:
> This PR prematurely introduces a `Deferrable` trait, even though the user-facing API for the registry builder already forces a concrete delineation between the circuit objects. We can add this indirection later if it's useful, for now it makes this PR too noisy.

the motivation for that was described in this thread https://github.com/alxiong/ragu/pull/1#discussion_r2759713505. I think it does simplify and consolidate eventual optimization efforts. In that regard, it seems like a cleaner separation of concerns.

ebfull:
@TalDerei We could pull it back in when we need it. I'd like motivated examples of introducing that indirection, but even without them such a change should be in another PR.

PR #417, opened 2026-02-01
test(fuse): add tests for fuse pipeline stages

### LINKED ISSUE #348: Testing fuse pipeline
11 fusion stages in `ragu_pcd/src/fuse/` have no dedicated tests.

LucidSamuel:
Closes #348

### DIFF
--- a/crates/ragu_pcd/src/fuse/mod.rs
+++ b/crates/ragu_pcd/src/fuse/mod.rs
@@ -308,7 +308,11 @@ mod tests {
             let output_value = output.value().map(|v| *v);
             let output_enc = Encoded::from_gadget(output);
 
-            Ok(((left_enc, right_enc, output_enc), output_value.clone(), output_value))
+            Ok((
+                (left_enc, right_enc, output_enc),
+                output_value.clone(),
+                output_value,
+            ))
         }
     }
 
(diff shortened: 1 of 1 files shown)


TalDerei (changes requested):
the issue asked for dedicated tests for all 11 fuse stages, but this PR covers 6 of 11 with only narrow commitment-binding checks, and leaves 5 stages (`s_prime, error_n, f, eval, circuits / compute_internal_circuits`) exercised only by the smoke test. we should: 

- collapse the 6 commitment tests into one test that asserts all commitment bindings on a single fused proof                                                                                                                                      
- add commitment-binding assertions for the 5 missing stages (`s_prime, error_n, f, eval, circuits`)                   
- maybe also separately check other properties like p evaluation claim, FS transcript replay, nested commitment binding, etc. (maybe part of https://github.com/tachyon-zcash/ragu/pull/498 which can check "does the verifier catch corrupted P evaluations, corrupted transcript challenges, corrupted nested commitments."),                                                                                                           
- remove `fuse_completes_without_error` — already covered by `nontrivial::various_merging_operations`                                                                                                                                                 
- consider whether these tests belong in `fuse/mod.rs` at all, or should just extend `nontrivial.rs` using the existing fixtures   

noticed also that internal recursion circuits (`hashes_1, hashes_2, partial_collapse, full_collapse, compute_v`) have no dedicated tests and we don't have an issue tracking that. that should be tracked in a separate issue.

TalDerei:
thanks for working on this! closing for now in favor of revisiting later as a more comprehensive proptest.

Issue #51, opened 2025-10-16
Reduce coupling of Ragu with Pasta field behavior


ebfull:
Right now, Ragu is built assuming the Pasta curves are used in the following (known) ways:

* We assume certain ranks are available and that users will want or need to choose them, and those ranks work for the Pasta curves but may not work for other curve cycles.
* We assume $2$-adicity for the mesh polynomial domain.
* We assume the least significant 64 bits are distinct for all $2^k$ roots of unity in the field after multiplying them by $5$, which is very specific to the Pasta curves.

There is probably a general need for the `Cycle` implementation to determine the _concrete_ `Rank` choice that Ragu will use for all applications of that cycle, as well as any `Cycle`-specific behaviors like the hashing and the interpolation polynomial domain used in `Mesh`. This may require a change/refactor of the APIs.

See also #1.

Issue #1, opened 2025-06-30
pasta_curves dependency wishlist


ebfull:
`ragu_arithmetic` exposes traits that are tied to the `pasta_curves` crate, even if `pasta_curves` isn't actually used by a user of Ragu. (secp256k1/secq256k1 cycle for example.) And for whatever reason, these traits need the `alloc` feature to be enabled in `pasta_curves`, which brings in dependencies like `blake2b_simd` which users may also not want.

We also may want things from `pasta_curves` that it doesn't or may never generically provide. One example is a way to compute the discrete log of 2^k multiplicative subgroup elements. This could _maybe_ be exposed by `pasta_curves` since an efficient lookup implementation (`sqrt-tables`) exists for computing square roots.

I'd like to have generic interface for interacting with a field's ideal evaluation domain to simplify the API. The pasta curves would just use the 2^k roots of unity, and sec2 could use... maybe some mixed-radix FFT? maybe ECFFTs?

The only way to know the ideal solutions to these problems is to implement ragu, then make pasta_curves completely optional for using ragu and/or implement secp256k1 as a separate backend crate. In the process we'd create new traits in ragu_arithmetic or in other lower-level zkcrypto crates that capture a wide set of use cases.

PR #422, opened 2026-02-03
feat: introduce Transcript wrapper

### LINKED ISSUE #31: Add `PoseidonTranscript` wrapping `PoseidonPermutation` for Fiat-Shamir
Implement a transcript interface (used for injecting non-interactivity) that wraps the [PoseidonPermutation](https://github.com/ebfull/ragu/blob/main/crates/ragu_arithmetic/src/lib.rs#L171) primitive to implement a sponge abstraction that absorbs protocol messages (accumulator commitments, mesh polynomials, etc.) and squeezes deterministic challenges. It will be properly domain separated. We can also source inspiration from [merlin transcripts](https://github.com/dalek-cryptography/merlin). 

```ts
/// Transcript interface for Fiat-Shamir
trait Transcript<C: CurveAffine> {
      fn absorb_scalar(&mut self, scalar: C::Scalar);
      fn absorb_point(&mut self, point: &C);
      fn squeeze_challenge(&mut self) -> C::Scalar;
  }

struct PoseidonTranscript<C: CurveAffine> {
      permutation: PoseidonPermutation<C::Scalar>,
      state: [C::Scalar; RATE],
  }

impl<C: CurveAffine> PoseidonTranscript<C> {
      fn new(domain_separator: &'static [u8]) -> Self { ... }
}

impl<C: CurveAffine> Transcript<C> for PoseidonTranscript<C> { ... }
```

alxiong:
Closes #31 

Changes include:

- Introduce `trait TranscriptProtocol` (only protocol level domain-separator as discussed)
  - I decided to piggyback on `Buffer` for absorbing to transcript, gadgets can be `write(dr, &mut tr)` to transcript
  - add a `trait FromElements` as the opposite ops of `Write: GadgetKind`, implement it for `Element` and `FixedVec<Element>` 
  - `Transcript::challenge::<T: FromElements>()` makes a unified challenge squeezing interface with type-awareness (see the module doc for example usage)
- Add `struct Transcript` to wrap around `Sponge` and implements `TranscriptProtocol`
- update pcd caller (mostly in `hashes_1` and `hashes_2`)
  - removed `Sponge`'s `resume_and_squeeze()` instead just use `Transcript::resume_from().challenge()`
  - add a `b"ragu-pcd-v1"` protocol tag in `fuse()` for now.

### DIFF
--- a/crates/ragu_pcd/src/components/transcript.rs
+++ b/crates/ragu_pcd/src/components/transcript.rs
@@ -17,7 +17,7 @@
 //!
 //! // Save/resume for multi-circuit protocols
 //! let state = transcript.save_state(dr)?;
-//! let mut resumed = Transcript::resume_from_state(dr, state, params);
+//! let mut resumed = Transcript::resume_from_state(state, params);
 //! let challenge = resumed.challenge(dr)?; // must squeeze first
 //! let mut transcript = resumed.into_transcript(); // then can absorb again
 //! ```
@@ -58,6 +58,15 @@ impl<'dr, D: Driver<'dr>, P: PoseidonPermutation<D::F>> Transcript<'dr, D, P> {
     ///
     /// The `tag` is absorbed as field elements (length-prefixed, 16 bytes per
     /// element via u128 conversion) to bind the transcript to a protocol context.
+    ///
+    /// # Field size constraint
+    ///
+    /// Assumes the field modulus exceeds 128 bits so that each 16-byte chunk
+    /// maps to a unique field element. See [#51] and [#1] for the broader
+    /// effort to decouple Ragu from Pasta-specific field assumptions.
+    ///
+    /// [#51]: https://github.com/tachyon-zcash/ragu/issues/51
+    /// [#1]: https://github.com/tachyon-zcash/ragu/issues/1
     pub(crate) fn new(dr: &mut D, params: &'dr P, tag: &[u8]) -> Result<Self>
     where
         D::F: PrimeField,
@@ -101,11 +110,10 @@ impl<'dr, D: Driver<'dr>, P: PoseidonPermutation<D::F>> Transcript<'dr, D, P> {
     /// Call [`ResumedTranscript::into_transcript`] to transition back to a full
     /// transcript that supports absorbing.
     pub(crate) fn resume_from_state(
-        dr: &mut D,
         state: TranscriptState<'dr, D, P>,
         params: &'dr P,
     ) -> ResumedTranscript<'dr, D, P> {
-        let sponge = Sponge::resume(dr, state, params);
+        let sponge = Sponge::resume(state, params);
         ResumedTranscript {
             sponge,
             params,
@@ -216,8 +224,7 @@ mod tests {
             .save_state(&mut dr)
             .expect("save_state should succeed");
 
-        let mut resumed =
-            Transcript::resume_from_state(&mut dr, state, Pasta::circuit_poseidon(params));
+        let mut resumed = Transcript::resume_from_state(state, Pasta::circuit_poseidon(params));
         let challenge2 = resumed.challenge(&mut dr)?;
 
         // Both flows should produce the same challenge
@@ -302,7 +309,7 @@ mod tests {
         let mut t = Transcript::new(&mut dr, poseidon, b"resume-test")?;
         v1.write(&mut dr, &mut t)?;
         let state = t.save_state(&mut dr).expect("save_state should succeed");
-        let mut resumed = Transcript::resume_from_state(&mut dr, state, poseidon);
+        let mut resumed = Transcript::resume_from_state(state, poseidon);
         let c1 = *resumed.challenge(&mut dr)?.value().take();
         let mut t = resumed.into_transcript();
         v2.write(&mut dr, &mut t)?;
--- a/crates/ragu_primitives/src/poseidon.rs
+++ b/crates/ragu_primitives/src/poseidon.rs
@@ -230,7 +230,7 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
     ///
     /// This method allows resuming a sponge and then performing custom operations
     /// before squeezing. Used by the `Transcript` API.
-    pub fn resume(_dr: &mut D, state: SpongeState<'dr, D, P>, params: &'dr P) -> Self {
+    pub fn resume(state: SpongeState<'dr, D, P>, params: &'dr P) -> Self {
         Sponge {
             mode: Mode::Squeeze {
                 values: state.get_rate(),
@@ -518,7 +518,7 @@ mod tests {
             let value = Element::alloc(dr, value)?;
             sponge.absorb(dr, &value)?;
             let state = sponge.save_state(dr).expect("save_state should succeed");
-            let mut sponge = Sponge::resume(dr, state, Pasta::circuit_poseidon(params));
+            let mut sponge = Sponge::resume(state, Pasta::circuit_poseidon(params));
             let squeezed = sponge.squeeze(dr)?;
             save_resume_output.set(*squeezed.value().take());
             Ok(())
--- a/crates/ragu_pcd/src/circuits/native/hashes_2.rs
+++ b/crates/ragu_pcd/src/circuits/native/hashes_2.rs
@@ -168,11 +168,8 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
 
         // Resume transcript from saved state (error_m already absorbed in hashes_1)
         // and squeeze mu, nu (challenges from error_m absorption)
-        let mut resumed = Transcript::resume_from_state(
-            dr,
-            error_n.sponge_state,
-            C::circuit_poseidon(self.params),
-        );
+        let mut resumed =
+            Transcript::resume_from_state(error_n.sponge_state, C::circuit_poseidon(self.params));
         let mu = resumed.challenge(dr)?;
         unified_output.mu.set(mu);
 


crates/ragu_primitives/src/io.rs
```diff
@@ -50,6 +54,41 @@ pub trait Buffer<'dr, D: Driver<'dr>> {
...
+/// impl FromElements<'dr, D, 1> for Element<'dr, D> { ... }
+///
+/// // Point needs 2 field elements (x and y coordinates)
+/// impl FromElements<'dr, D, 2> for Point<'dr, D, C> { ... }
+///
+/// // Array of N elements
+/// impl FromElements<'dr, D, N> for [Element<'dr, D>; N] { ... }
+/// ```
+///
+/// [`challenge`]: (crate::transcript::TranscriptProtocol::challenge)
```
TalDerei: personal preference is using `squeeze_` nomenclature

crates/ragu_primitives/src/transcript.rs
```diff
@@ -0,0 +1,335 @@
...
+        // Absorb protocol tag as field elements
+        // Pack bytes into u64 chunks (8 bytes each)
+        for chunk in tag.chunks(8) {
+            let bytes: [u8; 8] = core::array::from_fn(|i| chunk.get(i).copied().unwrap_or(0));
+            let elem = Element::constant(dr, D::F::from(u64::from_le_bytes(bytes)));
+
+            self.sponge.absorb(dr, &elem)?;
+        }
+        Ok(())
+    }
```
TalDerei: this seems dubious to me. comparing against merlin, they support [length prefixing](https://github.com/dalek-cryptography/merlin/blob/master/src/transcript.rs#L90-L100), so we should absorb the length first

```ts
let len_elem = Element::constant(dr, D::F::from(tag.len() as u64));                                                                                                                                                                            
self.sponge.absorb(dr, &len_elem)?;

for chunk in tag.chunks(8) { ... }
```

crates/ragu_primitives/src/transcript.rs
```diff
@@ -0,0 +1,335 @@
...
+    ///
+    /// This should be called once at initialization to bind the transcript to a
+    /// protocol context. The tag is absorbed as field elements (8 bytes per
+    /// element via u64 conversion).
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// let mut transcript = Transcript::new(dr, params);
+    /// transcript.domain_sep(dr, b"ragu-pcd-v1")?;
```
TalDerei: maybe the domain tag `ragu-pcd-v1` should be a global constant?

crates/ragu_primitives/src/transcript.rs
alxiong: Good catch! if we encode the length, it will result in an extra permutation, thus exceeding `hashes_1` circuit size limit. 

here's what I decide to do:

`dom_sep = PrimeField::from_u128( [ tag as u64 | len as u64 ])`

so that we do have length-encoding while avoiding extra permutation.

crates/ragu_primitives/src/io.rs
```diff
@@ -50,6 +54,41 @@ pub trait Buffer<'dr, D: Driver<'dr>> {
...
+/// [`enforce_consistency`]: (ragu_core::gadgets::Consistent::enforce_consistency)
```
TalDerei:
```suggestion
/// [`challenge`]: crate::transcript::TranscriptProtocol::challenge                                                                                                                                                                                
/// [`enforce_consistency`]: ragu_core::gadgets::Consistent::enforce_consistency

File: crates/ragu_primitives/src/transcript.rs [hunk 5]
```diff
@@ -0,0 +1,338 @@
...
+    /// let mut transcript = Transcript::resume_from(dr, state, params)?;
+    /// let c: Element<'dr, D> = transcript.challenge(dr)?;
+    /// ```
+    fn save_state(self, dr: &mut D) -> core::result::Result<TranscriptState<'dr, D, P>, SaveError>;
+
+    /// Resume transcript from saved state.
+    ///
+    /// After resuming, call [`TranscriptProtocol::challenge`] to squeeze
+    /// challenges.
+    fn resume_from(dr: &mut D, state: TranscriptState<'dr, D, P>, params: &'dr P) -> Result<Self>;
```
TalDerei: `resume_from` should probably be made infallible (`resume` is already infallible)?

crates/ragu_primitives/src/vec.rs
```diff
@@ -181,6 +184,14 @@ impl<F: Field, G: Write<F>, L: Len> Write<F> for FixedVec<PhantomData<G>, L> {
     }
 }
 
+impl<'dr, D: Driver<'dr>, const N: usize> FromElements<'dr, D, N>
+    for FixedVec<Element<'dr, D>, ConstLen<N>>
+{
+    fn from_elements(_dr: &mut D, elements: [Element<'dr, D>; N]) -> Result<Self> {
```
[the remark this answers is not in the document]
alxiong: this deserves a discussion. I reserve driver as input because I'd expect some implementators (such as `Point::from_elements()` might invoke `enforce_consistent()` on it. 

think of it as deserializing bytes (here elements) into a Curve point, some deserializer choose to enforce validity (on_curve and not_small_subgroup check), but up to the implementator (as noted in the doc for this trait)

crates/ragu_primitives/src/io.rs
```diff
@@ -50,6 +54,41 @@ pub trait Buffer<'dr, D: Driver<'dr>> {
...
+/// // Array of N elements
+/// impl FromElements<'dr, D, N> for [Element<'dr, D>; N] { ... }
+/// ```
+///
+/// [`challenge`]: crate::transcript::TranscriptProtocol::challenge
+/// [`enforce_consistency`]: ragu_core::gadgets::Consistent::enforce_consistent
+pub trait FromElements<'dr, D: Driver<'dr>, const N: usize>: Sized {
+    /// Construct this type from exactly `N` field elements.
+    fn from_elements(dr: &mut D, elements: [Element<'dr, D>; N]) -> Result<Self>;
+}
```
ebfull: I'm not entirely sure about this trait yet... I think it only enables some patterns that we don't really exercise in this PR, and the diff could be simplified if we just removed it. It's at least worth considering later, but would be easier to justify if it were added alongside a compelling improvement to the code simultaneously.

ebfull (changes requested):
First pass for feedback.

crates/ragu_pcd/src/circuits/native/hashes_1.rs
```diff
@@ -226,16 +226,17 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
...
 
-        // Create a single long-lived sponge for all challenge derivations
-        let mut sponge = Sponge::new(dr, C::circuit_poseidon(self.params));
+        // Create a transcript for all challenge derivations
+        let mut transcript = Transcript::new(dr, C::circuit_poseidon(self.params));
+        transcript.domain_sep(dr, RAGU_TAG)?;
 
         // Derive w by absorbing nested_preamble_commitment and squeezing
-        let w = {
+        let w: Element<'_, _> = {
```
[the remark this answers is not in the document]
alxiong: the explicit type annotation is by design. I want to have unified `.challenge<T>()` where `T` could be `Element`, `FixedVec<Element, Len>`, `Point` etc. 

in some repo (halo2, merlin), they use specialize api for each challenge type `.challenge_scalar()`, `challenge_point()`,
in others (`spongefish`), they propose this unified api [`.verifier_message<T>()`](https://arkworks.rs/spongefish/spongefish/struct.VerifierState.html#method.verifier_message)

I took inspiration from `spongefish` and follow that, shown in the[ code example in doc](https://github.com/alxiong/ragu/blob/1af2ebb214d6acbf020bded908dd665d56cd0d44/crates/ragu_primitives/src/transcript.rs#L86C1-L89C83).
if you disagree, I can switch the `challenge_<type>()` api. @ebfull

crates/ragu_pcd/src/components/transcript.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub(crate) fn save_state(
+        self,
+        dr: &mut D,
+    ) -> core::result::Result<TranscriptState<'dr, D, P>, SaveError> {
+        self.sponge.save_state(dr)
+    }
+
+    /// Resume transcript from saved state.
+    ///
+    /// After resuming, call [`Self::challenge`] to squeeze challenges.
```
TalDerei:
```suggestion 
  /// Resume transcript from saved state.                                                                                                                                                                                                            
  ///                                                                                                                                                                                                                                                
  /// The resumed transcript is in **squeeze mode**, ready to output challenges                                                                                                                                                                      
  /// via [`Self::challenge`]. The typical pattern is to squeeze all challenges                                                                                                                                                                      
  /// derived from previously-absorbed data before absorbing new messages.                                                                                                                                                                           
  ///                                                                                                                                                                                                                                                
  /// Note: Calling `write()` before `challenge()` will transition to absorb mode,                                                                                                                                                                   
  /// discarding any unsqueezed challenge capacity from the saved state.

File: crates/ragu_pcd/src/components/transcript.rs [hunk 10]
```diff
@@ -0,0 +1,234 @@
...
+use ragu_core::{Result, drivers::Driver};
+
+use ragu_primitives::{
+    Element,
+    io::Buffer,
+    poseidon::{SaveError, Sponge, SpongeState},
+};
+
+/// Transcript wrapper around Poseidon [`Sponge`] for Fiat-Shamir transforms.
+pub(crate) struct Transcript<'dr, D: Driver<'dr>, P: PoseidonPermutation<D::F>> {
```
TalDerei: reconfirmed https://github.com/tachyon-zcash/ragu/pull/422#discussion_r2761029369 after the refactor, against registry digest in upstream main `0x000c5762bc28cd8fc9d33c2e131f2b3cc9d3d5a88b236fc4ae9dd7cda157ec09`.

TalDerei:
can you launch the [agent skills](https://github.com/tachyon-zcash/ragu/tree/main/.claude) to flag any documentation violations?

TalDerei:
@ebfull think this deserves another review pass

crates/ragu_pcd/src/components/transcript.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Note: Calling `write()` before `challenge()` transitions back to absorb mode.
+    /// Any remaining squeeze capacity from the saved state becomes inaccessible.
+    pub(crate) fn resume_from_state(
+        dr: &mut D,
+        state: TranscriptState<'dr, D, P>,
+        params: &'dr P,
+    ) -> Self {
+        let sponge = Sponge::resume(dr, state, params);
+        Transcript { sponge, params }
+    }
```
ebfull: The old `resume_and_squeeze` API was deliberately designed so that the save/resume boundary would sit in a narrow part of the `Sponge` pipeline to avoid edge cases. You had to save in absorb mode (performing the permutation, which makes it so that the rate is wiped clean) and then you can resume by squeezing immediately. The caller was basically prohibited from accidentally putting the `Sponge` into a bad state.

This change hands back a `Transcript` with buffered rate values that the caller is free to either squeeze or absorb from. If they absorb first, the values are silently discarded.

ebfull (changes requested):

crates/ragu_pcd/src/components/transcript.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        // Then absorb the tag content in 16-byte chunks as u128
+        for chunk in tag.chunks(16) {
+            let bytes: [u8; 16] = core::array::from_fn(|i| chunk.get(i).copied().unwrap_or(0));
+            let elem = Element::constant(dr, D::F::from_u128(u128::from_le_bytes(bytes)));
+            sponge.absorb(dr, &elem)?;
+        }
+
+        Ok(Transcript { sponge, params })
+    }
```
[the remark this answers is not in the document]
TalDerei: these are relevant https://github.com/tachyon-zcash/ragu/issues/1, https://github.com/tachyon-zcash/ragu/issues/51 (was looking through them this morning), will add a comment to that effect.

crates/ragu_primitives/src/poseidon.rs
```diff
@@ -226,31 +226,27 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
...
-        dr: &mut D,
-        state: SpongeState<'dr, D, P>,
-        params: &'dr P,
-    ) -> Result<(Element<'dr, D>, Self)> {
-        let mut sponge = Sponge {
+    /// Resumes a [`Sponge`] from a saved [`SpongeState`].
+    ///
+    /// This method allows resuming a sponge and then performing custom operations
+    /// before squeezing. Used by the `Transcript` API.
+    pub fn resume(_dr: &mut D, state: SpongeState<'dr, D, P>, params: &'dr P) -> Self {
```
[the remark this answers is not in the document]
TalDerei: previously flagged, but this could be removed afaict

PR #424, opened 2026-02-03
docs: add inline docs for revdot claims, proof layout, and polynomial structure

LucidSamuel:
Note: I made these intentionally concise inline docs for dev context and thinking of adding the full conceptual derivations (revdot, t(X,z), structured layout) in the book. confirming if this is useful so I can open a follow‑up issue for that.

### DIFF
--- a/crates/ragu_circuits/src/polynomials/mod.rs
+++ b/crates/ragu_circuits/src/polynomials/mod.rs
@@ -41,7 +41,15 @@ pub trait Rank:
         Self::RANK - 2
     }
 
-    /// Computes the coefficients of $$t(X, z) = -\sum_{i=0}^{n - 1} X^{4n - 1 - i} (z^{2n - 1 - i} + z^{2n + i})$$ for some $z \in \mathbb{F}$.
+    /// Computes the coefficients of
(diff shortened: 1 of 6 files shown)


crates/ragu_circuits/src/polynomials/mod.rs
```diff
@@ -57,7 +64,12 @@ pub trait Rank:
         tmp
     }
 
-    /// Computes the coefficients of $$t(x, Z) = -\sum_{i=0}^{n - 1} x^{4n - 1 - i} (Z^{2n - 1 - i} + Z^{2n + i})$$ for some $x \in \mathbb{F}$.
+    /// Computes the coefficients of
+    /// $$t(x, Z) = -\sum_{i=0}^{n - 1} x^{4n - 1 - i} (Z^{2n - 1 - i} + Z^{2n + i})$$
+    /// for some $x \in \mathbb{F}$.
+    ///
+    /// This is the symmetric counterpart to [`Rank::tz`] and is used to
+    /// evaluate the same polynomial when the roles of `x` and `z` are flipped.
```
ebfull: Hm, this isn't an entirely natural way for me to read a description of such a polynomial relationship. `x` and `z` don't really have roles; they vary here in another quality (their variance).

ebfull (changes requested):

PR #428, opened 2026-02-05
chore: add CLAUDE.md for sharing

alxiong:
As per [claude's doc ](https://code.claude.com/docs/en/best-practices?search=share#write-an-effective-claude-md)recommends: 

> You can place CLAUDE.md files in several locations:
> - Home folder (~/.claude/CLAUDE.md): Applies to all Claude sessions
> - Project root (./CLAUDE.md): Check into git to share with your team, or name it CLAUDE.local.md and .gitignore it

This is my current instructions, some `/init` generated, some manually summarized by my experience of getting PR review from Sean and Tal. 

Don't need to merge if we don't want to force upon everyone. But feel free to pick piecemeal section (such as "documentation style") from it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,6 +24,18 @@ cargo test --all-features <test_name>
 - `just bench` - Auto-detects platform (Linux native, macOS uses Docker)
 - macOS benchmarks run in Docker with valgrind for profiling
 
+### Worktree Setup
+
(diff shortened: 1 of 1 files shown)


ebfull:
I think the `CLAUDE.md` files are supposed to be limited in scope, since every token counts and it is injected into every context. My picky standards stuff seems like it should be some kind of agent or skill or something.

PR #434, opened 2026-02-05
chore(pcd): evaluate a at xz, remove redundant rx(x) queries

TalDerei:
Evaluates `a_poly` at `xz` instead of `x`, eliminating 22 redundant `rx(x)` queries.

### DIFF
--- a/crates/ragu_pcd/src/circuits/native/compute_v.rs
+++ b/crates/ragu_pcd/src/circuits/native/compute_v.rs
@@ -7,7 +7,7 @@
 //!
 //! ### Revdot folding
 //! - Retrieve layer 1 challenges [$\mu$], [$\nu$] and layer 2 challenges [$\mu'$], [$\nu'$]
-//! - Compute $a(x)$ and $b(x)$ via two-layer revdot folding of evaluation claims
+//! - Compute $a(xz)$ and $b(x)$ via two-layer revdot folding of evaluation claims
 //!
 //! ### $f(u)$ computation
 //! - Compute inverse denominators $(u - x_i)^{-1}$ for all evaluation points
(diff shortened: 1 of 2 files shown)


crates/ragu_pcd/src/circuits/native/compute_v.rs
```diff
@@ -403,10 +403,10 @@ impl<'a, 'dr, D: Driver<'dr>> Source for EvaluationSource<'a, 'dr, D> {
     fn rx(&self, component: RxComponent) -> impl Iterator<Item = Self::Rx> {
         use RxComponent::*;
         let (left, right) = match component {
-            // Raw claims: only x evaluation is available
+            // Raw claims: a uses xz evaluation, b uses x evaluation
             AbA => (
-                RxEval::X(&self.left.a_poly_at_x),
-                RxEval::X(&self.right.a_poly_at_x),
+                RxEval::OnlyXz(&self.left.a_poly_at_xz),
+                RxEval::OnlyXz(&self.right.a_poly_at_xz),
```
TalDerei: there's _probably_ a `RxEval` simplification here but that felt like more work

crates/ragu_pcd/src/circuits/native/stages/query.rs
```diff
@@ -66,10 +66,12 @@ impl<'dr, D: Driver<'dr>> XzQuery<'dr, D> {
...
-/// claims, only the x evaluation is available.
+/// claims, only one evaluation is available (xz for a, x for b).
 pub enum RxEval<'a, 'dr, D: Driver<'dr>> {
-    /// Only the x evaluation is available (used for raw a/b queries).
+    /// Only the x evaluation is available (used for raw b queries).
     X(&'a Element<'dr, D>),
+    /// Only the xz evaluation is available (used for raw a queries).
+    OnlyXz(&'a Element<'dr, D>),
     /// Both x and xz evaluations are available.
     Xz(&'a Element<'dr, D>, &'a Element<'dr, D>),
```
ebfull: What do we still use this for?

TalDerei: the raw a claims for child proofs. related to the comment above, we should be able to simplify to just two variants here like before, but haven't gotten around to it. there was a conflict with using the existing variants, so I added another one for now just to get it compiling.

PR #436, opened 2026-02-06
chore(pcd): remove unused `rx(x)` witness fields

TalDerei:
Suggestions for https://github.com/tachyon-zcash/ragu/pull/434#discussion_r2771885450 (handled by the last commit since I can't stack PRs). `rx` polynomials (circuit / stage claims) are now only evaluated at `xz`, and used by both a and b, so we can remove the 22 unused `rx(x)` witness fields in `XzQueryWitness`/`XzQuery` (the raw `a/b` claims aren't in these structs).


ebfull (changes requested):
Awesome!

ebfull (approved):

Issue #321, opened 2026-01-08
Derive the number of unified circuits from circuit registration rather than hardcoding


TalDerei:
The `NUM_UNIFIED_CIRCUITS` constant is currently hardcoded in [partial_collapse.rs](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_pcd/src/circuits/partial_collapse.rs) and used in fuse and verification. Rather than hardcoding, we should derive this (and other similar hardcoded constants) from the actual circuit registration.

PR #443, opened 2026-02-07
chore(pcd): clarify `NUM_UNIFIED_CIRCUITS` doc comments

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/321,  clarifies doc comment for `NUM_UNIFIED_CIRCUITS` constant, listing which circuits use unified k(y). It previously stated `hashes_1`, which was incorrect as discovered below.


ebfull:
I'm kind of scared to do it like this. It requires us to keep the calculation definition consistent with the layout, which kind of defeats the purpose since we already do that now manually anyway.

TalDerei:
> Commit / PR titles do not reflect the PR changes. ACK on them, though.

updated stale references accordingly

Issue #440, opened 2026-02-07
clarify Poseidon RO vs CRH statement


TalDerei:
In the protocol overview, the text states "Prolific use of Poseidon as a random oracle, though ideally not as a collision-resistant hash function." This is contradictory, since a random oracle is inherently collision-resistant.

PR #446, opened 2026-02-07
book: poseidon RO implies collision resistance 

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/440, based on auditor suggestions to document this.


TalDerei:
rebased diff

ebfull (approved):

Issue #241, opened 2025-12-14
Implement custom `enforce_equal_gadget` for gadgets


TalDerei:
The documentation states: 

```ts
Gadgets offer the [`GadgetKind::enforce_equal_gadget`](ragu_core::gadgets::GadgetKind::enforce_equal_gadget) 
method to specify how two instances can be enforced to be equivalent. In theory, a gadget can provide a more efficient 
implementation of this comparison than a generic method that simply creates linear constraints between two gadgets' wires.
```

This means the `GadgetKind` trait was designed to be overloadable, where individual gadget types can provide custom implementations of `enforce_equal_gadget` instead of relying on `#[derive(Gadget)]`. We haven't exploited this optimization yet where instead of enforcing equality between two gadgets by checking each wire _individually_, we can use a random linear combination to compare N wires with a single constraint. This would be good, for instance, for checking equality across dozens of boolean gadgets using just one constraint. This can extensibly be applied to other cases as well. 

We should implement a custom `enforce_equal_gadget` method on the `GadgetKind` trait that:

- [ ] Uses fixed/random linear combination for bulk equality checks
- [ ] Reduces N equality constraints to a single constraint

TalDerei:
initial impl was closed in https://github.com/tachyon-zcash/ragu/pull/449#issuecomment-4347126056. this issue stays open, but reference design notes per https://github.com/tachyon-zcash/ragu/pull/449#pullrequestreview-4198646806.

PR #449, opened 2026-02-07
feat(primitives): batch equality for boolean gadget

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/241. New `BooleanVec` gadget with the first custom implementation of `enforce_equal_gadget` instead of relying on `#[derive(Gadget)]` (specifically other gadgets have `enforce_equal_gadget` impls but they are structurally similar to what `#[derive(Gadget)]` generates, which is naively an element-wise equality). Recall from the issue, "the `GadgetKind` trait was designed to be overloadable, where individual gadget types can provide custom implementations of `enforce_equal_gadget` instead of relying on `#[derive(Gadget)]`"). 

Specifically  implements `BooleanVec` gadget (specialized `FixedVec<Boolean, L>` wrapper with an optimized `enforce_equal_gadget`) that reduces N constraints to 1 using a fixed linear combination technique (eg. `FixedVec<Element`, L> and `FixedVec<Point, L>` can use RLC). 

```ts
  let a: BooleanVec<_, ConstLen<8>> = BooleanVec::new(bools_a);                                                                                                                                                                                      
  let b: BooleanVec<_, ConstLen<8>> = BooleanVec::new(bools_b);                                                                                                                                                                                      
  a.enforce_equal(dr, &b)?;  
```

--------------

**Update**: recognized that `FixedVec` had a fixed `Kind`, which was a limitation for doing something like [this](https://github.com/tachyon-zcash/ragu/pull/449#issuecomment-3883166056) as @ebfull suggested, so added a default method to `GadgetKind` and had the boolean gadget override it with the packed _fixed_ LC optimization, rather than introducing a new gadget `FixedVec` impl.


crates/ragu_primitives/src/boolean.rs
```diff
@@ -251,6 +255,101 @@ impl<'dr, D: Driver<'dr>> Consistent<'dr, D> for Boolean<'dr, D> {
...
+
+/// The [`GadgetKind`] for [`BooleanVec`].
+pub struct BooleanVecKind<L: Len> {
+    _marker: PhantomData<L>,
+}
+
+// Safety: `BooleanVec` contains only `Boolean` gadgets which satisfy the `Send`
+// requirement when `D::Wire` is `Send`.
+unsafe impl<F: Field, L: Len> GadgetKind<F> for BooleanVecKind<L> {
+    type Rebind<'dr, D: Driver<'dr, F = F>> = BooleanVec<'dr, D, L>;
```
TalDerei: just realized this doesn't account for field overflows (should the approach we take here be similar to `MultiPack` which adds `PrimeField` trait bound and adds some runtime assertions?)

crates/ragu_primitives/src/boolean.rs
```diff
@@ -251,6 +255,101 @@ impl<'dr, D: Driver<'dr>> Consistent<'dr, D> for Boolean<'dr, D> {
...
+
+impl<'dr, D: Driver<'dr>, L: Len> core::ops::Deref for BooleanVec<'dr, D, L> {
+    type Target = [Boolean<'dr, D>];
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl<'dr, D: Driver<'dr>, L: Len> Clone for BooleanVec<'dr, D, L> {
```
alxiong: directly `#[derive(Clone)]` since it's equivalent?

alxiong (changes requested):
Looks good as a first step! 
only lacking logic for dealing with longer than `F::NUM_BITS` length vector.

crates/ragu_primitives/src/boolean.rs
TalDerei: that's certainly more graceful, matching [MultiPack's](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_primitives/src/boolean.rs#L221) approach. to avoid field overflows, we  could chunk by `F::CAPACITY` (and call `enforce_zero()` per chunk), which yields `ceil(N/F::CAPACITY)` constraints. the assertion approach though strictly imposed a single constraint, panicking on overflow.

crates/ragu_primitives/src/boolean.rs
TalDerei: originally tried that but got the error message, which prompted that manual clone on the inner field. they didn't appear to be equivalent.

```ts
     294 | impl<'dr, D: Driver<'dr>, L: Len> Gadget<'dr, D> for BooleanVec<'dr, D, L> {                                                                                                                                                              
         |                                                      ^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `D` 
```

alxiong (approved):

ebfull:
I wonder if we could generalize this. Ideally, we'd have a trait that Boolean could implement so that a single Vec-like structure could exist for every type that has its own multi-object equality optimization. Then we wouldn't need BooleanVec. Thoughts on this?

alxiong:
> so that a single Vec-like structure could exist for every type that has its own multi-object equality optimization.

imo, it's a premature generalization. 
we can abstract this when we observe similar opportunities in another contexts. (to justify the mental cost of an extra trait)

TalDerei:
> I wonder if we could generalize this. Ideally, we'd have a trait that Boolean could implement so that a single Vec-like structure could exist for every type that has its own multi-object equality optimization. Then we wouldn't need BooleanVec. Thoughts on this?

It'd be nicer to define just `FixedVec<Boolean>` and have it work the same way, but iirc there was a type system limitation. wasn't able to specialize `FixedVec<G>` which already has a `GadgetKind` impl (`FixedVecKind`).

> we can abstract this when we observe similar opportunities in another contexts.

this implements fixed LC for booleans, but you can extensively do random LCs for `Point` and `Element` gadgets as well.

crates/ragu_core/src/gadgets.rs
```diff
@@ -240,6 +240,30 @@ pub unsafe trait GadgetKind<F: Field>: core::any::Any {
...
+
+    /// Batch-enforces equality across slices of gadgets.
+    ///
+    /// The default implementation calls [`enforce_equal_gadget`](Self::enforce_equal_gadget)
+    /// per element. Gadget kinds whose values are bounded (e.g. booleans) can
+    /// override this to pack multiple equalities into fewer constraints.
+    ///
+    /// Container kinds delegate to this method so that the element kind's
+    /// optimization is applied automatically.
+    fn enforce_equal_gadget_slice<
```
TalDerei, line 261: `enforce_equal_gadget_slice` is basically an overridable hook of sorts on `GadgetKind` (eg. boolean override that packs N equalities into `ceil(N/capacity)` constraints). `FixedVec` delegates to `enforce_equal_gadget_slice`.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -572,3 +652,177 @@ fn test_multipack_vector() -> Result<()> {
 
     Ok(())
 }
+
+/// Derived gadgets with multiple Boolean fields get packed equality via the
+/// macro's batching through `enforce_equal_gadget_slice`.
+#[test]
+fn test_derived_gadget_boolean_batching() -> Result<()> {
```
TalDerei, line 659: we have tests for `FixedVec<Boolean>` and `#[derive(Gadget)]` that demonstrate constraint savings (the same logic in practice since both end up calling `Boolean::enforce_equal_gadget_slice` under the hood). we don't see any constraint savings in prod since endoscaling is currently **unenforced** per https://github.com/tachyon-zcash/ragu/pull/353#discussion_r2702571715, which will change once we wire up the routing circuits, and then we should see some kind of constraint reduction. these are `enforce_equal` savings, so it's a lower priority. maybe this should block on that? cc @ebfull

TalDerei:
> I wonder if we could generalize this. Ideally, we'd have a trait that Boolean could implement so that a single Vec-like structure could exist for every type that has its own multi-object equality optimization. Then we wouldn't need BooleanVec. Thoughts on this?

amended the construction, see readme for details.

alxiong (changes requested):
i agree with this directionally, it aligns with Sean's original intent to avoid specialized vec: `BooleanVec`, `ElementVec` etc.

but observing that we hand-written implement `Clone` and `Gadget` (around 150 lines) just to override one single function. 
imo, this signals an improvement to derive-macro to allow optionally override, everything else is still auto-derived and default; extra benefit is that user don't need to write `unsafe impl`

`#derive(Gadget(enforce_equal_slice = my_func)`

TalDerei:
> #derive(Gadget(enforce_equal_slice = my_func)

was abstracting the UX for now, but agreed. not sure we want to pepper attributes everywhere, instead it'd be nice if the user implemented some kind of like marker trait and derive picks it up automatically or something? added the attribute approach for now and we can explore other options as we extend these fixed / random LC optimizations to other gadgets. we should also stringently audit these macro changes.

alxiong:
instructed claude to do the following:
- Replaced the two ad-hoc extractors (`extract_ragu_path`, `extract_ragu_where_predicates`) with a single `ContainerAttrs` struct (serde-like), parsed by one from_attrs pass that rejects unknown keys instead of silently consuming them.
- Renamed the attribute key `impl_where → bound` (matches `#[serde(bound = "...")]`).
- Extracted `build_where_clause(base, extra, transform)` helper; the `Clone/Gadget` and `GadgetKind` where-clauses now flow through it instead of duplicating the merge-and-emit logic.
- Added `test_fail_unknown_ragu_attribute` and `test_fail_duplicate_bound`

alxiong (approved):
ACK from me. both the ergonomic and the `packed_boolean_equality()` impl looks good.

crates/ragu_macros/src/derive/gadget.rs
```diff
@@ -1,16 +1,94 @@
...
+    }
+    for predicate in predicates.iter_mut() {
+        transform(predicate);
+    }
+    if predicates.is_empty() {
+        quote! {}
+    } else {
+        quote! { where #predicates }
+    }
+}
```
ebfull, line 90: This and some of the other changes to the macros are independently useful I think and should be factored out and integrated separately.

crates/ragu_core/src/gadgets.rs
```diff
@@ -240,6 +249,30 @@ pub unsafe trait GadgetKind<F: Field>: core::any::Any {
...
+        'dr,
+        D1: Driver<'dr, F = F>,
+        D2: Driver<'dr, F = F, Wire = <D1 as Driver<'dr>>::Wire>,
+    >(
+        dr: &mut D1,
+        a: &[Bound<'dr, D2, Self>],
+        b: &[Bound<'dr, D2, Self>],
+    ) -> Result<()> {
+        assert_eq!(a.len(), b.len());
```
ebfull, line 270: There's something wrong with this. The caller (circuit code) can trip this assertion up, since it's not enforcing a protected invariant.

crates/ragu_macros/src/derive/gadget.rs
```diff
@@ -180,22 +265,64 @@ pub fn derive(input: DeriveInput, ragu_core_path: RaguCorePath) -> Result<TokenS
         }
     };
 
-    let equality_calls = fields.iter().filter_map(|(id, ty)| match ty {
-        FieldType::Wire => {
-            Some(quote! { #ragu_core_path::drivers::Driver::enforce_equal(dr, &a.#id, &b.#id)? })
+    // Group gadget fields by type so same-kind fields can be batched
+    // through `enforce_equal_gadget_slice` for reduced constraints.
```
ebfull, line 269: This is fragile, since it depends on the _name_ of the type determining if it's the same between two fields and thus qualifies. There are numerous ways this can break: type aliases, varying path qualifications, re-exports, etc.).

crates/ragu_primitives/src/boolean.rs
```diff
@@ -24,9 +24,41 @@ use crate::{
...
+        dr.enforce_zero(|mut lc| {
+            for (a_bit, b_bit) in a_chunk.iter().zip(b_chunk.iter()) {
+                lc = lc.add(&a_bit.wire).sub(&b_bit.wire);
+                lc = lc.gain(Coeff::Two);
+            }
+            lc
+        })?;
+    }
+    Ok(())
+}
```
ebfull, line 55: This is basically an inlined version of the existing `multipack` right? I wonder if there's a way to merge the two functions.

ebfull (changes requested):
My intuition is that this is the wrong approach, even though it's _roughly_ the minimal change necessary to accomplish the goal.

It's worth noting that the entire need here is artificial and an architectural limitation. Any two wires that are unconditionally equal should be unified. In frameworks like Noir circuits are optimized to do things like that. In our protocol, constraints are ideally assembled in this homogeneous way to provoke structure in the polynomials that never need to exist in the constructions Noir targets.

Fortunately, it's pretty rare to need to enforce unconditional equality of two gadgets in my experience. In those rare cases the algorithms that produce the two comparands can sometimes be refactored to unify them directly within synthesis.

The entire reason we're discussing this particular approach is because, theoretically, the wire-by-wire comparison of two gadgets is expensive. But observe that we're working within a unique model. Unconditional equality of two variables in R1CS always requires at least one *multiplication* constraint, but we get linearly independent constraints between our variables without multiplications. In fact, they cost nothing to enforce at proving time; they do not affect the trace, or synthesis, and the `packed_boolean_equality` function in this PR is a no-op at proving time. It may even be dead-code eliminated. There are also thousands of free linear constraints in practice, so given it's a rare occurance, there's plenty of headroom.

Let's suppose it did matter. Well, the approach in this PR reaches into the `ragu_core` internals to achieve it. There are several tensions here:

1. Equality optimizations can span nesting levels (booleans nested under a struct nested under a `Vec` could in principle pack into one batch); this PR only permits batching within a single nesting level.
2. Are slice-of-same-type the only class of relations a strategy could apply to?
3. The lack of impl specialization in Rust makes some of these designs impossible due to coherence violations. This means we have to work with macros and manual annotations, which are limited because they cannot reason about the types and impls.

I suspect that a different design could generalize to address these issues. In my head, it works like this:

1. There are different "equality strategy" traits. These live in `ragu_primitives`, not in `ragu_core`.
2. `ragu_core` removes the `enforce_equal(_gadget)` API, since it is a simple (dumb) strategy of just comparing wires. We can do this using `WireMap` in a blanket impl, for example, so we don't need this to be integrated into `Gadget`/`GadgetExt` anymore.
3. We can add a new `PackedComparator` strategy trait, which `Boolean` implements. All strategies have a supertrait `Strategy` that defines a `Strategy::Context` type that is threaded through; the context type is used to store state for the comparison, like for chunking the wires in a boolean linear constraint.
4. Gadgets that contain `Boolean`s can implement `PackedComparator` to expose the equality strategies they support, even if they themselves do not directly have any boolean fields. They forward this to all the gadgets below them, unless they have reason to create a parallel equality strategy for some of their fields.

This all could be automated by a macro which requires the implementor to provide:

1. annotations for all the strategies they want their derived Gadget to implement.
2. annotations for all the strategies that each of their fields do implement, probably ordered in terms of preference

and the macro could figure out for itself how to create each impl sharing or forwarding the most common strategies amongst the fields.

(This is more or less what I pictured in my head earlier in this PR's development, but writing it out was time consuming. Sorry for the delayed review.)

Is this even worth it? **Probably not right now**, unless we thought we needed `ragu_core`'s help, but it looks like the approach I'm proposing here actually does not need to reach into `ragu_core`'s APIs at all. If anything, it could potentially simplify them.

I think it's safe to punt on this, and close this for now. There are some commits in here that may be worth breaking out into a separate PR, though.

crates/ragu_macros/src/derive/gadget.rs
ebfull: Note that the downside to this break is just compile-time errors (I think) or missed optimization opportunities, which is probably a tolerable failure condition.

TalDerei:
@ebfull also noted OOB that this impl was a no-op at proving time (only targets linear constraints), so that alone justifies its closure, and you're probably right about the DCE pass. there are obviously more amenable targets for fixed / random LC optimizations where that wouldn't be the case, but boolean seemed like the simplest gadget to start playing around with for this kind of optimization. closing accordingly.

Issue #56, opened 2025-10-21
Endoscaling convergence problem


TalDerei:
We’d like to determine how many _total_ circuits (and by extension endoscalngs) this converges to given that each circuit can only handle K endoscaling operations (K = 4 with out overflowing recursion threshold of 2^11 constraints). We can model the convergence using a geometric series.

_Update:_ the blowup factor was determined to be `k / (k - 1)` which suggests this converges very quickly. We should document this somewhere.

PR #453, opened 2026-02-07
book: endoscaling convergence problem

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/56 and extracts an internal research writeup verbatim about endoscalars directly into the book for supplemental context. Book entry needs revision.

[Endoscaling Convergence Problem.pdf](https://github.com/user-attachments/files/25155536/Endoscaling.Convergence.Problem.pdf)

### DIFF
--- a/book/src/protocol/extensions/endoscalar.md
+++ b/book/src/protocol/extensions/endoscalar.md
@@ -1,6 +1,7 @@
 # Endoscalars
 
-Introduced in the Halo protocol, an _endoscalar_ $\endo{s}\in\{0,1\}^\lambda$
+Introduced in the [Halo protocol](https://eprint.iacr.org/2019/1021), an
+_endoscalar_ $\endo{s}\in\{0,1\}^\lambda$ (where $\lambda$ is 128 or 136 bits)
 is a small binary string used to perform scalar multiplication on curves with
 an efficient endomorphism (such as both Pasta curves).
 The endoscalar space is smaller than both $\F_p$ and $\F_q$, allowing it to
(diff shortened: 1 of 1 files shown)


book/src/protocol/extensions/endoscalar.md
```diff
@@ -36,3 +36,58 @@ We can move $\alpha$ across circuits via an endoscalar:
 first, run $\mathsf{extract}$ in the $\F_p$-circuit to obtain the endoscalar as a
 public output; then use the same endoscalar as the public input of the
 $\F_q$-circuit and constrain $\endo{s}\cdot G$ completely natively.
+
+## Convergence Problem
+
+In a curve cycle, endoscalars are scalar multiplications where the scalar is from the foreign
```
alxiong:
```suggestion
In a curve cycle, endoscaling is a scalar multiplication where the scalar is from the foreign
```

book/src/protocol/extensions/endoscalar.md
```diff
@@ -1,6 +1,7 @@
 # Endoscalars
 
-Introduced in the Halo protocol, an _endoscalar_ $\endo{s}\in\{0,1\}^\lambda$
+Introduced in the [Halo protocol](https://eprint.iacr.org/2019/1021), an
+_endoscalar_ $\endo{s}\in\{0,1\}^\lambda$ (where $\lambda$ is 128 or 136 bits)
```
ebfull: This is weirdly specific, though I see what you're trying to say (it's a security level parameter akin to what people would expect takes the value 128 or thereabouts).

ebfull (approved):
Additive, prefer to merge.

PR #454, opened 2026-02-07
book: nullifier derivation scheme

TalDerei:
Extracts an internal research writeup verbatim about a nullifier derivation scheme directly into the book. Although this may be premature, perhaps material like this should live under a dedicated application circuits section that we’ll eventually need anyway (or maybe this should live in the tachyon book). Book entry needs revision.

[Nullifier_Derivation_Scheme.pdf](https://github.com/user-attachments/files/25155636/Nullifier_Derivation_Scheme.pdf)


ebfull:
This should exist in the tachyon book, I think.

Issue #22, opened 2025-10-09
Compute bounds for Ragu aggregated verification versus Orchard batch verification


TalDerei:
We hypothesis that Ragu verification is _probably_ roughly an order of magnitude slower than Orchard in raw verification terms. For context, the current orchard verifier is ~10-15ms IIRC on a single thread. 

Ragu has the following cost drivers: the circuit polynomial has a degree of 4N linear constraints, where N = 2^K and K = 11 (preferred recursion threshold), which corresponds to a multi-exponentiation size of N = 4 * 2^11 = 2^13, which largely determines the verifier’s cost for verifying a single recursive proof batch. We’re also operating on a cycles of curves, which will naturally be slower than a single curve setting.

The provoking question here is when does orchard batch verification become worse than Ragu aggregated verification. The amortization properties of Ragu afford that at some threshold and number of transactions per block, verifying one aggregated Ragu proof beats batch-verifying N independent Orchard proofs. 

If we fix the recursion threshold 2^K and compare costs: 
- Orchard batch = `N_tx × 10ms / log(N_tx)`
- Ragu = estimated 100ms constant based on 2^K

then N = ~50 transactions where Ragu becomes more efficient. This is a _rough_ estimate that we should eventually calculate more stringently. There also appears to be an interesting tradeoff space: the recursion threshold appears to be directly proportional to the number of transactions required for aggregated verification to become more efficient. For example, doubling the recursion threshold roughly halves both the verification time and the number of transactions needed to reach the break-even point. However, a smaller N implies both lower verification times and deeper PCD trees, which implies more recursive steps compared to shallower trees.

PR #455, opened 2026-02-07
book: orchard versus ragu verification bounds

TalDerei:
Basically fed the issue description from https://github.com/tachyon-zcash/ragu/issues/22 into an LLM and asked it to create a book entry; just a placeholder. Book entry likely needs revision.


ebfull (approved):
Additive, prefer to merge.

Issue #13, opened 2025-10-08
Optimize concurrent evaluations mesh polynomials at different points.


ebfull:
The mesh polynomial will sometimes be (partially) evaluated at different points at the same time during proof generation. As an example, we may need to compute:

```rust
let s1 = [
    CommittedPolynomial::new(mesh.wx(w, acc1.x), &mut thread_rng()),
    CommittedPolynomial::new(mesh.wx(w, acc2.x), &mut thread_rng()),
];
```

which involves computing the mesh polynomial with the same `W` restriction and two different `X` restrictions. (This also emerges in other contexts with different restrictions.) If restrictions are shared (or the structure of the computation is the same) they can share computational effort, meaning the `Mesh` API should provide a way to supply a _matrix_ of points for evaluation.

PR #458, opened 2026-02-09
feat(circuits): registry evaluation cache 

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/13. Rather than implementing the matrix-based API suggested in the issue, we do something slightly simpler. This introduces a `RegistryAt` struct that binds the registry to a specific `W` point and caches the Lagrange coefficients. `Registry::at(w)` returns a `RegistryAt`, which provides `wx`, `wy`, and `wxy` methods that reuse the cached coefficients. Convenience methods on registry delegate through this API. Consequently, fuse can cache the w-restriction once and evaluate at multiple x/y points without recomputing lagrange coefficients (scalars), replacing three separate computations with one.

### DIFF
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -401,14 +401,6 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
         coeffs
     }
 
-    /// Returns true if the circuit's $\omega^j$ value is in the registry domain.
-    ///
-    /// See [`CircuitIndex::omega_j`] for details on the $\omega^j$ mapping.
-    pub fn circuit_in_domain(&self, i: CircuitIndex) -> bool {
-        let w: F = i.omega_j();
-        self.domain.contains(w)
-    }
-
     /// Index the $i$th circuit to field element $\omega^j$ as $w$, and evaluate
     /// the registry polynomial unrestricted at $X$.
     ///
@@ -417,6 +409,14 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
     pub fn circuit_y(&self, i: CircuitIndex, y: F) -> structured::Polynomial<F, R> {
         let w: F = i.omega_j();
         self.at(w).wy(y)
+    }
+
+    /// Returns true if the circuit's $\omega^j$ value is in the registry domain.
+    ///
+    /// See [`CircuitIndex::omega_j`] for details on the $\omega^j$ mapping.
+    pub fn circuit_in_domain(&self, i: CircuitIndex) -> bool {
+        let w: F = i.omega_j();
+        self.domain.contains(w)
     }
 
     /// Evaluate the registry polynomial unrestricted at $X$.


crates/ragu_circuits/src/registry.rs
```diff
@@ -308,6 +308,20 @@ pub struct Registry<'params, F: PrimeField, R: Rank> {
...
+/// The cache is tied to the specific `w` value used to create it.
+pub enum LagrangeCache<F> {
+    /// Must interpolate across circuits (w not in domain).
+    Interpolate(Vec<F>),
+    /// Direct circuit lookup (w in domain).
+    Direct(usize),
+    /// No circuit at this point.
+    Empty,
+}
+
```
TalDerei: enum makes the branching logic explicit through match rather than nested if-else

TalDerei:
> meta question: do we want to make uncached `Registry::w()` private (or even test only api)? This would nudge all usage to be `w_cached()` which has no down side afaict.

Ideally they should have pub(crate) visibility, but found that there's unused method warnings.

crates/ragu_circuits/src/registry.rs
```diff
@@ -558,6 +603,49 @@ mod tests {
...
+        let cache = registry.cache_lagrange(w);
+
+        assert_eq!(
+            registry.wx_cached(&cache, x).eval(eval_point),
+            registry.wx(w, x).eval(eval_point)
+        );
+        assert_eq!(
+            registry.wy_cached(&cache, y).eval(eval_point),
+            registry.wy(w, y).eval(eval_point)
+        );
```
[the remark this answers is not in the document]
TalDerei: added assertion

TalDerei:
currently failing on main because of trait bounds introduced in https://github.com/tachyon-zcash/ragu/pull/459.

crates/ragu_circuits/src/registry.rs
```diff
@@ -440,48 +413,32 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+        let w: F = i.omega_j();
+        self.wy(w, y)
+    }
 
-            let digest_state = hasher.finalize();
-            w = field_from_hash(&digest_state, 0);
-            x = field_from_hash(&digest_state, 1);
-            y = field_from_hash(&digest_state, 2);
+    /// Evaluate the registry polynomial unrestricted at $X$.
+    pub fn wy(&self, w: F, y: F) -> structured::Polynomial<F, R> {
```
TalDerei: not personally a fan of having two sets of callers, but the alternative is every caller needs to explicitly pass in a cache (as apposed to having these convenience wrappers).

alxiong (approved):

TalDerei:
> `OmegaKey` maps field elements to `u64` via truncation, which is only collision-free for elements of multiplicative order 2 S (see `OmegaKey::from`). This code short-circuits by consulting the map first, when the original code only consulted it when `domain.ell` indicated the element was in the domain. A random `w` could collide with a root of unity and produce false results with some non-negligible probability.

really good catch, switched the ordering there! modified to match the original if-else chain.

alxiong:
I think simple reorder is not enough. 
the fact that we missed it in the first place implies a style rule: "Encoding invariants in control flow isn't explicit enough, refactor or add test to expose it"

TalDerei:
> I think simple reorder is not enough. the fact that we missed it in the first place implies a style rule: "Encoding invariants in control flow isn't explicit enough, refactor or add test to expose it"

added a test that exercises the omega key collision using byte manipulation (couldn't think of another way to test it).

crates/ragu_circuits/src/registry.rs
```diff
@@ -595,6 +655,95 @@ mod tests {
...
+
+        let x = Fp::from(42u64);
+        let y = Fp::from(43u64);
+        assert_ne!(registry.at(w).wxy(x, y), registry.at(omega).wxy(x, y));
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_omega_key_collision() -> Result<()> {
```
alxiong: ACK, test logic looks good. Add some comment with pointers to OmegaKey::from doc, and explain why the potential error.

crates/ragu_circuits/src/registry.rs
```diff
@@ -398,76 +409,125 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+    /// Index the $i$th circuit to field element $\omega^j$ as $w$, and evaluate
+    /// the registry polynomial unrestricted at $X$.
+    ///
+    /// Wraps [`Registry::at`] and [`RegistryAt::wy`].
+    /// See [`CircuitIndex::omega_j`] for more details.
+    pub fn circuit_y(&self, i: CircuitIndex, y: F) -> structured::Polynomial<F, R> {
+        let w: F = i.omega_j();
+        self.at(w).wy(y)
+    }
+
```
ebfull: Diff would be smaller if this function were located where it originally was.

ebfull (approved):

Issue #316, opened 2026-01-06
Cache circuit constraint structures for fast mesh evaluation


TalDerei:
Both mesh-key computation (#78) and verification require mesh polynomial evaluation, and currently each call re-synthesizes all circuits. This becomes a performance bottleneck when the mesh must be evaluated many times. We should optimize by running synthesis once with a driver to capture the constraint structure, then evaluate directly from runtime _cached_ structure. We may also consider using BLAKE2b instead of Poseidon here (via `blake2b_simd`, a dependency already supported in [Zebra](https://github.com/ZcashFoundation/zebra/blob/main/Cargo.toml)).

alxiong:
Putting aside registry poly evaluation from `compute_registry_digest()` (as it was subject to changes), 

evals from `fuse()` and `verify()` are as follows: 

<details>
<summary>AI Summary of `wx/wy/wxy` calls</summary>

```
  fuse() — Native registry (native_registry)

  ┌────────────────────────┬────────────┬───────────────────────────────────────────────────┬───────────────────────────┬─────────┐
  │       File:Line        │   Method   │                        Via                        │          Reason           │  Count  │
  ├────────────────────────┼────────────┼───────────────────────────────────────────────────┼───────────────────────────┼─────────┤
  │ fuse/_03_s_prime.rs:34 │ wx(w, x0)  │ direct                                            │ left child x restriction  │ 1       │
  ├────────────────────────┼────────────┼───────────────────────────────────────────────────┼───────────────────────────┼─────────┤
  │ fuse/_03_s_prime.rs:38 │ wx(w, x1)  │ direct                                            │ right child x restriction │ 1       │
  ├────────────────────────┼────────────┼───────────────────────────────────────────────────┼───────────────────────────┼─────────┤
  │ fuse/_04_error_m.rs:54 │ wy(w, y)   │ direct                                            │ error-M w restriction     │ 1       │
  ├────────────────────────┼────────────┼───────────────────────────────────────────────────┼───────────────────────────┼─────────┤
  │ fuse/_07_query.rs:54   │ xy(x, y)   │ direct                                            │ registry_xy polynomial    │ 1       │
  ├────────────────────────┼────────────┼───────────────────────────────────────────────────┼───────────────────────────┼─────────┤
  │ claims/mod.rs:93       │ wy(ω_j, y) │ circuit_y → circuit_impl → app circuit × 2 proofs │ app circuit claims        │ 2×N_app │
[...]

PR #459, opened 2026-02-09
chore(circuits): replace Poseidon with BLAKE2b for registry digest

TalDerei:
(Partially) addresses https://github.com/tachyon-zcash/ragu/issues/316. Since the registry digest isn't computed in-circuit, we can replace the slower `Poseidon` sponge with faster native `blake2b_simd`.  This change also removes the `Poseidon` parameter from `RegistryBuilder::finalize` and adds `FromUniformBytes<64>` bound to the `Cycle` trait.

### DIFF
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -451,8 +451,16 @@ impl<F: PrimeField + FromUniformBytes<64>, R: Rank> Registry<'_, F, R> {
     fn compute_registry_digest(&self) -> F {
         let mut hasher = Params::new().personal(b"ragu_registry___").to_state();
 
-        let field_from_hash = |hasher: &blake2b_simd::State, index: u8| {
-            F::from_uniform_bytes(hasher.clone().update(&[index]).finalize().as_array())
+        let field_from_hash = |digest_state: &blake2b_simd::Hash, index: u8| {
+            F::from_uniform_bytes(
+                Params::new()
+                    .personal(b"ragu_registry___")
+                    .to_state()
+                    .update(digest_state.as_bytes())
+                    .update(&[index])
+                    .finalize()
+                    .as_array(),
+            )
         };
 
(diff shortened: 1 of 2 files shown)


crates/ragu_circuits/src/registry.rs
```diff
@@ -444,31 +444,49 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+    fn field_from_hash(hash: &blake2b_simd::Hash, index: u8) -> F {
+        let derived = Params::new()
+            .hash_length(64)
+            .personal(b"ragu_derive_____")
+            .to_state()
+            .update(hash.as_bytes())
+            .update(&[index])
+            .finalize();
+
+        F::from_uniform_bytes(derived.as_bytes().try_into().expect("64 bytes"))
```
TalDerei: standard hash-to-field construction, and `halo2` and `pasta_curves` use `from_uniform_bytes` internally. domain separated hash to 64-bytes, and then wide reduction mod p.

crates/ragu_arithmetic/src/lib.rs
```diff
@@ -106,10 +106,10 @@ pub use u128 as Uendo;
 pub trait Cycle: Copy + Clone + Default + Send + Sync + 'static {
     /// The field that circuit developers will primarily work with, and the
     /// scalar field of the [`HostCurve`](Cycle::HostCurve).
-    type CircuitField: WithSmallOrderMulGroup<3>;
+    type CircuitField: WithSmallOrderMulGroup<3> + FromUniformBytes<64>;
 
     /// The scalar field of the [`NestedCurve`](Cycle::NestedCurve).
-    type ScalarField: WithSmallOrderMulGroup<3>;
+    type ScalarField: WithSmallOrderMulGroup<3> + FromUniformBytes<64>;
```
TalDerei: without these trait bounds on the `Cycle` trait, we'd have nasty cascading trait bounds and wasn't sure how to get around.

ebfull (approved):

Issue #78, opened 2025-11-04
Formalize safety requirements for mesh commitment


TalDerei:
We’ve discussed likely needing to commit to the mesh before starting the accumulation procedure. Intuitively this makes sense for binding purposes where the mesh can't be changed later, but we should delineate the explicit safety properties for doing this.

TalDerei:
> I will append registry key to transcript in the [#422](https://github.com/tachyon-zcash/ragu/pull/422)

why wouldn't the suggestion work? the mesh polynomial structure depends on circuit constraints, not runtime wire values. so it shouldn't conflict with sponge initialization, which is supplied at runtime (instead of `RAGU_TAG`, we'd supply runtime `key_wire`).

PR #460, opened 2026-02-10
feat(circuits): cache circuit constraint structure

TalDerei:
(Partially) addresses https://github.com/tachyon-zcash/ragu/issues/316 (other half of https://github.com/tachyon-zcash/ragu/pull/459 which is independent). As the issue suggested "mesh-key computation (https://github.com/tachyon-zcash/ragu/issues/78) and verification require mesh polynomial evaluation, and currently each call re-synthesizes all circuits. This becomes a performance bottleneck when the mesh must be evaluated many times. We should optimize by running synthesis once with a driver to capture the constraint structure", this does exactly that. 

Registration time captures the circuit constraint structure once (stores `SynthesisTrace` in `ProcessedCircuit` during `into_object() `call), and subsequent `sxy` evaluations replay the cached trace (instead of re-executing the circuit witness). It extends `Counter` driver to also record traces, so `metrics::eval()` returns `(CircuitMetrics, SynthesisTrace)` (so single synthesis pass does both metrics counting and generates synthesis trace). 

The key constraint requires special handling: the key constraint is enforced at evaluation time (as before), but since the trace is captured before the registry key is known, it needs to be enforced explicitly before replaying the trace.

**Note:** only currently applying to `sxy`, and `sx / sy` still run full synthesis every time. I added `sxy_poseidon_repeated` bench which calls `sxy` on poseidon circuit multiple times, and only found ~15% improvement over main, suggesting that synthesis overhead was only ~10-15% of the total cost which I thought would be more of a dominant cost. we're still running the field ops for `sxy` which dominates the cost. curious to see the improvements from parallelizing the sxy evaluation with rayon, and obviously extending to `sx / sy`.

**Need to think through how this interplay's with the routine memoization work.**

### DIFF
--- a/crates/ragu_circuits/src/metrics.rs
+++ b/crates/ragu_circuits/src/metrics.rs
@@ -40,11 +40,8 @@ pub(crate) struct TraceTerm<F> {
     pub wire: usize,
 }
 
-/// Sum of [`TraceTerm`]s.
-#[derive(Clone, Debug)]
-pub(crate) struct TraceLC<F> {
-    pub terms: Vec<TraceTerm<F>>,
-}
+/// A linear combination of [`TraceTerm`]s.
+pub(crate) type TraceLC<F> = Vec<TraceTerm<F>>;
 
 /// Recorded circuit structure for polynomial evaluation replay.
 #[derive(Clone, Debug, Default)]
@@ -74,9 +71,7 @@ impl<F: Field> Default for TraceLCBuilder<F> {
 
 impl<F: Field> TraceLCBuilder<F> {
     fn build(f: impl Fn(Self) -> Self) -> TraceLC<F> {
-        TraceLC {
-            terms: f(Self::default()).terms,
-        }
+        f(Self::default()).terms
     }
 }
 
@@ -96,24 +91,13 @@ impl<F: Field> LinearExpression<usize, F> for TraceLCBuilder<F> {
     }
 }
 
+#[derive(Default)]
 struct Counter<F: Field> {
     next_wire_id: usize,
     mul_wire_ids: Vec<(usize, usize, usize)>,
     add_wires: Vec<(usize, TraceLC<F>)>,
     constraints: Vec<TraceLC<F>>,
     available_b: Option<usize>,
-}
-
-impl<F: Field> Default for Counter<F> {
-    fn default() -> Self {
-        Self {
-            next_wire_id: 0,
-            mul_wire_ids: Vec::new(),
-            add_wires: Vec::new(),
-            constraints: Vec::new(),
-            available_b: None,
-        }
-    }
 }
 
 impl<F: Field> Counter<F> {
--- a/crates/ragu_circuits/src/s/sxy.rs
+++ b/crates/ragu_circuits/src/s/sxy.rs
@@ -102,7 +102,6 @@ pub fn eval<F: Field, R: Rank>(trace: &SynthesisTrace<F>, x: F, y: F, key: &regi
     // Evaluate add wires that were lazily deferred during trace capture
     for (id, lc) in &trace.add_wires {
         let sum: F = lc
-            .terms
             .iter()
             .map(|term| term.coeff * wire_evals[term.wire])
             .sum();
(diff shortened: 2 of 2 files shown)


crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -302,20 +318,34 @@ pub fn eval<F: Field, C: Circuit<F>, R: Rank>(
...
-    evaluator.enforce_one()?;
+    // Allocate all wires by calling mul()
+    let wires: Vec<WireEval<F>> = (0..trace.num_mul_gates)
+        .flat_map(|_| {
+            let (a, b, c) = evaluator
+                .mul(|| unreachable!())
+                .expect("mul should not fail");
+            [a, b, c]
+        })
+        .collect();
```
TalDerei: basically pre-allocates all wires upfront (replays same number of `mul` calls that were recorded in the trace).

crates/ragu_circuits/src/lib.rs
```diff
@@ -109,15 +109,18 @@ pub trait CircuitExt<F: Field>: Circuit<F> {
...
         }
 
-        struct ProcessedCircuit<C> {
+        // Capture synthesis trace once when circuit object is created.
+        let synthesis_trace = s::sxy::capture_trace(&self)?;
+
+        struct ProcessedCircuit<F: Field, C> {
             circuit: C,
             metrics: metrics::CircuitMetrics,
+            synthesis_trace: SynthesisTrace<F>,
```
TalDerei: now passing the synthesis trace inside of circuit object.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -86,37 +179,43 @@ impl<'dr, F: Field> Driver<'dr> for Counter<F> {
...
+    // The key constraint is NOT recorded - it's handled at eval time.
+    let (_key_wire, _, _one) = counter.mul(|| unreachable!())?;
+
+    let mut outputs = Vec::new();
+    let (io, _) = circuit.witness(&mut counter, Empty)?;
+    io.write(&mut counter, &mut outputs)?;
+
+    // Enforcing public inputs
+    counter.enforce_public_outputs(outputs.iter().map(|output| output.wire()))?;
+    counter.enforce_one()?;
```
TalDerei: key enforcement is inlined and now happens here.

crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -35,287 +34,98 @@
...
+    let mut wire_evals: Vec<F> = Vec::with_capacity(num_wires);
+    wire_evals.resize(num_wires, F::ZERO);
+
+    // Evaluate mul wires
+    for (a_id, b_id, c_id) in &trace.mul_wire_ids {
+        let (a, b, c) = monomials.next_gate();
+        wire_evals[a_id.0] = a;
+        wire_evals[b_id.0] = b;
+        wire_evals[c_id.0] = c;
+    }
```
TalDerei: basically pre-allocates all wires upfront (replays same number of `mul` calls that were recorded in the trace).

crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -35,287 +34,98 @@
...
-    io.write(&mut evaluator, &mut outputs)?;
+    // Evaluate add wires (each references only earlier wires)
+    for (id, lc) in &trace.add_wires {
+        let sum: F = lc
+            .terms
+            .iter()
+            .map(|term| term.coeff.value() * wire_evals[term.wire.0])
+            .sum();
+        wire_evals[id.0] = sum;
+    }
```
TalDerei: evaluates add wires that were lazily deferred during trace capture (without this, we'd get memory blowup and OOM errors).

TalDerei:
this was arguably an experimental exercise to understand the synthesis costs outside of pure field arithmetic. surprisingly, traversing the circuit structure consistutes ~10-15% of the total cost. closing as unplanned, but we should investigate the performance here after memoization is in place. cc @ebfull @alxiong

Issue #217, opened 2025-12-09
Macroize the unified instance


TalDerei:
Currently, appending a new field to the [UnifiedInstance](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_pcd/src/internal_circuits/unified.rs) requires manually updating multiple call sites: the gadget definition, `Instance`, `Output`, builder (`OutputBuilder` struct, `OutputBuilder::new()`, `OutputBuilder::finish()`), which is error prone. 

Instead, we might want to consider a declarative macro where you define the field once, and it generates all the corresponding code in each location:

```ts
unified_instance! {
    c_rx_commitment: Point,
    alpha: Challenge,
    ...
}
```

Moreover, circuit code that has verbose boilerplate probably warrants either a macro or a closure abstraction:

```ts
  pub fn emulate_u<C: Cycle>(
      alpha: C::CircuitField,
      nested_f_commitment: C::NestedCurve,
      params: &C,
  ) -> Result<C::CircuitField> {
      Emulator::emulate_wireless((alpha, nested_f_commitment), |dr, witness| {
          let (alpha, comm) = witness.cast();
          let alpha_elem = Element::alloc(dr, alpha)?;
          let point = Point::alloc(dr, comm)?;
          Ok(*derive_u::<_, C>(dr, &alpha_elem, &point, params)?
              .value()
              .take())
      })
  }
```

PR #461, opened 2026-02-10
macros: unified_instance! for Output, Instance, OutputBuilder codegen

alxiong:
An alternative function-like proc macro approach to #217, as a comparison to #447, 

addressing my own comment in https://github.com/tachyon-zcash/ragu/pull/447#discussion_r2780360495. I originally propose to use `#[derive(UnifiedInstance)` proc macro, but that makes less sense because we never use that struct. A function proc macro is more appropriate here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,6 +519,7 @@ dependencies = [
  "ragu_arithmetic",
  "ragu_circuits",
  "ragu_core",
+ "ragu_macros",
  "ragu_pasta",
  "ragu_primitives",
  "rand",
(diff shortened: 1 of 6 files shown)


TalDerei (commented):
personally still echoing sentiment from https://github.com/tachyon-zcash/ragu/pull/447#discussion_r2788930871

TalDerei:
closing in favor of https://github.com/tachyon-zcash/ragu/pull/447#pullrequestreview-3787364474 for now, and we can eventually revisit a more holistic declarative versus procedural macro pass.

PR #465, opened 2026-02-13
chore(circuits): implement RAII based API

TalDerei:
References https://github.com/tachyon-zcash/ragu/pull/413#discussion_r2795925066, a latent synthesis bug that effected our existing code. Implements `RestoreGuard<T>`, a RAII guard across `rx / sx / sy / sxy / metrics` callers. Initially, the RAII impl was fighting the borrow checker by holding `&mut self.available_b` while also needing `&mut self`. I modified it to convert  `&mut T` to `*mut T` (raw pointer), which ends the borrow checker's tracking and requires marking with `unsafe`. RAII ensures the wire value is restored on drop (especially on early `?` returns / panics), preventing situations which were possible before where routines can leak allocation state across boundaries.

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -46,14 +46,14 @@ use polynomials::{Rank, structured, unstructured};
 /// A trait for drivers that stash a spare wire from paired allocation (see
 /// [`Driver::alloc`]).
 ///
-/// Provides [`Self::with_fresh_b`], which saves [`Self::available_b`], resets it to its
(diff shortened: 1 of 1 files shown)


ebfull (changes requested):
NACK, commit history is a bit confusing after the squash. Let's also _link_ to `Self::with_fresh_b` rather than explicitly stating `Self::with_fresh_b`. (This could be added to my picky standards I guess; `Self` should rarely be stated in documentation.)

TalDerei:
> The commit title for `7f8ffd244c05f695e8112da4d2fbd94673447683` still does not reflect what's actually in the commit. The `unsafe` code has been removed retroactively.

rebased accordingly, will more carefully check rebases along these axis going forward

ebfull (approved):

Issue #63, opened 2025-10-22
Implement parallel witness computation with predict


TalDerei:
Synergetic with https://github.com/tachyon-zcash/ragu/issues/55; Currently, in [rx](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_circuits/src/rx.rs), the routine _ignores_ the predicted output and executes regardless (predictions aren't currently used in the rx witness collector). `Predict()` enables:

- (a) computing an `output` much faster, allowing the circuit to continue execution without blocking on expensive interstitial witness computations, 
- (b) spawns a thread to compute intermediate `aux` values that `execute()` can use as hints. Execute eventually joins the thread and retrieves the aux, and performs witness generation by invoking rx driver methods. 

Witness generation could be overlapped, but is currently fully _sequential_. Specifically, we can pipeline `predict` to have `aux` computations run in parallel, and defer execute until the end of the circuit / pipelined where it will be invoked in parallel to perform witness generation.

Issue #18, opened 2025-10-08
Determine `no_std` policy and feature-gated parallelization strategy


TalDerei:
Is the intention for Ragu to remain `no_std` (ie. a standalone artifact that doesn’t depend on the Rust standard library), which is the current status quo? That's antithetical to and precludes parallelization techniques like Rayon threading (for multi-circuit witness synthesis, MSMs, polynomial evaluations, and so forth), which require `std` support. Should we gate that optimization behind feature flags and enable it conditionally?

PR #476, opened 2026-02-16
feat(circuits): async witness generation

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/63. 

draft; experiments with deferred witness generation, where circuit _doesn't_ block on routine execution like before (execute is deferred and batched, but can also possibly be parallelized with rayon in https://github.com/tachyon-zcash/ragu/issues/18 or reorganized based on floor planning). It's currently only batched and at least achieves better cache locality than the status quo impl.                                                                                                                                                                                                                         

specifically, for the `rx` driver, circuit traversal (`circuit.witness()`) can proceed more efficiently without blocking on routine execution. direct driver calls (`dr.alloc(), dr.mul()`) execute immediately, while routine calls (`dr.routine()`) with known predictions defer their `execute()` to the end. `Aux` cannot be deferred because both output computation and execute depend on it. anyways, the side-effect here is circuit traversal is decoupled from expensive and blocking execute work. 

obviously it's not this simple, because reordering calls breaks things. we'd need _some_ approach when deferring that, loosely speaking: record the current starting index/position, reserve slots, store the deferred closure at that starting index, then flush to overwrite the reserved slots. Interestingly, that two-phase reservation/filling mechanism has parallels with our staging system. 

_**update:**_ rebased over https://github.com/tachyon-zcash/ragu/pull/483 which essentially provides that exact scaffolding I was looking for. now, the deferred execute pipeline is: when calling a routine with a known prediction, it will create an empty segment (reserves its position in the floor planner), defers the actual execute and fills in the actual gate values during the flush phase at the end when the execute's closure runs. for unknown predictions, we similarly create the segment, run execute inline (no deferral). the whole point of this is just to establish a deferral pattern. the deferrals still run sequentially, but we could eventually perform actual parallel execution.

### DIFF
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -384,8 +384,8 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
 
     /// Returns the cached subtree sizes for a registered circuit.
     ///
-    /// These can be passed to [`rx::eval`](crate::rx::eval) to avoid
-    /// recomputing metrics.
+    /// These can be passed to [`CircuitExt::rx_with`]
+    /// to avoid recomputing metrics.
     pub fn subtree_sizes(&self, circuit: CircuitIndex) -> &[usize] {
         self.circuits[usize::from(circuit)].subtree_sizes()
     }


crates/ragu_circuits/src/rx.rs
```diff
@@ -246,8 +264,25 @@ impl<'a, F: Field> Driver<'a> for Evaluator<'a, F> {
...
+                        this.pending
+                            .push(Box::new(move |dr: &mut Evaluator<'a, F>| {
+                                dr.with_scope(
+                                    EvalState {
+                                        available_b: None,
+                                        current_segment: seg,
+                                    },
+                                    |dr| routine.execute(dr, input, aux).map(|_| ()),
+                                )
+                            }));
```
ebfull: When `routine.execute()` runs during the flush phase (later), any nested `dr.routine()` calls within it will append segments to the end of the trace, out of their correct DFS positions. The captured segment index is correct for _this_ routine's segment (reserved above), but nested routines weren't traversed during predict (it ran on a wireless emulator) so their segments were never reserved.

TalDerei:
> Perhaps we could throw Claude at this to make a test suite triggering all these goofy edge cases

Generated combination of edge cases exercising:
- DFS misordering: `Known` outer + `Unknown` nested inner 
- Flush bug: `Known` Outer + `Known` nested inner 

Added these tests to demonstrate failure against the **current** code.

https://github.com/tachyon-zcash/ragu/pull/476/changes/32e267083a4e3dc6fe2a92ba007ffaf09ae8611a, https://github.com/tachyon-zcash/ragu/pull/476/changes/24bc72388e38ccbaa68b570829de2b1fca18930b, https://github.com/tachyon-zcash/ragu/pull/476/changes/3e819629b231b51903d6eb7e0750f40b19c02fde.  

Prior to reading through your proposal, I was entertaining a simpler stopgap where the loop keeps draining until pending is empty (basically processing pending entries as a queue), but that’s naively only a surface level fix. It addresses the issue where pending items get orphaned, but the DFS ordering issue we identified is inherently _structural_, so it wouldn’t be addressed purely by flushing.

TalDerei:
@ebfull this diverges from the initial proposal (recovering the canonical DFS ordering without a metrics pass). the approach taken here sources inspiration from the `sxy` evaluator (which indexes into the floor planner). similarly, we index into the floor planner by using the metrics pass (`CircuitMetrics` which contains full structural DFS) as external advice into the `rx` evaluator. the evaluator pre-allocates all segments upfront, and then for `Known` routines, skips past the subtree (of nested routines) using pre-computed subtree sizes. witness generation remains sequential, and parallel flush is follow-up work.

TalDerei:
_Note:_ we can ignore last two commits if it simplifies review (`rx_with` is technically unnecessary and was wired up there to bypass the redundant metrics pass in `rx`).

TalDerei:
closing in favor of https://github.com/tachyon-zcash/ragu/pull/519/changes, https://github.com/ebfull/ragu/pull/1, https://github.com/ebfull/ragu/pull/2. there _may_ be forces we can't predict that lead us to shift approaches in the future and frontload as much of the structural knowledge the floor planner provides as possible. this can remain a design artifact for future reference. cc @ebfull @alxiong

PR #483, opened 2026-02-17
Floor planner scaffolding

ebfull:
## Background

`Routine`s are intended to naturally induce a correspondence between isolated portions of synthesis (sequences of multiplication gates and constraints that are mainly over _those_ newly introduced wires) and the resulting portions of $s(X, Y)$ that can then be memoized and manipulated by simple linear transformations. Routines can then be invoked multiple times with less effort.

In order to get the most out of this, we must also rearrange those contributions to $s(X, Y)$ in each circuit so that they tend to line up across circuits in the `Registry`. This means the `Registry::finalize` method must apply an optimization pass (a **floor planner**) that finds ideal placements for the routine's $s(X, Y)$ contributions for each individual circuit in the context of the others. This floor plan can be consumed by the evaluation drivers later to efficiently evaluate $s(X, Y)$.

@TalDerei's recent #413 introduces some infrastructure for this floor planning concept. I wanted to play with an orthogonal portion of the infrastructure -- mainly the API, but also some of the algebra involved in the floor plan -- in order to create a minimal change that sets the groundwork for the floor planner.

### Scope

* This doesn't actually introduce a floor planning algorithm. The "floor planner" in this PR preserves synthesis order of the routines.
* This doesn't introduce a routine identity pass.
* This doesn't introduce any memoization.

In other words, it does not really overlap at all with #413.

## Goal

The main goal here is to make an adjustment to the algebra of how routines map to portions of $s(X, Y)$. This emerges out of my tendency to prefer development paths that increase the odds of finding bugs _during transitions_, either through testing or just by accident.

Right now, as we synthesize routines their $s(X, Y)$ contributions materialize in synthesis-order in the wiring polynomial. We want the future floor planner to be able to fix certain routines in particular spots depending on how identical routines are arranged in different circuits. The approach in #413 is to encounter a routine, fix it, and then treat nested routines as part of the same routine _for floor planning purposes_. (I know that this is intended as a temporary stop-gap in the implementation, since the focus on #413 is to make progress on other parts of the development path.)

This general approach is probably unavoidable if we're to maintain the simple linear quality of the routines' $s(X, Y)$ contributions. But it does not mean the routines need to be embedded within each other, just fixed relative to each other. By decoupling this, two things could be more simple: the evaluators only need to care about the scaling to apply to each routine called, and the witness generation logic just needs to follow a simple floor plan to rearrange the locations of the trace for each routine.

All the hard parts should be in the floor planner.

### DIFF
--- a/crates/ragu_circuits/src/rx.rs
+++ b/crates/ragu_circuits/src/rx.rs
@@ -138,7 +138,7 @@ impl<F: Field> Trace<F> {
                 let slot = &floor_plan[seg_idx];
 
                 // Verify segment size matches floor plan expectation.
-                debug_assert_eq!(
+                assert_eq!(
                     seg.a.len(),
                     slot.num_multiplication_constraints,
                     "segment {} size must match floor plan",
--- a/crates/ragu_circuits/src/s/sx.rs
+++ b/crates/ragu_circuits/src/s/sx.rs
@@ -289,12 +289,12 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
             let result = routine.execute(this, input, aux)?;
 
             // Verify this routine consumed exactly the expected constraints.
-            debug_assert_eq!(
+            assert_eq!(
                 this.scope.multiplication_constraints,
                 slot.multiplication_start + slot.num_multiplication_constraints,
                 "routine multiplication constraint count must match floor plan"
             );
-            debug_assert_eq!(
+            assert_eq!(
                 this.scope.linear_constraints,
                 slot.linear_start + slot.num_linear_constraints,
                 "routine linear constraint count must match floor plan"
@@ -381,17 +381,17 @@ pub fn eval<F: Field, C: Circuit<F>, R: Rank>(
     evaluator.enforce_one()?;
 
     // Verify all floor plan slots were consumed and counts match.
-    debug_assert_eq!(
+    assert_eq!(
         evaluator.current_routine + 1,
         evaluator.floor_plan.len(),
         "floor plan routine count must match synthesis"
     );
-    debug_assert_eq!(
+    assert_eq!(
         evaluator.scope.multiplication_constraints,
         evaluator.floor_plan[0].num_multiplication_constraints,
         "root multiplication constraint count must match floor plan"
     );
-    debug_assert_eq!(
+    assert_eq!(
         evaluator.scope.linear_constraints, evaluator.floor_plan[0].num_linear_constraints,
         "root linear constraint count must match floor plan"
     );
--- a/crates/ragu_circuits/src/s/sy.rs
+++ b/crates/ragu_circuits/src/s/sy.rs
@@ -621,12 +621,12 @@ impl<'table, 'sy, F: Field, R: Rank> Driver<'table> for Evaluator<'table, 'sy, '
             let result = routine.execute(this, input, aux)?;
 
             // Verify this routine consumed exactly the expected constraints.
-            debug_assert_eq!(
+            assert_eq!(
                 this.scope.multiplication_constraints,
                 slot.multiplication_start + slot.num_multiplication_constraints,
                 "routine multiplication constraint count must match floor plan"
             );
-            debug_assert_eq!(
+            assert_eq!(
                 this.scope.linear_constraints,
                 slot.linear_start + slot.num_linear_constraints,
                 "routine linear constraint count must match floor plan"
@@ -732,17 +732,17 @@ pub fn eval<F: Field, C: Circuit<F>, R: Rank>(
             evaluator.enforce_one()?;
 
             // Verify all floor plan slots were consumed and counts match.
-            debug_assert_eq!(
+            assert_eq!(
                 evaluator.current_routine + 1,
                 evaluator.floor_plan.len(),
                 "floor plan routine count must match synthesis"
             );
-            debug_assert_eq!(
+            assert_eq!(
                 evaluator.scope.multiplication_constraints,
                 evaluator.floor_plan[0].num_multiplication_constraints,
                 "root multiplication constraint count must match floor plan"
             );
-            debug_assert_eq!(
+            assert_eq!(
                 evaluator.scope.linear_constraints, evaluator.floor_plan[0].num_linear_constraints,
                 "root linear constraint count must match floor plan"
             );
--- a/crates/ragu_circuits/src/metrics.rs
+++ b/crates/ragu_circuits/src/metrics.rs
@@ -135,11 +135,11 @@ impl<'dr, F: Field> Driver<'dr> for Counter<F> {
 
                 // Verify internal consistency: current_record unchanged and
                 // all paired allocations consumed.
-                debug_assert_eq!(
+                assert_eq!(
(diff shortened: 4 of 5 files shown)


crates/ragu_circuits/src/lib.rs
```diff
@@ -39,22 +43,23 @@ use alloc::{boxed::Box, vec::Vec};
...
+/// [`scope`](Self::scope), replaces it with a caller-supplied value, runs a
+/// closure with `&mut self`, then restores the original value. This isolates
+/// driver state within routines.
+pub(crate) trait DriverScope<S> {
+    /// Returns a mutable reference to the scoped state.
+    fn scope(&mut self) -> &mut S;
+
+    /// Runs `f` with [`scope`](Self::scope) temporarily replaced by `init`,
+    /// then restores the original value.
+    fn with_scope<R>(&mut self, init: S, f: impl FnOnce(&mut Self) -> R) -> R {
```
TalDerei: ACK – this stack emulation (dynamic state) is what enables the outer / inner routines to get their own contiguous blocks rather than interleaving them. what if instead of a closure pattern we modeled this using an RAII guard (restoration in `Drop`)?

crates/ragu_circuits/src/s/sy.rs
```diff
@@ -646,20 +655,42 @@ pub fn eval<F: Field, C: Circuit<F>, R: Rank>(
...
+        }
+
+        {
+            let mut evaluator = Evaluator::<'_, '_, '_, F, R> {
+                scope: SyScope {
+                    available_b: None,
+                    current_y: y.pow_vartime([(root_linear_constraints - 1) as u64]),
+                    multiplication_constraints: 0,
+                    linear_constraints: 0,
+                },
```
TalDerei: see previous comment about underflow

```ts
scope: SyScope {
    available_b: None,
    current_y: if root_linear_constraints == 0 {
        F::ONE 
    } else {
        y.pow_vartime([(root_linear_constraints - 1) as u64])
    },
    multiplication_constraints: 0,
    linear_constraints: 0,
},
```

crates/ragu_circuits/src/s/sy.rs
```diff
@@ -581,29 +574,44 @@ impl<'table, 'sy, F: Field, R: Rank> Driver<'table> for Evaluator<'table, 'sy, F
...
+        // Routine's y-power starts at y^{linear_start + num_linear_constraints - 1}
+        // and decrements through the routine's linear constraint range.
+        let init_scope = SyScope {
+            available_b: None,
+            current_y: self
+                .y
+                .pow_vartime([(slot.linear_start + slot.num_linear_constraints - 1) as u64]),
+            multiplication_constraints: slot.multiplication_start,
+            linear_constraints: slot.linear_start,
+        };
```
TalDerei: there isn't an invariant that guarantees child/nested routines always have at _least_ one linear constraint, so if child routine has no linear constraints (only `mul` gates), there's an underflow risk.

```ts
let init_scope = SyScope {
    available_b: None,
    current_y: if slot.num_linear_constraints == 0 {
        F::ONE 
    } else {
        self.y
            .pow_vartime([(slot.linear_start + slot.num_linear_constraints - 1) as u64])
    },
    multiplication_constraints: slot.multiplication_start,
    linear_constraints: slot.linear_start,
};
```

crates/ragu_circuits/src/rx.rs
```diff
@@ -14,45 +16,186 @@ use ragu_core::{
...
+    pub(crate) fn assemble_with_key<R: Rank>(
+        &self,
+        key: &registry::Key<F>,
+        floor_plan: &[RoutineSlot],
+    ) -> Result<structured::Polynomial<F, R>> {
+        debug_assert_eq!(
+            floor_plan.len(),
+            self.segments.len(),
+            "floor plan and witness must have the same number of routine entries"
+        );
```
TalDerei: what about release builds that skip validation?

```ts
        if floor_plan.len() != self.segments.len() {
            return Err(Error::FloorPlanMismatch {
                expected: self.segments.len(),
                actual: floor_plan.len(),
            });
        }
```

crates/ragu_circuits/src/rx.rs
```diff
@@ -14,45 +16,186 @@ use ragu_core::{
...
+                view.b[offset..offset + seg.b.len()].copy_from_slice(&seg.b);
+                view.c[offset..offset + seg.c.len()].copy_from_slice(&seg.c);
+            }
+
+            // Overwrite segment 0's zeroed ONE gate placeholder with
+            // actual key values.
+            view.a[0] = key.value();
+            view.b[0] = key.inverse();
+            view.c[0] = F::ONE;
+        }
```
TalDerei: assertion for per-segment size validation:

```ts
// Scatter each segment to its floor-plan position.
for (seg_idx, seg) in self.segments.iter().enumerate() {
    let slot = &floor_plan[seg_idx];

    // Verify segment size matches floor plan expectation.
    debug_assert_eq!(
        seg.a.len(),
        slot.num_multiplication_constraints,
        "segment {} size must match floor plan",
        seg_idx
    );

    let offset = slot.multiplication_start;
    view.a[offset..offset + seg.a.len()].copy_from_slice(&seg.a);
    view.b[offset..offset + seg.b.len()].copy_from_slice(&seg.b);
    view.c[offset..offset + seg.c.len()].copy_from_slice(&seg.c);
}
```

TalDerei (commented):
first review pass

crates/ragu_circuits/src/rx.rs
```diff
@@ -14,66 +16,168 @@ use ragu_core::{
...
 }
 
+/// Trace data produced by evaluating a circuit.
+///
+/// Pass to [`Registry::assemble`](crate::registry::Registry::assemble)
+/// to obtain the corresponding [`structured::Polynomial`].
+pub struct Trace<F> {
+    /// Per-routine gate groups. Segment 0 is the root; segments 1+ are
+    /// created by [`Driver::routine`] calls.
+    pub(crate) segments: Vec<Segment<F>>,
```
alxiong: why don't we use consistent `slot` for this as well? 
even though "segment" within a trace is more semantically natural, using "slot" implies familiar correspondence.

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -39,21 +39,25 @@
...
-//! The final coefficient order is:
-//! 1. $c\_0$: `ONE` wire constraint (the constant $x^{4n - 1}$)
-//! 2. $c\_1, \ldots, c\_k$: public output constraints
-//! 3. $c\_{k+1}$: registry key binding constraint
-//! 4. $c\_{k+2}, \ldots, c\_{q-1}$: circuit-specific constraints
+//! After reversal, the root routine's coefficients are ordered as:
+//! 1. $c\_{0}$: `ONE` wire constraint (the constant $x^{4n - 1}$)
+//! 2. $c\_{1}, \ldots, c\_{p}$: public output constraints
+//! 3. $c\_{p+1}, \ldots, c\_{p+m}$: circuit-specific constraints
+//! 4. $c\_{p+m+1}$: registry key binding constraint
```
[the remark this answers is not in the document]
ebfull: Claude found it.

PR #485, opened 2026-02-17
chore(testing): migrate fixtures to `ragu_testing` crate 

TalDerei:
A dedicated `ragu_testing` crate that provides shared test and bench fixtures without introducing feature flags into the main crates. Specifically, it removes `test_fixtures` references, `cfg` patterns, `unstable-test-fixtures` feature in favor of using `ragu_testing` as a dev-depdency for tests/benchmarks.

~Deferring crate registration to @ebfull. ~ (done -- sean)

### DIFF
--- a/crates/ragu_testing/README.md
+++ b/crates/ragu_testing/README.md
@@ -4,7 +4,7 @@
 
 # `ragu_testing`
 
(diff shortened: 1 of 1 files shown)


ebfull (approved):
ACK modulo nit.

PR #490, opened 2026-02-18
Replace FreshB with PairAllocatedDriver subtrait

alxiong:
A follow-up of 465, addressing my own comment: https://github.com/tachyon-zcash/ragu/pull/465#discussion_r2812352274

To conform to the default implementation of `PairAllocatedDriver` for `rx::Evaluator` and `metric::Counter`, had to change their `available_b()` return type to be exactly `Option<Self::Wire>`, but these should be semantic changes only, no effect on performance or logic correctness. 

This is just an alternative option, feel free to close it if you disagree.

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -28,6 +28,7 @@ mod trivial;
 mod tests;
 
 use ff::Field;
+use ragu_arithmetic::Coeff;
 use ragu_core::{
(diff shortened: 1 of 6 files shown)


TalDerei:
> Feel free to close this one if you find your PR subsuming this one. @ebfull

yeah seems like https://github.com/tachyon-zcash/ragu/pull/483 already subsumes this

PR #491, opened 2026-02-18
structural hash as registry digest

alxiong:
As mentioned in https://github.com/tachyon-zcash/ragu/issues/78#issuecomment-3885588746:

> Concretely, we need to increase the current iteration from 6 to 4n+1, so that the vanishing polynomial of degree 4n-1 doesn't exist.
Given the computation load, I think it's reasonable to consider changing to key = blake2b(m(X)) by directly hashing the coefficients of m(X) instead of using evaluations to bind it.

This PR attempts to introduce a `Hasher` driver under `s/hash.rs` to stream through the circuit logic and hash the structure in a binding digest as the registry key.

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -200,6 +200,12 @@ pub trait CircuitObject<F: Field, R: Rank>: Send + Sync {
     ///
     /// This captures wire identities and scalar coefficients without evaluating
     /// the polynomial.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the underlying synthesis exceeds the multiplication or linear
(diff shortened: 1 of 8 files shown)


ebfull:
I mentioned elsewhere that we'd prefer to use evaluations to bind to the polynomial because 1) it's a forcing function to make the evaluations efficient, and 2) it doesn't invite the possibility for ambiguous interpretation of its encodings, which needs to be defended in this PR via all these little discriminants and stuff that are hashed in as well.

We could pivot back to a polynomial encoding later.

TalDerei:
this can probably be closed at this point since we've settled on optimizing polynomial evaluations for now.

Issue #240, opened 2025-12-14
Implement custom serialization for `UnifiedInstance`


TalDerei:
Currently, `UnifiedInstance` uses derived serialization, encoding fields in _struct_ definition order (that itself follows the protocol ordering). However, the optimal serialization order _may_ need to differ to support different performance optimization vectors. This would be the first non-derived `Write` / `Read` impl in the codebase.

Synergetic with https://github.com/tachyon-zcash/ragu/issues/191

Issue #191, opened 2025-12-06
Cached / expanded proof structures and serialization API


TalDerei:
Currently, the proof structure stores all fields directly in the proof object without distinguishing between verifier-derivable vs prover-provided fields:
1. Fields that **must** be provided by the prover (lying would cause verification to fail), 
2. Fields that **must** be derived by the verifier (accepting prover-supplied values would introduce manipulation vectors),
3. Fields that are **cached** purely for efficiency

Importantly, we should document where each field comes from. 

We should distinguish between **compressed** and **expanded** proof representations: the compressed form contains only the _minimal_ necessary data required by the verifier (essential polynomials, non-derivable commitments, etc.), and the expanded form serves as the internal working representation and includes all the verifier-derivable fields (derived ones like all pre-computed nested commitments, all challenges, etc.) for efficient caching. 

The motivation here is that many fields in the incremental proof structure we’re assembling can actually be derived from other fields by the verifier:
- Nested commitments derived from polynomials,
- Challenges can be derived from hashes of nested commitments,
- Several other challenges and commitments are verifier-derivable

This has the benefit for clear API boundaries that's explicit about what's required versus derived, smaller proof sizes because compressed proofs omit derivable fields, and prover efficiency. 

**(This is premature until the accumulation impl stabilizes, but filing for visibility)** 

-----------

- [ ] Audit all proof fields,
- [ ] Design `CompressedProof` struct (minimal verifier input)
- [ ] Design `ExpandedProof` struct (prover's internal working representation)
- [ ] Implement `expand(): CompressedProof -> ExpandedProof`
- [ ] Implement `compress(): ExpandedProof -> CompressedProof`
- [ ] Update verifier API to accept compressed proofs

PR #493, opened 2026-02-19
chore(circuits): Implement custom write impl for `UnifiedOutput`

TalDerei:
Motivated by https://github.com/tachyon-zcash/ragu/issues/191 and closes https://github.com/tachyon-zcash/ragu/issues/240. 

Implements the first custom `Write` serialization impl for `UnifiedOutput` to selectively omit fields (replacing the existing derived impl where serialization always writes every struct field in order). From the 21 internal fields that map to 29 wires, 11 were challenges (deterministic Poseidon squeezes from absorbed nested commitments), and can be removed as public inputs because the verifier can rederive these challenges by replaying the FS transcript. Consequently, this cuts it down to 18 wires: lower `k(Y)` degree and slightly fewer constraints across internal recursion circuits.

### DIFF
--- a/crates/ragu_pcd/src/circuits/native/unified.rs
+++ b/crates/ragu_pcd/src/circuits/native/unified.rs
@@ -26,7 +26,10 @@ use ragu_core::{
     gadgets::{Bound, Consistent, Gadget, Kind},
     maybe::Maybe,
 };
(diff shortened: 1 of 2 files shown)


TalDerei:
dishonest prover tests would have caught this soundness bug, filed https://github.com/tachyon-zcash/ragu/issues/494.

Issue #179, opened 2025-12-03
Hardcode rank R


TalDerei:
Remove rank parameterization R and fix it as a generic parameter with a fixed constant (R = 13). I don't think there's any practical use case for varying the rank?

PR #495, opened 2026-02-19
chore(circuits): rank handling with named aliases

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/179. Replaces arbitrary `R<N>` with named rank aliases: `ProductionRank = R<13>` and `TestRank = R<7>` (paring down `impl_rank_for_R`). We can't use `ProductionRank` everywhere because some of our tests would take a few minutes to run in debug mode.

### DIFF
--- a/crates/ragu_circuits/src/polynomials/mod.rs
+++ b/crates/ragu_circuits/src/polynomials/mod.rs
@@ -17,8 +17,9 @@ mod private {
 pub trait Rank:
     private::Sealed + Clone + Send + Sync + 'static + PartialEq + Eq + core::fmt::Debug + Default
 {
-    /// Ragu currently only supports ranks between $2$ and $28$ to avoid
-    /// overflows on 32-bit architectures.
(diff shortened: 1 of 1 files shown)


crates/ragu_circuits/src/polynomials/mod.rs
```diff
@@ -100,11 +101,24 @@ pub trait Rank:
     }
 }
 
-/// Explicit implementations for various basis sizes supported for use in Ragu.
-/// `R<N>` implements [`Rank`] for $N \in [2, 28]$.
+/// `R<N>` implements [`Rank`] for supported values of $N$. The [`Rank`] trait
+/// provides a [`ProductionRank`] ($N = 13$) and [`TestRank`] ($N = 7$). Additional
+/// implementations can be added to `impl_rank_for_R!` as needed.
```
ebfull: nit: The `Rank` trait itself doesn't provide these, we reference the specific type aliases here though.

Issue #256, opened 2025-12-17
Unify ky definitions


TalDerei:
PR #188 introduced a streaming `ky::emulate` function that evaluates `k(Y)` at a point using Horner's method, avoiding intermediate coefficient collection. Both approaches use the same `Write` impl on the circuit's instance output, which is the unified source of truth for k(Y) structure.

Two approaches exist in `ragu_circuits/src/ky.rs`:
   - `eval`: Collects k(Y) coefficients into a `Vec<F>`
   - `emulate`: Streaming Horner evaluation at a point `y`

We have some call sites that invoke `circuit.ky(instance)` + `arithmetic::eval(&ky, y)`, but these can be replaced with a single `ky::emulate(&circuit, instance, y)` to unify these definitions. Optionally, we may consider replacing the `CircuitExt::ky()` trait method.

PR #497, opened 2026-02-20
chore(circuits): unify ky evaluation using a shared/generic horner impl

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/256.

Unifies multiple `ky` definitions by removing the nearly duplicate `ky` wrapper and `Horner` impls from the `ragu_pcd` crate, and relocating `Horner` to `ragu_circuits` as a generic utility. I think this migration is suitable because if we keep the implementation in `ragu_pcd`, then crates will have two different implementations of the same logic (evaluating `ky`), which was the status quo. With this in place:
- the older `CircuitExt::ky()` API can change from coefficient vector to streaming evaluation (without needing to collect into an intermediate `Vec<F>`)
- two-step call site consolidation `(staged.ky(()) + ragu_arithmetic::eval(&ky, y)` collapses to `staged.ky((), y))`


TalDerei:
> registry_ops

this is a good example of noise from the CI runner environment

PR #498, opened 2026-02-20
test: add dishonest prover test suite

LucidSamuel:
## Summary

- Ports the dishonest-prover verifier suite onto the current flattened `Proof` representation.
- Replaces the old `trivial_proof()`-based tests with deterministic seeded real proofs, giving the verifier an honest proof before each single-field corruption.
- Adds a positive acceptance test and a full Fiat-Shamir transcript replay test covering all 11 challenges.
- Adds 36 focused rejection tests covering challenges, bridge commitments, native and nested polynomials, circuit IDs, and header lengths.

## Scope

This is a tests-only change. It does not add `ab_commitment_claim` or `f_commitment_claim` binding checks, or modify any other verifier production logic. The branch was rebuilt on current `main` after the proof representation was flattened.

The port is not byte-identical to the previously reviewed diff because current APIs changed import paths and local naming, but the suite remains semantically the same.

## Validation

- `RUSTUP_TOOLCHAIN=nightly-2026-04-11 cargo fmt --all -- --config-path rustfmt.nightly.toml --check`
- `cargo clippy --workspace --lib --tests --benches --locked -- -D warnings`
- `cargo test -p ragu_pcd --locked`

### DIFF
--- a/crates/ragu_pcd/src/verify.rs
+++ b/crates/ragu_pcd/src/verify.rs
@@ -238,15 +238,19 @@ mod nested {
 #[cfg(test)]
 mod tests {
     use ragu_arithmetic::{
+        FixedGenerators,
         ff::Field,
         rand::{SeedableRng, rngs::StdRng},
     };
     use ragu_circuits::{polynomials::ProductionRank, registry::CircuitIndex};
     use ragu_pasta::Pasta;
+    use ragu_primitives::{GadgetExt, Point};
 
     use super::*;
-    use crate::ApplicationBuilder;
+    use crate::{ApplicationBuilder, step::internal::trivial::Trivial};
 
+    type CF = <Pasta as Cycle>::CircuitField;
+    type NestedCurve = <Pasta as Cycle>::NestedCurve;
     type TestR = ProductionRank;
     const HEADER_SIZE: usize = 4;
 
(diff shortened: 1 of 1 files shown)


crates/ragu_pcd/src/verify.rs
```diff
@@ -266,30 +266,68 @@ mod nested {
...
+            .seed(&mut rng, Trivial::new(), ())
+            .expect("seed should not fail");
+        proof
+    }
+
+    /// Macro to reduce boilerplate for trivial-proof corruption tests.
+    ///
+    /// Creates a trivial proof, applies a corruption, then asserts that
+    /// verification rejects the corrupted proof.
+    macro_rules! assert_trivial_rejects {
```
TalDerei: remove; we shouldn't be testing anything here against the trivial proof, only seeded trivial proof.

crates/ragu_pcd/src/verify.rs
```diff
@@ -368,14 +394,246 @@ mod tests {
...
+    assert_trivial_rejects!(trivial_rejects_corrupted_compute_v_blind, |proof| {
+        proof.circuits.compute_v_blind = CF::from(999u64)
+    });
+
+    // -----------------------------------------------------------------------
+    // Seeded proof tests (7 tests)
+    //
+    // These use a real proof produced by seed() that actually passes
+    // verification, providing meaningful coverage of every verification path.
+    // -----------------------------------------------------------------------
```
TalDerei: let's be more rigorous when creating these tests; there's a lot of duplication in these tests that can be extracted out. moreover, this is a relatively _sparse_ set: covers `x, y, u` (3 of 11 challenges) and `p.blind, p.v, ab.c` (scalars). we should extend this to:
- `w, z, mu, nu, mu_prime, nu_prime, alpha, pre_beta` challenges
- have more useful tests around `preamble, s_prime, error_n, error_m, ab, query, f, eva`l nested commitments
- polynomial coefficient corruptions as noted in the original issue which isn't tracked here, specifically  one per `rx` component, plus `p.poly, f.poly, query.registry_xy_pol`, from `Source::rx()` at `verify.rs`.

TalDerei (changes requested):

LucidSamuel:
Updated per review

nb: while extending coverage, two cases (ab.b_poly, f.poly) initially still verified. so fixed by adding explicit verifier
commitment-binding checks for:
- ab.a_poly / ab.a_blind / ab.a_commitment
- ab.b_poly / ab.b_blind / ab.b_commitment
- f.poly / f.blind / f.commitment

LucidSamuel:
Rebuilt this branch on current `main` and force-pushed it as a single tests-only commit (`95f87942`). The production verifier changes from the old version remain dropped.

One caveat when comparing this with the previously reviewed diff: it is not byte-identical because of cosmetic API drift on current main. `rand` is now re-exported through `ragu_arithmetic`, polynomial paths use the shorter `sparse::Polynomial::new()` form, and `Emulator` comes from the current `verify.rs` scope. The suite and its semantics are otherwise unchanged.

Local validation is clean: pinned nightly rustfmt, workspace clippy with `-D warnings`, and the full `ragu_pcd` test suite.

TalDerei:
taking over this PR, rebasing to resolve merge conflicts and merging.

TalDerei:
oh totally forgot that we haven't fully wired in all the bridge commitments in (reference https://github.com/tachyon-zcash/ragu/pull/498/changes/627755c887f661896ade16229a42ecb4982cb086#diff-be95f5b912639cbd1263c8eb036f1c7f4a77a5dd7aaa2752784a77e54a9fbf4aR525-R532), so i'd like to defer this kind of testing until recursion is finished. cc @ebfull (@LucidSamuel thanks for opening this PR and this is great, so when we're at the stage for this kind of testing we can resurrect this PR, but opting to close for now to reduce the PR queue pressure).

PR #499, opened 2026-02-20
Add cross-circuit coverage protection on set slots

alxiong:
The abstraction of `Slot` provides one-time allocation (via `get()` or lazily via `take()`) and allows fields in `unified::Output` to be set by one circuit and get by another. 

But an invariant it didn't enforce is that internal circuits had indeed `set` to all fields in the Output (thus enforce in circuit as the output of other procedures) rather than blindly allocated to. 
A thought experiment: if we never call any `set()` in any internal circuits, the failure and error message won't be obvious.

This PR 
- introduce a `Coverage` bit-mask to explicitly track the list of fields (challenges + `v`) that supposed to be enforced (thus `set()`) by whichever circuits. 
- Piggyback on `Aux` to deliver this coverage info crossing circuit boundary and validate at `_11_circuit.rs` level
- add drift test to make sure the manual `fn coverage()` didn't miss any fields as `Output` struct changes. 
  - privately I tried to introduce `#[challenge]` on `defined_unified_output!` macro, but ends up way too messy. thus this manual, little verbose `coverage()` is the minimal changes i can come up with

### DIFF
--- a/crates/ragu_pcd/src/circuits/native/unified.rs
+++ b/crates/ragu_pcd/src/circuits/native/unified.rs
@@ -170,6 +170,17 @@ macro_rules! define_unified_instance {
             pub coverage: Coverage,
         }
 
+        impl<C: Cycle> Instance<C> {
+            /// Asserts that every Element slot has been covered by some circuit.
+            ///
+            /// # Panics
+            ///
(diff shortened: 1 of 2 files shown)


TalDerei (approved):
nice work 👍

PR #503, opened 2026-02-21
feat(circuits): floor planning infrastructure for routine memoization

TalDerei:
Depends on https://github.com/tachyon-zcash/ragu/pull/511 and https://github.com/tachyon-zcash/ragu/pull/557. Rebased #413's floor planning and memoization design onto main, replacing the stale `RoutineId`/`RoutineRegistry` system with PR #511's `RoutineFingerprint`. Considerations in https://github.com/tachyon-zcash/ragu/issues/598 should be kept in mind here during review.

---

Memoization is treated purely as an optimization and [shouldn't change](https://github.com/tachyon-zcash/ragu/issues/67) the underlying circuit polynomials.

The thing we're tackling here is when evaluating `s(x,y)` for multiple circuits in the registry, identical routines in the same positions (think of a region that occupies space in the X and Y dimensions), each circuit recomputes the same polynomial contribution. This work caches routine contributions, and when detecting cache hits, skips routine execution entirely and reconstructs the output from cached wires. The different phases of the implementation are documented below.

---

**1. Routine Identity and Shape Tracking**

Lifted from #511. The stale `RoutineId` (type-name based), `RoutineInfo`, `RoutineRegistry`, and `nesting_level` from #413 are deleted entirely, addressing [review feedback](https://github.com/tachyon-zcash/ragu/pull/503#pullrequestreview-3857761660).

---

**2. Floor Planner Implementation**

References https://github.com/tachyon-zcash/ragu/issues/59. Implements the red section in the diagram.

The primary function of the floor planner is to align the same routine types across circuits (**inter-circuit**) and minimize fragmentation in the registry. The implementation is a simple greedy placement algorithm, `from_segment_records` (as opposed to more complex bin-packing) which extracts `RoutineFingerprint` from each circuit's segment records, groups by fingerprint, sorts by `circuit_count * max_invocations * area` (terming as "benefit"), and assigns the same routine types to canonical positions across all circuits based on that. It effectively produces a placement map with X, Y offsets.

- `Placement`: One entry mapping a routine fingerprint to its `ConstraintSegment` (position + dimensions)
- `RegistryPosition`: The (x, y) coordinates in the registry

---

**3. Integrate Floor Plan with Registry**

References https://github.com/tachyon-zcash/ragu/issues/60. Implements the orange section in the diagram.

`RegistryBuilder::finalize()` builds the type-based floor plan directly from `circuit.segment_records()`. The registry uses these positions when placing routines, enabling memoization both inter-circuit (same routines will have the same canonical positions and we can combine them via Lagrange coefficients) and intra-circuit (scaling by X and Y offsets).

---

**4. Cache Infrastructure**

Provides scaffolding for green section in the diagram.

Adds memoization types (`MemoCache`, `CachedRoutine`, `WireExtractor`, `WireInjector`) and `sxy_with_cache` to `CircuitObject` trait. `MemoCache` is keyed on `(RoutineFingerprint, RegistryPosition)`. `WireExtractor`/`WireInjector` use the `WireMap` trait. The `sxy::Evaluator` carries a `segment_records` slice and extracts fingerprints at each routine index during synthesis.

---

**5. Exercise Memoization and Benchmarks/Tests**

(**Partially**) References https://github.com/tachyon-zcash/ragu/issues/58, https://github.com/tachyon-zcash/ragu/issues/389.

Implements `wxy_combined`, an optimized registry evaluation that shares memoization cache across circuits during Lagrange interpolation. This is **inter-circuit** routine memoization (cache polynomial evaluations shared across circuits), not intra-circuit (multiple invocations within one circuit that can be cached by scaling, which isn't implemented _yet_ and will be considered follow-up work).

On cache hits, routine execution is skipped entirely via `WireExtractor`/`WireInjector`, reconstructing the output gadget from cached wires.

---

_Not yet implemented (follow-up work):_
- Intra-circuit memoization (scaling by offsets for repeated routine invocations within one circuit)
- Static/dynamic input analysis for determining which polynomial parts are cacheable
- Memoization for `sx` and `sy` restrictions (only `sxy` currently)
- Nested routine memoization (currently only top-level routines memoize)
- Wiring memoization into production call sites (currently only exercised in tests/benches)

---

Screenshot 2026-01-30 at 2 48 32 PM


crates/ragu_circuits/benches/circuits.rs
```diff
@@ -155,9 +155,23 @@ fn wxy((registry, (w, x, y)): (Registry<'_, Fp, TestRank>, (Fp, Fp, Fp))) {
...
+    black_box(registry.wxy(w, x, y));
+}
+
+#[library_benchmark(setup = setup_with_rng)]
+#[bench::wxy_combined_heavy_routines(registry_heavy_routines(), (f, f, f))]
+fn wxy_combined_heavy_routines(
+    (registry, (w, x, y)): (Registry<'_, Fp, ProductionRank>, (Fp, Fp, Fp)),
+) {
+    black_box(registry.wxy_combined(w, x, y));
+}
```
TalDerei: references https://github.com/tachyon-zcash/ragu/pull/413#discussion_r2807263618; importantly, we only currently exercise memoization in tests/benches and compare against production paths to make sure it's correct. follow-up work will be to integrate it fully by replacing call sites with memoization variants.

crates/ragu_circuits/src/routines.rs
```diff
@@ -0,0 +1,165 @@
...
+impl RoutineId {
+    /// Creates a `RoutineId` for the given routine type.
+    pub fn of<R>() -> Self {
+        Self(core::any::type_name::<R>())
+    }
+}
+
+/// Describes the circuit resources consumed by a routine.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct RoutineShape {
```
alxiong: can we reuse `SegmentRecord`? maybe type alias?

crates/ragu_circuits/src/routines.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+impl RoutineRegistry {
+    /// Creates a new, empty routine registry.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Registers a routine invocation with the given shape.
+    pub fn register<R: 'static>(&mut self, shape: RoutineShape) {
```
alxiong: i don't see any usage of it, (nor `enter_routine()`). Are we planning to use them in future PR, roughly where?

alxiong (commented):
Just a half-first pass. 

meta-question: what's the intended relationship between this `floor_plan.rs` and logic (and structs) in `floor_planner.rs`?

crates/ragu_circuits/src/routines.rs
```diff
@@ -0,0 +1,165 @@
...
+/// The `RoutineRegistry` collects information about all routine invocations
+/// during a discovery pass over circuits. This information is used for:
+///
+/// - **Structural analysis**: Identifying which circuits use the same routines
+/// - **Floor planning**: Determining optimal routine placement within the registry
+/// - **Memoization**: Enabling reuse of polynomial evaluations for identical routines
+#[derive(Debug, Clone, Default)]
+pub struct RoutineRegistry {
+    entries: BTreeMap<RoutineId, Vec<RoutineInfo>>,
+    current_nesting_level: usize,
```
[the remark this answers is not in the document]
TalDerei: nested routines are ignored for memoization purposes (currently only top-level routines memoize).

crates/ragu_circuits/src/floor_plan.rs
```diff
@@ -0,0 +1,255 @@
...
+/// All circuits using a routine type should place it at the canonical position
+/// to maximize inter-circuit memoization.
+#[derive(Debug, Default)]
+pub struct FloorPlan {
+    placements: Vec<Placement>,
+    next_x: usize,
+    next_y: usize,
+    row_height: usize,
+    max_width: usize,
+}
```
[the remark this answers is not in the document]
TalDerei: > * I find it little hard to follow the variable/field name, regarding `row` and `width`, what's the orientation of the matrix

It's just a 2D region with x and y coordinates. as we discussed OOB, this is all a greedy floor planner algorithm really only needs to know for performing some kind of (ideally) bin packing optimization. we do something simpler than bin packing for now, but we could follow-up on improving it. Halo2 doesn't even do bin packing, they just have a simpler greedy placement algorithm iirc.

ebfull (changes requested):
High level pass:

* I think routine invocation identity and comparison (#62) needs to be sorted out in a separate PR. It can be figured out independently of these changes, and it really matters that it's done correctly or we could develop the wrong intuitions in the rest of the floor planning design.
* This PR's routine identity concept conflicts with my [intuition of how it should work.](https://github.com/tachyon-zcash/ragu/issues/62#issuecomment-3963126884) The typename approach doesn't work even as an approximation. There's an algebraic understanding of identity we need to capture first before we know what flexibility the floor planner actually has.
* The registry _for routines_ seems unnecessary. We already go through the circuits during metrics; we should just enhance that driver / pass to collect fingerprints of routine (invocations) according to whatever identity system we adopt, and that information should be stored in the routine records. Perhaps I'm misunderstanding the implementation.
* I have a bad feeling whenever I see that we're tracking things like the nesting level. It seems like a crutch for the identity/comparison system not being sound.

TalDerei:
personal reminder: convert existing test coverage here using proptests.

TalDerei:
converting to draft since this depends on https://github.com/tachyon-zcash/ragu/pull/557.

PR #400, opened 2026-01-29
feat(pcd): add `endoscale_challenges` circuit for challenge smuggling

TalDerei:
Supersedes https://github.com/tachyon-zcash/ragu/pull/356 and references https://github.com/tachyon-zcash/ragu/issues/71. Extracts y / z / etc. challenges from the unified instance and produces field-scaled coefficients (calls `Endoscalar::extract_endoscalar()` and `Endoscalar::lift_endoscalar()` on each challenge) placed in a-positions of the trace polynomial (essentially models a staging polynomial inside the circuit). Challenges need to end up in a-positions of the trace polynomial so they can be endoscaled on the nested curve side.

Intended to build on the anticipated dual-pipeline PR, from which some supplemental context is currently missing. Once the wiring is in place, we'll still require nested circuit verification of the smuggled challenges.

**Update:** extracted stage mask API independently into https://github.com/tachyon-zcash/ragu/pull/504, so this will need to be rebased. 

Additionally, added nested-side logic for stage mask checks:                                                                                                                                                                                              
  - `SmuggledChallengesStage` and `SmuggledChallengesWitness`                                                                                                                                                                                            
  - `smuggled_challenges_rx` generation in fuse                                                                                                                                                                                                        
  - `StageMask` registration in nested registry                                                                                                                                                                                                       
  - `rx.revdot(sy) == 0` verification in nested claims

PR #504, opened 2026-02-21
feat(circuits): `StageExt` generator API for challenge smuggling

TalDerei:
Extracted from https://github.com/tachyon-zcash/ragu/pull/400. 

`StageExt` API method for computing generator indices and points for A-coefficient positions, enabling challenge smuggling.
  - `generator_index_for_a()`: returns generator index for i-th A coefficient                                                                                                                                                                          
  - `generator_for_a_coefficient()`: internal `StageMask` helper for tests that returns the actual generator point

### DIFF
--- a/crates/ragu_circuits/src/staging/mask.rs
+++ b/crates/ragu_circuits/src/staging/mask.rs
@@ -54,27 +54,6 @@ impl<R: Rank> StageMask<R> {
             _marker: core::marker::PhantomData,
         })
     }
-
-    /// Returns the generator point for the i-th A coefficient of this stage.
-    ///
-    /// This is useful for computing commitments to values placed in A positions
-    /// of the witness polynomial, such as challenge coefficients for smuggling.
-    #[cfg(test)]
-    pub fn generator_for_a_coefficient<C: ragu_arithmetic::CurveAffine>(
-        &self,
-        generators: &impl ragu_arithmetic::FixedGenerators<C>,
-        coefficient_index: usize,
-    ) -> C {
-        assert!(
-            coefficient_index < self.num_multiplications,
-            "coefficient_index {} exceeds num_multiplications {}",
-            coefficient_index,
-            self.num_multiplications
-        );
-
-        let idx = 2 * R::n() + 1 + self.skip_multiplications + coefficient_index;
-        generators.g()[idx]
-    }
 }
 
 impl<F: Field, R: Rank> CircuitObject<F, R> for StageMask<R> {
@@ -262,7 +241,7 @@ mod tests {
     use ff::Field;
     use group::{Curve, prime::PrimeCurveAffine};
     use proptest::prelude::*;
-    use ragu_arithmetic::{Coeff, Cycle, FixedGenerators, Uendo};
+    use ragu_arithmetic::{Coeff, CurveAffine, Cycle, FixedGenerators, Uendo};
     use ragu_core::{
         Result,
         drivers::{Driver, DriverValue, LinearExpression, emulator::Emulator},
@@ -331,6 +310,28 @@ mod tests {
             }
 
             Ok(((), D::just(|| ())))
+        }
+    }
+
+    impl<R: Rank> StageMask<R> {
+        /// Returns the generator point for the i-th A coefficient of this stage.
+        ///
+        /// This is useful for computing commitments to values placed in A positions
+        /// of the witness polynomial, such as challenge coefficients for smuggling.
+        fn generator_for_a_coefficient<C: CurveAffine>(
+            &self,
+            generators: &impl FixedGenerators<C>,
+            coefficient_index: usize,
+        ) -> C {
+            assert!(
+                coefficient_index < self.num_multiplications,
+                "coefficient_index {} exceeds num_multiplications {}",
+                coefficient_index,
+                self.num_multiplications
+            );
+
+            let idx = 2 * R::n() + 1 + self.skip_multiplications + coefficient_index;
+            generators.g()[idx]
         }
     }
 


crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -757,4 +804,156 @@ mod tests {
...
+        for i in 0..4 {
+            let gen_idx = stage.generator_index_for_a(i);
+            let expected_gen = generators.g()[gen_idx];
+            let actual_gen = stage.generator_for_a_coefficient(generators, i);
+            assert_eq!(actual_gen, expected_gen);
+        }
+    }
+
+    #[test]
+    fn test_a_only_commitment_for_challenge_smuggling() {
```
TalDerei: this creates a stage, places values only in a-coefficients, commits to the stage with `rx.commit(generators, blind)`, then computes summation of `challenge[i] * generator_for_a_coefficient(i)`, and checks against the commitment and assert them match. Uses a zero-blinding factor just for testing.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -54,6 +55,37 @@ impl<R: Rank> StageMask<R> {
             _marker: core::marker::PhantomData,
         })
     }
+
+    /// Returns the generator index for the i-th A coefficient of this stage.
+    ///
+    /// The A coefficients are placed at positions `2n + 1 + skip + i` in the
```
TalDerei: `2 * R::n() + 1 + skip + i` was the key to getting this right

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -54,6 +55,37 @@ impl<R: Rank> StageMask<R> {
...
+        );
+
+        2 * R::n() + 1 + self.skip_multiplications + coefficient_index
+    }
+
+    /// Returns the generator point for the i-th A coefficient of this stage.
+    ///
+    /// This is useful for computing commitments to values placed in A positions
+    /// of the witness polynomial, such as challenge coefficients for smuggling.
+    #[allow(dead_code)]
```
ebfull: Why is this `dead_code`?

TalDerei: to suppress clippy warnings in CI, currently only used in tests

ebfull (changes requested):
Ah. I see what's going on. I think these should not be methods on `StageMask`, which isn't intended to be exported outside of `ragu_circuits`. This should be a default method on `StageExt`. You wouldn't have any dead_code issues and the API would be consistent with the other methods.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -945,27 +824,140 @@ mod tests {
...
+        let rx: structured::Polynomial<Fp, R> = ParentAOnlyStage::rx(challenges).unwrap();
+        let poly_commitment: EqAffine = rx.commit(generators, blind);
 
-        // Manually compute expected commitment using generator_for_a_coefficient.
+        // Manually compute expected commitment using StageExt::generator_for_a_coefficient.
         let mut manual_commitment = EqAffine::identity();
         for (i, &challenge) in challenges.iter().enumerate() {
-            let a_gen = stage_obj.generator_for_a_coefficient(generators, i);
+            let a_gen =
+                <ParentAOnlyStage as StageExt<Fp, R>>::generator_for_a_coefficient(generators, i);
```
TalDerei: modified the tests to use the `StageExt` structure.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -793,101 +761,12 @@ mod tests {
...
-        );
-    }
-
     /// A stage that allocates values only in a-positions (b = 0) for challenge smuggling.
     ///
     /// Each value is paired with a zero to ensure it lands in an a-coefficient position
     /// when the polynomial is built. This mimics the pattern used for smuggling challenges.
     #[derive(Default)]
-    struct AOnlyStage;
+    struct ParentAOnlyStage;
```
TalDerei: `ParentAOnlyStage` and `ChildOfParentAOnlyStage` are test stages for exercising the generator index logic for `generator_index_for_a`.

crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -400,6 +401,34 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
...
+        2 * R::n() + 1 + Self::skip_multiplications() + coefficient_index
+    }
+
+    /// Returns the generator point for the i-th A coefficient of this stage.
+    ///
+    /// This is useful for computing commitments to values placed in A positions
+    /// of the witness polynomial, such as challenge coefficients for smuggling.
+    fn generator_for_a_coefficient<C: CurveAffine>(
+        generators: &impl FixedGenerators<C>,
+        coefficient_index: usize,
```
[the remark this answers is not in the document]
TalDerei: `generator_index_for_a` already embeds a necessary bounds check to make sure we're not accessing outside of `R::n()` (generators for reserved gates). technically verification should catch this downstream, but this is a defense-in-depth check rather than trusting the caller to propagate.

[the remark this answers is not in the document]
ebfull: But `generator_for_a_coefficient` is being exposed publicly in the trait.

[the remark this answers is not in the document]
TalDerei: sorry what's the functional issue? `generator_for_a_coefficient` internally calls `generator_index_for_a`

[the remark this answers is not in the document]
TalDerei: wanted to expose both for maximal flexibility because I wasn't sure in what form these convenience methods were going to be used _exactly_ yet by callers, but reverting to a simpler API

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -54,6 +55,27 @@ impl<R: Rank> StageMask<R> {
             _marker: core::marker::PhantomData,
         })
     }
+
+    /// Returns the generator point for the i-th A coefficient of this stage.
+    ///
+    /// This is useful for computing commitments to values placed in A positions
+    /// of the witness polynomial, such as challenge coefficients for smuggling.
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
ebfull: Use `#[cfg(test)]` because those tests are unit tests anyway.

Issue #21, opened 2025-10-08
Flame-graph profiling for R1CS-NARK and split-accumulation performance


TalDerei:
The objective is to visually analyze the relative cost of and hot paths for cryptographic operations, constraint counts, etc. to understand the performance characteristics of the implementation across layers.

PR #507, opened 2026-02-24
chore(bench): flamegraph profiling via cargo-flamegraph and perf

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/21. Instead of the approach in https://github.com/tachyon-zcash/ragu/pull/506 (which tries to structurally fix gungraun flamegraphs and doesn't currently work), this implements flamegraph support for profiling individual gungraun benchmarks. It runs the gungraun bench binary directly under [perf](https://www.brendangregg.com/perf.html) (Linux kernel profiler) via [cargo flamegraph](https://crates.io/crates/flamegraph), bypassing gungraun-runner's valgrind orchestration, so any benchmark listed in a `library_benchmark_group!` can be targeted by name. Integrating into CI will [exceed memory limits](https://github.com/tachyon-zcash/ragu/issues/21#issuecomment-3929213537), so this profiling can be invoked a as discretionary local workflow for dev purposes.

---------

**Usage:** `just flamegraph <package> <target>` (eg. `just flamegraph ragu_pcd fuse` and `open target/flamegraph-profile_fuse.svg`) 

Screenshot 2026-02-23 at 8 23 18 PM

### DIFF
--- a/justfile
+++ b/justfile
@@ -67,7 +67,7 @@ _bench_macos *ARGS:
     docker attach --no-stdin $container
 
 _bench_linux *ARGS: _gungraun_setup
-    cargo bench --workspace --all-features {{ARGS}}
+    cargo bench --workspace --all-features --bench arithmetic --bench circuits --bench pcd --bench primitives {{ARGS}}
 
 # generate flamegraph in target/*.svg
 flamegraph PACKAGE GROUP TARGET *ARGS:


alxiong (approved):
ACK.  I can produce the same flamegraph locally (after my changes)

I pushed a few changes:
1. cleaner `just flamegraph_macos` and maximize docker layer reuse (with what's already pulled in `just bench`) by following the `nixery.dev` approach:
   - Uses `nixery.dev/{{_nixery_meta}}/cargo-binstall/gcc/just/rustup/perf` instead of `rust:1.90-bookworm + apt`
  - Mounts `ragu-cargo` and `ragu-rustup` named volumes for caching
  - Delegates to `just flamegraph-linux` 
2. hide all platform-specific just recipes, only `just bench` and `just flamegraph`
3. explicit `GROUP` and `TARGET` argument, so that the just logic is simpler, only parse `func_idx` -- this is reasonable because when i need to know which bench to flamegraph, i will locate the bench file and see: `library_benchmark_group!(name =app_proof; benchmarks = seed, fuse);` and know i will need to `just flamegraph app_proof fuse`
4. a handy alias `just flamegraph_fuse`

PR #510, opened 2026-02-24
feat(arithmetic): maybe_rayon and criterion support 

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/18 and minimally resurrects https://github.com/tachyon-zcash/ragu/pull/46. Synergetic with https://github.com/tachyon-zcash/ragu/pull/507.

`rayon` dependency (data-parallelism library) wasn't currently employed, which leaves a clean slate for implementing a parallelization strategy. Ragu will remain `no_std` by default, which is common for cryptographic libraries and will support wider deployment targets. Multi-core support is gated behind an optional `[multicore]` feature. When enabled, `rayon` becomes available and `std` is pulled in as a required dependency.

Now that we have a bench suite, arguably the most obvious optimization to match Halo2's [best_multiexp](https://github.com/zcash/halo2/blob/main/halo2_proofs/src/arithmetic.rs#L138-L143) in the _interim_ is: parallelizing the MSM window loop, where each window's bucket sum is computed on its own rayon thread. This implements a single rayon optimization.

However, gungraun uses Valgind to _simulate_ execution on single virtual CPU and counts instructions, and we can't measure multi-threaded speedups because it doesn't capture wall-clock time. For our existing benches, that's fine because wall-clock time doesn't offer the same determinism as instruction counts, but we'll need [Criterion](https://github.com/bheisler/criterion.rs) to explicitly measure performance improvements from rayon, so this adds a `criterion/` subfolder under `benches/` for local, discretionary wall-clock benchmarks (not run in CI).

---------------

Sequential versus parallel MSM benchmarks (multicore saturates at ~6x faster):

Screenshot 2026-02-24 at 7 46 39 AM

Screenshot 2026-02-24 at 7 46 51 AM

### DIFF
--- a/crates/ragu_arithmetic/src/util.rs
+++ b/crates/ragu_arithmetic/src/util.rs
@@ -1,6 +1,7 @@
-use alloc::{boxed::Box, vec, vec::Vec};
 use ff::{Field, PrimeField};
 use pasta_curves::{arithmetic::CurveAffine, group::Group};
+
+use alloc::{boxed::Box, vec, vec::Vec};
 
 use crate::domain::Domain;
 use crate::multicore::*;


crates/ragu_arithmetic/src/util.rs
```diff
@@ -240,13 +240,52 @@ pub fn mul<
...
-            acc += &running_sum;
+            sum += &running_sum;
         }
+        sum
     }
 
-    acc
+    let bases_vec: Vec<C> = bases.into_iter().copied().collect();
+
+    #[cfg(feature = "multicore")]
```
TalDerei: feature gating parallel and sequential paths

crates/ragu_arithmetic/src/util.rs
```diff
@@ -240,13 +241,65 @@ pub fn mul<
...
+
+        acc
     }
 
-    acc
+    #[cfg(not(feature = "multicore"))]
+    {
+        let mut acc = C::Curve::identity();
+
+        for current_segment in (0..segments).rev() {
```
TalDerei: no-op diff; code block is the same as the original sequential impl, just shifted down

Cargo.toml
```diff
@@ -68,6 +71,8 @@ lazy_static = "1.5.0"
 proptest = "1.7.0"
 gungraun = "0.17.0"
 blake2b_simd = "1.0"
+rayon = "1.11.0"
+criterion = { version = "0.5", default-features = false }
```
alxiong: curious, why disabling default features?

alxiong (commented):
ACK. 

pushed one commit ([ad0f03f](https://github.com/tachyon-zcash/ragu/pull/510/commits/ad0f03ff34d221def5f22436f80c1697bc42f216)) for better code reuse and minimize the scope of feature-bifurcated logic.

crates/ragu_arithmetic/src/util.rs
```diff
@@ -250,57 +249,29 @@ pub fn mul<
...
-                }
-            }
-
-            // Summation by parts
-            let mut running_sum = C::Curve::identity();
-            for exp in buckets.into_iter().rev() {
-                running_sum = exp.add(running_sum);
-                acc += &running_sum;
-            }
+    let window_sums = (0..segments)
```
TalDerei: ACK – refactored version uses `window_sum` as a shared helper, and we now gate on how windows are dispatched.

.github/workflows/rust.yml
```diff
@@ -56,36 +56,55 @@ jobs:
...
     needs: changes
     if: needs.changes.outputs.rust == 'true' || github.event_name == 'push'
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - name: sequential
+            cargo_features: ""
+          - name: multicore
+            cargo_features: "--all-features"
```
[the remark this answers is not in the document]
TalDerei: cross-referencing against Halo2, they have a multicore feature like Ragu, but their CI just runs with all features enabled and doesn't separately test the no-rayon path. maybe we should adopt that approach where we test once with `--all-features`, trust that the `#[cfg(not(feature = "multicore"))]` path works if it compiles

.github/workflows/bench.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           find target/gungraun -name 'summary.json' -delete
 
       - name: Run benchmarks
-        run: cargo bench --workspace --all-features --locked -- --save-summary=json
+        run: cargo bench --workspace --all-features --locked --bench arithmetic --bench circuits --bench pcd --bench primitives -- --save-summary=json
```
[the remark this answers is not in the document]
TalDerei: this was actually necessary to prevent CI from failing, reference https://github.com/tachyon-zcash/ragu/actions/runs/22359457092/job/64708420976.

crates/ragu_arithmetic/src/util.rs
```diff
@@ -1,4 +1,7 @@
 use ff::{Field, PrimeField};
+use maybe_rayon::iter::IntoParallelIterator as _;
+#[allow(unused_imports)]
+use maybe_rayon::iter::ParallelIterator as _;
```
alxiong: remove this?

crates/ragu_arithmetic/Cargo.toml
```diff
@@ -30,11 +32,18 @@ ff = { workspace = true }
 pasta_curves = { workspace = true }
 ragu_macros = { path = "../ragu_macros", version = "0.0.0" }
 rand = { workspace = true }
+maybe-rayon = { workspace = true }
```
alxiong: nit: this crate is small-scoped enough that I would just copy over to reduce one less external dep (and thus open to adding more)

another suggestion is just cherry-pick logic from [`p3-maybe-rayon`](https://github.com/Plonky3/Plonky3/tree/main/maybe-rayon/src) over (with credit ofc). They ported more `iter` apis (like `repeat`).

crates/ragu_arithmetic/src/util.rs
```diff
@@ -240,10 +243,29 @@ pub fn mul<
...
+        let mut sum = C::Curve::identity();
         for exp in buckets.into_iter().rev() {
             running_sum = exp.add(running_sum);
-            acc += &running_sum;
+            sum += &running_sum;
+        }
+        sum
+    }
+
+    let bases_vec: Vec<C> = bases.into_iter().copied().collect();
```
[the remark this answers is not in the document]
alxiong: done in [cbfca7b](https://github.com/tachyon-zcash/ragu/pull/510/commits/cbfca7b93f231752f710b6628e1101d14ba67f93)

PR #511, opened 2026-02-26
feat(circuits): routine identity fingerprinting

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/62 and precursor to https://github.com/tachyon-zcash/ragu/pull/503. Cherry picks @ebfull's https://github.com/tachyon-zcash/ragu/commit/12231901ad2295154841caa512b38a960669595e and resolves merge conflicts on current main. 

Instead of folding child fingerprints into the parent (parent's fingerprint basically reflects the entire subtree there), we want a routine's fingerprint to capture only its own _local_ constraints. So instead, outputs of nested routines are auxiliary inputs to a routine, and this is done by remapping the child's output wires as fresh allocations in the parent's scope. This was a simpler change than the discussion context in the closing issue implied. Side-effects of this are segments float independently (parent and child fingerprints are decoupled so floor planner can reorder them based on their internal contributions), and it still piggybacks on the single DFS driver pass to avoid quadratic blowup in routine synthesis.

### DIFF
--- a/crates/ragu_circuits/src/tests/identity.rs
+++ b/crates/ragu_circuits/src/tests/identity.rs
@@ -2,7 +2,7 @@ use ff::Field;
 use ragu_core::maybe::Always;
 use ragu_core::{
     Result,
-    drivers::{Driver, DriverValue},
+    drivers::{Driver, DriverValue, LinearExpression},
     gadgets::{Bound, Kind},
     maybe::Maybe,
     routines::{Prediction, Routine},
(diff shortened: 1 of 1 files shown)


crates/ragu_circuits/src/metrics.rs
```diff
@@ -57,17 +135,24 @@ use super::{Circuit, DriverScope};
...
     /// The number of multiplication constraints in this segment.
     pub num_multiplication_constraints: usize,
 
     /// The number of linear constraints in this segment, including constraints
     /// on wires of the input gadget and on wires allocated within the segment.
     pub num_linear_constraints: usize,
+
+    /// The structural identity of this routine invocation.
+    // TODO: consumed by the floor planner (not yet implemented)
+    #[allow(dead_code)]
```
TalDerei: this is currently dead code because there isn't a consumer yet; thinking about general optimizations here, we currently fingerprint every routine. what if `SegmentRecord`'s records `(TypeId_in, TypeId_out),` and lazily fingerprints (stores closures that lazily compute fingerprints)? the benefit is we only need to fingerprint if two routines share a (`TypeId_in, TypeId_out`) pair, rather than fingerprinting on unique pairs. I think this would model some kind of two-phase fingerprinting pass.

moreover, referencing discussion context in https://github.com/tachyon-zcash/ragu/pull/503#discussion_r2855259468, why isn't hash-based fingerprinting on the table? maybe a dedicated `LCHash` LC type that computes a running sum of the internal constraint structure, stores in a `u64`, and avoids field arithmetic.

TalDerei: was experimenting with BLAKE2b hashing in https://github.com/tachyon-zcash/ragu/pull/512 (references last commit https://github.com/tachyon-zcash/ragu/pull/512/changes/be8797e1a8942ec27daa038b14c36368c730f784).

crates/ragu_core/src/drivers/emulator.rs
```diff
@@ -584,6 +584,18 @@ mod tests {
...
+        let acc = DirectSum::<F>::default()
+            .add(&F::from(5))
+            .gain(Coeff::Arbitrary(F::from(2)))
+            .add(&F::from(3))
+            .gain(Coeff::NegativeOne)
+            .add(&F::from(4));
+
+        assert_eq!(acc.value, F::from(3));
+    }
+
```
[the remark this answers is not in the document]
TalDerei: oh good catch, removed. I had copied that test over and modified it for the emulator, then reverted and forgot to delete it.

ebfull (changes requested):
These review comments were produced by Claude.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -1,24 +1,119 @@
...
+//! computes Schwartz–Zippel fingerprints for each routine invocation via the
+//! merged [`Counter`] driver, which combines constraint counting with identity
+//! evaluation in a single DFS traversal.
+//!
+//! # Fingerprinting
+//!
+//! A routine's fingerprint is the tuple `(TypeId(Input), TypeId(Output),
+//! eval)`. The [`TypeId`] pairs cheaply narrow equivalence candidates by type;
+//! the scalar confirms structural equivalence via random evaluation
+//! (Schwartz–Zippel).
```
ebfull: This isn't technically true anymore.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -156,40 +391,187 @@ impl<'dr, F: Field> Driver<'dr> for Counter<F> {
...
-        },
-        num_linear_constraints: 0,
-        num_multiplication_constraints: 0,
-        segments: alloc::vec![SegmentRecord::default()],
-        _marker: PhantomData,
-    };
+/// [`WireMap`] adapter that maps wires from an arbitrary source driver into
+/// [`Counter`] via fresh allocations. Used by [`fingerprint_routine`] where the
+/// source driver is generic.
+#[cfg(test)]
```
alxiong: nit: can we just move them into local definition inside `fingerprint_routine`, so it won't be individually marked as `#[cfg(test)]`

crates/ragu_circuits/src/metrics.rs
```diff
@@ -88,66 +191,199 @@ pub struct CircuitMetrics {
...
+    ///
+    /// Passed to [`WireEvalSum::new`] so that [`WireEval::One`] variants can be
+    /// resolved during linear combination accumulation.
+    one: F,
+
+    /// Initial value of the Horner accumulator for each routine scope.
+    ///
+    /// A nonzero seed derived from the same BLAKE2b PRF ensures that leading
+    /// `enforce_zero` calls with zero-valued linear combinations still shift
+    /// the accumulator (via `result = result * y + 0`), preventing structurally
```
alxiong: typo: 
```suggestion
    /// the accumulator (via `result = h * y + lc_value`), preventing structurally
```

"when they differ only by prepended trivial constraints." lacks clarity.

I'd just give a concrete edge case more explicitly: R1 could have zero-valued LC (though unreasonable, but allowed) where `lc(WireEvalSum).value = 0`

all `enforce_zero` build up: `result = c1 * y^n-1 + c2 * y^n-2 + ... + cn`, 
and if `c1 = 0`, it will degenerate the result and might collide with another routine with `result = c2 * y^n-2 + ... + cn`.

setting the initial result to `h` instead of `ZERO`, gives `result = h^n + c1 * y^n-1 + ...`.

crates/ragu_circuits/src/registry.rs
```diff
@@ -84,13 +84,13 @@ pub struct RegistryBuilder<'params, F: PrimeField, R: Rank> {
     application_steps: Vec<Box<dyn CircuitObject<F, R> + 'params>>,
 }
 
-impl<F: PrimeField, R: Rank> Default for RegistryBuilder<'_, F, R> {
+impl<F: PrimeField + FromUniformBytes<64>, R: Rank> Default for RegistryBuilder<'_, F, R> {
```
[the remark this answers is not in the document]
TalDerei: think you meant `FromUniformBytes` is the subtrait of `PrimeField`, but this indeed seems like it can be shortened.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -156,40 +392,187 @@ impl<'dr, F: Field> Driver<'dr> for Counter<F> {
...
+        //    b-wire or create a new one.
+        //
+        // Effect (1) is kept; effect (2) is rolled back. The asymmetry is
+        // deliberate:
+        //
+        // Sequences must advance because the remapped output wires need
+        // evaluations that are distinct from each other and from any
+        // subsequent parent allocation. Letting the sequences advance past
+        // the remap positions achieves this — each output wire lands at a
+        // unique geometric position in the parent's sequence space.
```
alxiong: while I agree with this, it reads cryptic the first time i read it.

I'd be more ELI5-style: a/b/c wire in child scope are reset to (x0,x1,x2) upon entry, thus the allocated wires for the output gadget is local to that scope. When restoring the parent scope, we need globally-distinct WireEval, thus re-allocation via remapping is needed.

PR #512, opened 2026-02-27
feat(circuits): routine identity fingerprinting using hashing

TalDerei:


crates/ragu_circuits/src/metrics.rs
```diff
@@ -88,66 +165,207 @@ pub struct CircuitMetrics {
...
+        Coeff::Zero => 0,
+        Coeff::One => 1,
+        Coeff::Two => 2,
+        Coeff::NegativeOne => 3,
+        Coeff::Arbitrary(_) => 4,
+        Coeff::NegativeArbitrary(_) => 5,
+    }
+}
+
+/// Returns the low 64 bits of a BLAKE2b finalization.
```
TalDerei: bottleneck shifts from field muls to BLAKE2b compression every 128 bytes – interested to compare performance.

TalDerei:
this was experimental, closing in favor of more established impl in https://github.com/tachyon-zcash/ragu/pull/511, cc @ebfull @alxiong

Issue #192, opened 2025-12-06
Refactor proof fields into `CommittedPolynomial` types


TalDerei:
Instead of storing individual fields like polynomials, blindings, commitments in the proof structure, we should store idiomatic groupings like `CommittedStructuredPolynomial` and `CommittedUnstructuredPolynomial` tuples. 

```ts
pub struct _CommittedPolynomial<P, C: Cycle> {
    pub poly: P,
    pub blind: C::CircuitField,
    pub commitment: C::HostCurve,
}

pub type _CommittedStructured<R, C> =
    CommittedPolynomial<structured::Polynomial<<C as Cycle>::CircuitField, R>, C>;
pub type _CommittedUnstructured<R, C> =
    CommittedPolynomial<unstructured::Polynomial<<C as Cycle>::CircuitField, R>, C>;
```

We should also distinguish between **Native** and **Nested** commitments that are committed in staging polynomials: 
- Native Stage (Over `C::CircuitField`) which contains` C::CircuitField` and `C::ApplicationCurve` elements,
- Nested Stage (Over `C::ScalarField`) which contains `C::ScalarField` and `C::HostCurve` elements

The one concern here is that hiding fields inside these typed structures loses visibility into derivability relationships.

TalDerei:
closed by PR that introduced `RxTriple`.

PR #515, opened 2026-03-01
feat: polynomial smart pointers

### LINKED ISSUE #178: Add smart pointers for polynomials
Wrap polynomials in smart pointers (`Arc` or `Rc` or `Cow`) to enable cheap cloning and commitment caching. Cloning polynomials currently involves full copies. Additionally, if the same polynomial is committed to multiple times (across merge steps on different sides of the curve cycle), the commitment is recomputed each time.

- [x] Wrap polynomial storage in `Arc<T>` or `RC<T>` 
- [x] Cache computed commitments alongside the polynomial
- [x] If a polynomial hasn't changed since the last commitment, return the cached value instead of recomputing the MSM

alxiong:
Close #178, #192

See https://github.com/tachyon-zcash/ragu/issues/178#issuecomment-3973673762 for rationale.

### DIFF
--- a/crates/ragu_circuits/src/polynomials/committed.rs
+++ b/crates/ragu_circuits/src/polynomials/committed.rs
@@ -3,7 +3,79 @@
 //! [`CommittedPolynomial`] bundles a polynomial, its blinding factor, and a
 //! pre-computed commitment into one immutable type.
 
-use ragu_arithmetic::CurveAffine;
+use ff::Field;
+use group::Curve as _;
+use ragu_arithmetic::{CurveAffine, FixedGenerators};
+use rand::CryptoRng;
+
+/// Trait for types that can produce a Pedersen commitment in projective form.
+///
+/// Implement this to make a polynomial type usable with [`batch_commit`] and
+/// [`batch_commit_with_blinds`].
+pub trait Committable<C: CurveAffine> {
+    /// Compute a commitment in projective (non-normalized) form.
+    fn commit<G: FixedGenerators<C>>(&self, generators: &G, blind: C::ScalarExt) -> C::Curve;
+}
+
+/// Commit to `N` polynomials in a single batch, sampling fresh blinding
+/// factors from `rng`.
+///
+/// Performs only one batch inversion for all affine normalizations. Returns an
+/// array of [`CommittedPolynomial`]s in the same order as the input array.
+pub fn batch_commit<P, G, C, RNG, const N: usize>(
+    rng: &mut RNG,
+    generators: &G,
+    polys: [P; N],
+) -> [CommittedPolynomial<P, C>; N]
+where
+    P: Committable<C>,
+    G: FixedGenerators<C>,
+    C: CurveAffine,
+    RNG: CryptoRng,
+{
+    let blinds: [C::ScalarExt; N] = core::array::from_fn(|_| C::ScalarExt::random(&mut *rng));
+    batch_commit_with_blinds(generators, polys, blinds)
+}
+
+/// Commit to `N` polynomials in a single batch using the provided blinding
+/// factors.
+///
+/// Performs only one batch inversion for all affine normalizations. Returns an
+/// array of [`CommittedPolynomial`]s in the same order as the input arrays.
+pub fn batch_commit_with_blinds<P, G, C, const N: usize>(
+    generators: &G,
+    polys: [P; N],
+    blinds: [C::ScalarExt; N],
+) -> [CommittedPolynomial<P, C>; N]
+where
+    P: Committable<C>,
+    G: FixedGenerators<C>,
+    C: CurveAffine,
+{
+    let projectives: [C::Curve; N] =
+        core::array::from_fn(|i| polys[i].commit(generators, blinds[i]));
+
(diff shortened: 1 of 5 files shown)


crates/ragu_circuits/src/polynomials/committed.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.commitment
+    }
+
+    /// Constructs a `CommittedPolynomial` from raw parts **without** verifying
+    /// that the commitment is consistent with the polynomial and blind.
+    ///
+    /// Intended for cases where the commitment is known externally (e.g. from
+    /// a proof transcript) or for tests that deliberately craft an inconsistent
+    /// triple.
+    pub fn new_unchecked(poly: P, blind: C::Scalar, commitment: C) -> Self {
```
TalDerei: currently unused

crates/ragu_circuits/src/polynomials/structured.rs
```diff
@@ -317,6 +272,100 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
...
+    }
+    fn add_assign(r: &mut Self, other: &Self) {
+        RawPolynomial::add_assign(r, other);
+    }
+    fn sub_assign(r: &mut Self, other: &Self) {
+        RawPolynomial::sub_assign(r, other);
+    }
+}
+
+impl<F: Field, R: Rank> PartialEq for Polynomial<F, R> {
```
TalDerei: noticed an inconsistency between structured and unstructured – unstructured derives both `PartialEq + Eq`, but structured only has a manual `PartialEq` (missing `Eq`). should these be derived or have manual impls?

alxiong: Yeah, I think both can be auto-derived, i don't see a problem with that. will change.

crates/ragu_circuits/src/polynomials/committed.rs
TalDerei: based on https://github.com/tachyon-zcash/ragu/pull/515#discussion_r2869645946, `from_parts()` seems more idiomatic based on call site usage.

crates/ragu_pcd/src/fuse/_10_p.rs
```diff
@@ -67,8 +67,8 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
     where
         D: Driver<'dr, F = C::CircuitField, MaybeKind = Always<()>>,
     {
-        let mut poly = f.poly.clone();
-        let mut blind = f.blind;
+        let mut poly = f.poly.poly().clone();
+        let mut blind = f.poly.blind();
```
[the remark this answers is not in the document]
alxiong: oops, missed it. 
I will rename `P.aggregated` to `agg_qx` instead to imply "aggregated quotient", and `F.poly` to `.aggregated` to imply "final aggregated polynomial"

crates/ragu_circuits/src/polynomials/committed.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use ff::Field;
+use ragu_arithmetic::{CurveAffine, FixedGenerators};
+use rand::CryptoRng;
+
+use crate::polynomials::{Rank, structured, unstructured};
+
+/// A trait implemented by polynomial types that know how to Pedersen-commit
+/// themselves, producing a [`CommittedPolynomial`].
+pub trait Committable<C: CurveAffine>: Sized {
```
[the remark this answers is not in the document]
alxiong: done in [fc7fd66](https://github.com/tachyon-zcash/ragu/pull/515/commits/fc7fd6600f11c4742982f3ff104fd611350a52f9)

TalDerei (commented):
I think there's some unnecessary indirection in the commit API here (three layers doing the same thing: `RawPolynomial::commit(), Committable::commit(), batch_commit()`). I'd suggest removing the `Committable` trait, `batch_commit()` and `batch_commit_with_blinds()`. 

The `Committable` impls are pass-throughs, since `Polynomial` already derefs to `RawPolynomial`. `Committable` trait only exists from what I can see to give `batch_commit` a generic bound, but we don't need that because there's already a `batch_to_affine` method (which can replace `batch_commit` itself).

As an aside, it generally seems less error prone that instead of `batch_commit(rng, generators, [poly])` which hides the blind sampling, normalization, and wrapping behind one opaque call, you can see each step explicitly rather than hiding behind a trait. 

Suggesting keeping `CommittedPolynomial` + `from_parts()`, and the `commit() / commit_to_affine() / 
batch_to_affine` conventions from #549 which are already pretty minimal in `structured.rs` / `unstructured.rs` (in fact some of those methods are currently _used_ after the rebase and should be used). Call sites in fuse and other places then collapse to either `commit_to_affine` or `batch_to_affine` (some supplemental context in https://github.com/tachyon-zcash/ragu/pull/549#discussion_r2902674709). Summarizing this, `Polynomial` already has all the commit utilities it needs, via deref to `RawPolynomial`. cc @ebfull

TalDerei:
this should probably be closed at this point since a new approach will require an entire restructuring, and we can reference this PR in that implementation. closed https://github.com/tachyon-zcash/ragu/pull/554

PR #519, opened 2026-03-03
circuits: simple parallelizable witness generation

ebfull:
References https://github.com/tachyon-zcash/ragu/issues/63. Like #476, except no metrics needs to be passed in. Slower in many ways due to the added sort, but simpler architecturally.

### DIFF
--- a/crates/ragu_circuits/src/rx.rs
+++ b/crates/ragu_circuits/src/rx.rs
@@ -24,17 +24,27 @@ use super::{
     structured,
 };
 
-/// A deferred routine evaluation, queued for serial execution.
+/// Deferred `execute()` for a Known-predicted routine.                                                                                                                                                                                            
+///                                                                                                            
+/// Created when `predict()` returns [`Known`](Prediction::Known), allowing                                                                                                                                                                        
+/// the main traversal to continue with the predicted output while deferring                                   
+/// the actual witness computation. When invoked, runs `execute()` in a fresh
+/// [`Evaluator`] and returns the resulting trace segments.
 struct Thunk<'env, F: Field>(
     Box<dyn FnOnce(&mut Vec<Thunk<'env, F>>) -> Result<Vec<AnnotatedSegment<F>>> + Send + 'env>,
 );
+
+impl<'env, F: Field> Thunk<'env, F> {
+    fn run(self, thunks: &mut Vec<Thunk<'env, F>>) -> Result<Vec<AnnotatedSegment<F>>> {
+        (self.0)(thunks)
+    }
+}
 
 /// A contiguous group of multiplication gates.
 ///
 /// Segment 0 is the root segment and holds the placeholder `ONE` gate
 /// at position 0. Routine calls create additional segments (see
 /// [`Evaluator::routine`]).
-#[derive(Clone)]
 pub(crate) struct Segment<F> {
     pub(crate) a: Vec<F>,
     pub(crate) b: Vec<F>,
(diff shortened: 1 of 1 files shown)


crates/ragu_circuits/src/rx.rs
```diff
@@ -162,34 +177,56 @@ impl<F: Field> Trace<F> {
...
+    dfs_prefix: Vec<usize>,
 }
 
-struct Evaluator<'a, F: Field> {
-    trace: &'a mut Trace<F>,
+struct Evaluator<'scope, 'env, F: Field> {
+    segments: Vec<AnnotatedSegment<F>>,
+    thunks: &'scope mut Vec<Thunk<'env, F>>,
     state: TraceScope,
 }
```
TalDerei:
```suggestion
/// Driver that records multiplication gates into trace segments.
struct Evaluator<'scope, 'env, F: Field> {
    /// Trace segments produced by this evaluator's routine scope.
    segments: Vec<AnnotatedSegment<F>>,
    /// Deferred `execute()` closures for Known-predicted routines.
    thunks: &'scope mut Vec<Thunk<'env, F>>,
    /// Per-routine state saved and restored across routine boundaries.
    state: TraceScope,
}

File: crates/ragu_circuits/src/rx.rs [hunk 2]
```diff
@@ -262,18 +336,43 @@ pub fn eval<'witness, F: Field, C: Circuit<F>>(
...
 
-        aux.take()
-    };
-    Ok((trace, aux))
+        Ok((evaluator.segments, aux))
+    }?;
+
+    while let Some(thunk) = thunks.pop() {
+        segments.extend(thunk.0(&mut thunks)?);
+    }
```
TalDerei:
```suggestion
    // Flush deferred execute() calls, popping until all nested
    // Known thunks are drained.
    while let Some(thunk) = thunks.pop() {
        segments.extend(thunk.0(&mut thunks)?);
    }

File: crates/ragu_circuits/src/rx.rs [hunk 3]
```diff
@@ -262,18 +336,43 @@ pub fn eval<'witness, F: Field, C: Circuit<F>>(
...
-        let (io, aux) = circuit.witness(&mut dr, Always::maybe_just(|| witness))?;
-        io.write(&mut dr, &mut ())?;
 
-        aux.take()
-    };
-    Ok((trace, aux))
+        Ok((evaluator.segments, aux))
+    }?;
+
+    while let Some(thunk) = thunks.pop() {
```
alxiong: add TODO on `while let Some(thunk) = thunks.pop()` for potential multithreading in future PR

crates/ragu_circuits/src/rx.rs
alxiong: more documentation on the design.

crates/ragu_circuits/src/rx.rs
```diff
@@ -231,25 +268,62 @@ impl<'a, F: Field> Driver<'a> for Evaluator<'a, F> {
...
+                    },
+                    |this| routine.execute(this, input, aux),
+                )
+            }
+        }
+    }
+}
+
+/// Sorts segments by DFS path and strips annotations.
+fn finish<F: Field>(mut segments: Vec<AnnotatedSegment<F>>) -> Trace<F> {
```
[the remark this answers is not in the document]
TalDerei: this is the final assembly step which sorts by DFS path and strips the annotation wrapper, and your naming suggestions don't cover the entire scope of what `finish` does. think it's fine to keep it for now?

PR #524, opened 2026-03-04
Prop test for DFS order consistency

alxiong:
supersede https://github.com/ebfull/ragu/pull/2 and subsume all individual unit tests in #476 (as @TalDerei concurred)

Ensuring that `rx::eval` and `metric::eval` produce the Segment witness (And records) in the same DFS order using a proptest.


alxiong:
You are absolutely right! My mistake to miss that. Rectified in [f244779](https://github.com/tachyon-zcash/ragu/pull/524/commits/f244779995ea7d50a5278ab578f6de87ce951a7d)

iiuc, the proptest test runner will explore all 4 combinations (inner/outer ⊗ `Knonw/Unknown`)

PR #527, opened 2026-03-04
Disambiguate Buffer and Sink as two types of destination buffers

alxiong:
Motivated by the discussion: https://github.com/tachyon-zcash/ragu/pull/522#discussion_r2881409658.

Currently we mix two purposes of `Gadget.write()` into a single `trait Buffer::write(&mut self, dr: &mut D, val: &Element)`:
- serialization: writing elements in gadgets to a buffer (vector, array etc.) 
- stream processing: `Horner`, `Sponge`, `Ky`, `Pipe` are taking the fields in gadgets and compute some derivative state.

It started to affect readability as discerning the actual purpose take a bit of sniffing around the site of `.write()`.
This confusion also leads us to add unnecessary and mystical-looking `io.write(dr, &mut ())` as the motivating comment.
When i first encounter `trait Buffer::write()`, it's also unclear to me that why "serialization" requires potential state-mutation on `&mut dr`, which feels very counterintuitive. (only later do I understand the multiple purpose of this trait).

--- 

This PR proposes to separate them into two traits:

```rust
/// A plain destination for [`Element`]s that does not cause any side effects
/// on the driver context when being written.
pub trait Buffer<'dr, D: Driver<'dr>> {
    /// Push a single [`Element`] into this buffer.
    fn write(&mut self, value: &Element<'dr, D>) -> Result<()>;
}

pub trait Write<F: Field>: GadgetKind<F> {
    /// Write this gadget's elements into the provided buffer.
    fn write_gadget<'dr, D: Driver<'dr, F = F>>(
        this: &Bound<'dr, D, Self>,
        buf: &mut impl Buffer<'dr, D>,
    ) -> Result<()>;
}

/// Represents a general destination for [`Element`]s to be written to and
/// processed using a provided [`Driver`] context. This may cause side effects
/// on the driver in contrast to a [`Buffer`], which is a passive destination
/// for elements.
pub trait Sink<'dr, D: Driver<'dr>> {
    /// Push an `Element` into this sink using the provided driver `D`.
    fn write(&mut self, dr: &mut D, value: &Element<'dr, D>) -> Result<()>;
}
```

then each gadget has:
```rust
pub trait GadgetExt<'dr, D: Driver<'dr>>: Gadget<'dr, D> {
    fn write(&self, buf: &mut impl io::Buffer<'dr, D>) -> Result<()>;
    fn sink<B: Sink<'dr, D>>(&self, dr: &mut D, buf: &mut B) -> Result<()>;
```

It become semantically clear and less error-prone to know when to use which. 
To demonstrate, in `ky.rs`, it's obviously clear 
```rust
    let mut pubinputs = vec![];
    circuit
        .instance(&mut dr, Always::maybe_just(|| instance))?
        .write(&mut pubinputs)?;

// instead of 
    let mut pubinputs = vec![];
    circuit
        .instance(&mut dr, Always::maybe_just(|| instance))?
        .sink(&mut dr, &mut pubinputs)?;
```

because the intention is clearly writing to some buffer, thus use `.write()` not `sink()`.

This change also would have prevented us from writing the unnecessary `.write()` imo.

### DIFF
--- a/crates/ragu_primitives/src/io.rs
+++ b/crates/ragu_primitives/src/io.rs
@@ -45,6 +45,15 @@ pub trait Buffer<'dr, D: Driver<'dr>> {
 /// Gadgets that consist mainly of other gadgets are candidates for [automatic
 /// derivation](derive@Write) of this trait.
 pub trait Write<F: Field>: GadgetKind<F> {
+    // TODO: use `const LEN` once `const-generic-expressions` is stabilized.
+    /// Returns the number of [`Element`]s this gadget writes via [`write_gadget`].
+    ///
+    /// This must always return the same value for a given type and must agree
+    /// with the actual number of elements written by [`write_gadget`].
+    ///
+    /// [`write_gadget`]: Write::write_gadget
+    fn len() -> usize;
+
     /// Write this gadget's elements into the provided buffer.
     fn write_gadget<'dr, D: Driver<'dr, F = F>>(
         this: &Bound<'dr, D, Self>,
--- a/crates/ragu_circuits/src/rx.rs
+++ b/crates/ragu_circuits/src/rx.rs
@@ -452,6 +452,10 @@ mod tests {
     }
 
     impl<F: ff::Field> ragu_primitives::io::Write<F> for Kind![F; @MulOnSink<'_, _>] {
+        fn len() -> usize {
+            1
+        }
+
         fn write_gadget<'dr, D: Driver<'dr, F = F>>(
             this: &MulOnSink<'dr, D>,
             buf: &mut impl ragu_primitives::io::Buffer<'dr, D>,
--- a/crates/ragu_macros/src/derive/gadgetwrite.rs
+++ b/crates/ragu_macros/src/derive/gadgetwrite.rs
@@ -2,13 +2,13 @@ use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, quote};
 use syn::{
     AngleBracketedGenericArguments, Data, DeriveInput, Error, Fields, GenericParam, Generics,
-    Ident, Result, parse_quote, spanned::Spanned,
+    Ident, Result, Type, parse_quote, spanned::Spanned,
 };
 
 use crate::{
     helpers::{GenericDriver, attr_is},
     path_resolution::{RaguCorePath, RaguPrimitivesPath},
-    substitution::replace_driver_field_in_generic_param,
+    substitution::{normalize_for_kind_macro, replace_driver_field_in_generic_param},
 };
 
 pub fn derive(
@@ -53,7 +53,7 @@ pub fn derive(
         Skip,
     }
 
-    let fields: Vec<(Ident, FieldType)> = match data {
+    let fields: Vec<(Ident, FieldType, Type)> = match data {
         Data::Struct(s) => {
             let fields = match &s.fields {
                 Fields::Named(named) => &named.named,
@@ -73,9 +73,9 @@ pub fn derive(
                 let is_phantom = f.attrs.iter().any(|a| attr_is(a, "phantom"));
 
                 if is_skip || is_phantom {
-                    res.push((fid, FieldType::Skip));
+                    res.push((fid, FieldType::Skip, f.ty.clone()));
                 } else {
-                    res.push((fid, FieldType::Serialize));
+                    res.push((fid, FieldType::Serialize, f.ty.clone()));
                 }
             }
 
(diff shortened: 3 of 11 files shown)


crates/ragu_primitives/src/io.rs
```diff
@@ -38,19 +45,19 @@ pub use pipe::Pipe;
 /// Gadgets that consist mainly of other gadgets are candidates for [automatic
 /// derivation](derive@Write) of this trait.
 pub trait Write<F: Field>: GadgetKind<F> {
-    /// Write this gadget into wires that are written the provided buffer,
-    /// using the driver to synthesize the elements if needed.
-    fn write_gadget<'dr, D: Driver<'dr, F = F>, B: Buffer<'dr, D>>(
+    /// Write this gadget's elements into the provided buffer.
+    fn write_gadget<'dr, D: Driver<'dr, F = F>>(
```
TalDerei: need to document on that `Write` impl something along the lines that "this method must _not_ synthesize constraints; use `Sink` for that"

alxiong:
Our current `Buffer` captures a broad class f `write()`-able buffers. 

The core claim behind this PR is:
There are **two disjoint sets of buffers** based on their intended usage -- serialization and stream processing.

### Against arguments

But as communicated OOB with @ebfull, there are buffer destination that want to accept both (thus implement both `Buffer` and `Sink`).
One example is `multipack(&[Boolean]) -> Vec<Element>` sinking during which linear constraints are added.
https://github.com/tachyon-zcash/ragu/blob/67632fe3dbc8dfd3baad2a704c011cf361860694/crates/ragu_primitives/src/boolean.rs#L216-L219

In this case, `Vec<Element>` needs to implement `Sink` (in my terminology), but meanwhile it can also be used as plain buffer that implements `Buffer`. 
This scenario corroborates with @TalDerei's concern about leading to error-prone implementor:
> circuit developers need to maintain context on both Write and Sink now, and I wouldn’t be sure when to use which as circuit developer (when I write to something like a buffer, why should I have to know whether to do it with or without a driver?)

### For argument

My reaction to the `multipack` example is that, we _should_ have two buffer types for `Vec<Element>`. The one that sunk from `[Boolean;N]` should be named `Unpacked(Vec<Element>)` to signal that elements inside are constrained during construction. imo, this buffer should NOT piggyback on the plain `Vec<Element>` buffer.

I don't envision any reliance on specialized implementation (unstable rust). I contend that distinct buffer types for different purposes is an intentional design and will obviate the need for specialization.  

I'm open to challenges and counterargument to the core claim above. 

Here's what I have in mind for future circuit dev: 
- Gadget contains structured `Elements`, the proc-macro-implemented `Write::write_gadget(this, buf: &mut Buffer)` is conceptually a `ToElements for MyGadget`.
- serialization buffers are happy with those serialized elements squeezed out of a gadget. (e.g. `public_outputs: Vec<Elements>`, they implement `Buffer` only.
- processing buffers will add driver logic or constraints, they implement `Sink` only. **Dev should think of `Sink::write()` as a streaming constructor `Self::from_elements_iter()`**. Given a streaming source of `Elements`, how to construct `Self` with proper constraints imposed. 
- When writing gadgets, nothing to worry or implement. (`GadgetExt` already take care of `.write(buf: &mut Buffer)` and `.sink(buf: &mut Sink)` for you, and you don't overwrite them ever)
- When introducing buffer types, think about which kind of buffer is this. decide to implement either `Buffer` or `Sink`.
  - if you have two paths (deserialized from elements), then _it's a sign that you should introduce a new type_ (possibly a newtype).
- When invoking `gadget -> buffer` logic, just look at the destination buffer type, it should be obvious whether you should use `gadget.write()` or gadget.sink()`. 

Debate me please @TalDerei @ebfull

alxiong:
Some context for the last commit.

Previously the `PaddedContent` gadget only holds the inner `gadget` and relies on `.write()`-time padding up to `HEADER_SIZE`.

Now with the refactor, gadget is supposed to produce driver-independent (non-alloc) `ToElements` logic during `G::write_gadget()`. In other words, `write_gadget(this, buf: &mut impl Buffer)` doesn't take in a driver and don't assume any allocation. Therefore, we need upfront allocation of padding at construction time.

This requires knowledge of the padding length, which we deduce from `HEADER_SIZE - G::len()` where `Write::len()` is a new method that indicates how many elements is the gadget expected to be serialized into.

- For the `#[derive(Write)]` to produce the correct `.len()` implementation, we need know the `GadgetKind` of the inner gadget of `PaddedContent`, thus you see the extra parsed type info in `gadgetwrite.rs`:
  `let fields: Vec<(Ident, FieldType, Type)> = match data`
- What I decide to do is expand through `<Kind![F; G] as Write>::len()`, and since`G` might have `'dr, D` lifetime, they need to be substituted. See `normalize_for_kind_macro()` doc for the necessary substitution.
- in `substitution.rs`, we now need to deal with `Type::Macro` variant for both `Kind![]` and for `unified_output_type!(Point, 'dr, D, C)` type fields in a gadget struct. Thus `impl Substitution for Macro` handles both of those cases.

With all that, we finally can auto-derive `Gadget` for 

```rust
#[derive(Gadget, Write)]
pub(crate) struct PaddedContent<
    'dr,
    D: Driver<'dr>,
    G: GadgetKind<D::F> + Write<D::F>,
    const HEADER_SIZE: usize,
> {
    #[ragu(gadget)]
    gadget: Bound<'dr, D, G>,
    #[ragu(gadget)]
    padding: FixedVec<Element<'dr, D>, PadLen<HEADER_SIZE, D::F, G>>,
}
```

PR #528, opened 2026-03-05
replace FromDriver trait with WireMap

ebfull:
Inspired by #526, and recent experiences with `FromDriver`, I asked Claude to clean it up. Everything here is generated by it:

## Summary

- Replace the `FromDriver<'dr, 'new_dr, D>` trait with a new `WireMap<F>` trait in `ragu_core::convert`, eliminating the extra lifetime parameter and moving `Driver` bounds to associated type bounds on `WireMap` (e.g. `WM: WireMap<F, Src: Driver<'src, F = F>, Dst: Driver<'dst, F = F>>`)
- Add `EraseWires<D>`, a `WireMap` that discards wire values for wireless emulators (replaces the blanket `FromDriver` impl on `Emulator`)
- Add `CloneWires<Src, Dst>`, a `WireMap` that clones wires unchanged, with a `CloneWires::convert()` convenience method that replaces verbose manual adapter structs (like the `Lifter` in `rx.rs` and the manual `DemotedDriver` mapping in `promotion.rs`)
- Add `Emulator::predict()`, which consolidates the repeated three-line boilerplate of creating a wireless emulator, erasing input wires, and calling `Routine::predict` into a single associated function — simplifying all six `Driver::routine` impls to one-liners
- Rename lifetime pairs from `'dr`/`'new_dr` to `'src`/`'dst` and parameter names from `ndr` to `wm` throughout the `GadgetKind`/`Gadget` API
- Update the `#[derive(Gadget)]` proc macro to emit `WireMap`-based code
- Update book references from `FromDriver` to `WireMap`

## Motivation

`FromDriver<'dr, 'new_dr, D>` carried two lifetime parameters and a full `Driver` bound at the trait level, which infected every signature that touched it — including every `GadgetKind::map_gadget` impl, the derive macro output, and every ad-hoc conversion adapter. The new `WireMap<F>` design pushes source and destination into associated types (`Src`/`Dst`), so the trait itself is parameterized only by the field. `Driver` constraints move to associated type bounds directly on the `WireMap` bound rather than appearing in separate where-clauses, keeping signatures shorter and more uniform.

The old design also conflated two roles: `FromDriver` was both a wire conversion trait *and* the mechanism for creating wireless emulator inputs (via a blanket impl on `Emulator`). Splitting these into `WireMap` (general wire conversion), `EraseWires` (discard wires for prediction), and `CloneWires` (pass-through cloning) makes each conversion strategy explicit and composable. This eliminated the `Lifter` adapter in `rx.rs` entirely and simplified `DemotedDriver` in `promotion.rs` to a one-line `CloneWires::convert` call.

Every `Driver::routine` impl also repeated the same three-line pattern to predict auxiliary data: create an emulator, remap the input, call predict. `Emulator::predict()` encapsulates this, making the six callsites across `ragu_core` and `ragu_circuits` uniform and removing direct dependence on `EraseWires` and `GadgetKind` from driver implementations.

### DIFF
--- a/crates/ragu_core/src/convert.rs
+++ b/crates/ragu_core/src/convert.rs
@@ -1,20 +1,20 @@
-//! Conversions that remap gadget wires from one [`Driver`] context to another.
+//! Conversions that translate gadgets from one [`Driver`] context to another.
 //!
-//! [Routines](crate::routines) execute circuit synthesis within well-defined
-//! abstraction boundaries and may need to move gadgets (and their wires) from
-//! one [`Driver`] context to another. For example, [`Routine::predict`] is
-//! evaluated on a [`Wireless`] [`Emulator`] so that no constraints are
-//! synthesized, but the input gadget originates from a concrete source driver.
-//! Without a uniform conversion mechanism, every call site would need custom
-//! remapping logic.
+//! Gadgets are polymorphic over drivers, and translating a gadget from one
+//! driver context to another while preserving its structure and semantics is a
+//! fundamental operation. Any code that operates across multiple driver
+//! contexts will need this: [routines](crate::routines) translate their inputs
+//! onto [`Wireless`] [`Emulator`]s for prediction, wire-counting passes discard
+//! wire values entirely, and driver implementations may need to inject or
+//! rewrite wires during circuit analysis.
 //!
-//! [`WireMap`] provides that mechanism: an implementor fixes a source and
-//! destination driver via associated types, then converts wires one at a time.
-//! Because [`WireMap`] is a separate trait rather than an associated type on
-//! [`GadgetKind`](crate::gadgets::GadgetKind), the same gadget kind can be
-//! remapped by many different conversion strategies (discarding wires, cloning
-//! them, recording them) without adding type parameters to [`GadgetKind`](
-//! crate::gadgets::GadgetKind).
+//! [`WireMap`] provides a uniform mechanism for these conversions: an
+//! implementor fixes a source and destination driver via associated types, then
+//! converts wires one at a time. Because [`WireMap`] is a separate trait rather
+//! than an associated type on [`GadgetKind`](crate::gadgets::GadgetKind), the
+//! same gadget kind can be remapped by many different conversion strategies
+//! without adding type parameters to
+//! [`GadgetKind`](crate::gadgets::GadgetKind).
 //!
 //! ### Public API
 //!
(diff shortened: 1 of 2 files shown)


crates/ragu_core/src/drivers/emulator.rs
```diff
@@ -388,12 +401,28 @@ fn short_circuit_routine<'dr, D: Driver<'dr>, R: Routine<D::F> + 'dr>(
...
-/// [`Routine::predict`] to fulfill type system obligations.
-impl<'dr, D: Driver<'dr>> FromDriver<'dr, '_, D> for Emulator<Wireless<D::MaybeKind, D::F>> {
-    type NewDriver = Self;
+/// Maps any driver's wires to `()` for use with
+/// `Emulator<Wireless<D::MaybeKind, D::ImplField>>`, discarding wire
+/// values.
+///
+/// Useful for passing a gadget from a concrete driver into
+/// [`Routine::predict`], which operates
+/// on a [`Wireless`] emulator. A wrapper struct is required because
```
alxiong: nit: nice to add half/one more sentence that "wireless emulator for memory efficient witness computation with wires erased" to indicate the motivation to switch different driver.

alxiong (commented):
I like the general direction, but disagree with some naming.

- `CloneWires` -> `CloneWireMap`: former sounds like a type of wire or a type that holds wire. 
  - `CloneWires::convert` -> `CloneWireMap::map()` or `::apply()`
- `WirelessFrom` -> `EraseWireMap`: former sounds like a trait. (the assumed usage from the name is: `WirelessFrom::from(gadget) -> Emulator<Wireless>`).

crates/ragu_core/src/convert.rs
```diff
@@ -0,0 +1,90 @@
...
+        &mut self,
+        wire: &<Self::Src as DriverTypes>::ImplWire,
+    ) -> Result<<Self::Dst as DriverTypes>::ImplWire>;
+}
+
+/// A [`WireMap`] that passes wires through unchanged by cloning them.
+///
+/// Useful when the source and destination share the same wire type --- for
+/// example, when demoting a driver to strip witness data while preserving
+/// wire values.
```
ebfull: Too specific to mention in these docs.

crates/ragu_core/src/convert.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl<F: Field, Src, Dst> CloneWires<Src, Dst>
+where
+    Src: DriverTypes<ImplField = F>,
+    Dst: DriverTypes<ImplField = F, ImplWire = Src::ImplWire>,
+{
+    /// Maps a gadget to a destination driver by cloning its wires.
+    ///
+    /// `Src` is inferred from the gadget and `Dst` from the return context,
+    /// so no turbofish is needed:
```
ebfull: I generally frown on this kind of justification; the AI put this here because I complained about something using turbofish, it did a refactor to remove it, but then grew obsessed with documenting how bad turbofish is.

PR #530, opened 2026-03-05
chore: `WireMap` suggestions

TalDerei:
Post-hoc review and suggestions for https://github.com/tachyon-zcash/ragu/pull/528.


ebfull (changes requested):
Nice! I experimented with this locally, but I must have gone down a slightly different path and didn't like it. But this looks quite nice.

I think a great way to make this slightly better would be to unify the `remap` and `strip` methods under the trait itself, using the same name (`remap`): use a provided method impl, that `where` bounds `Self: Default`. This also makes sense at the callsite, because `StripWires::strip` is redundant anyway.

PR #533, opened 2026-03-06
book: guide/structure improvements

ebfull:
Also adds a cute TODO badge to pages that haven't gotten my official blessing yet.

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -15,10 +15,14 @@ command = "mdbook-mermaid"
 command = "mdbook-admonish"
 assets_version = "3.0.0"
 
+[preprocessor.todo]
+command = "python3 preprocessor-todo.py"
+after = ["links"]
+
 [output]
 
 [output.html]
-additional-js = ["mermaid.min.js", "mermaid-init.js"]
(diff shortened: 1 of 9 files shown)


crates/ragu_core/src/drivers/emulator.rs
```diff
@@ -104,78 +104,12 @@ pub trait Mode {
 /// Wired mode always has witness availability (i.e., `MaybeKind = Always<()>`).
 pub struct Wired<F: Field>(PhantomData<F>);
 
-/// Container for a [`Field`] element representing a wire assignment.
-///
-/// Wire values are exposed through [`Emulator::wires`].
-pub enum WiredValue<F: Field> {
```
TalDerei: tracing through the commit history on why this type was introduced, it seems like it was a simplification of a previous type, `MaybeWired<M, F>` (parameterized by `MaybeKind`). It was used until `Wired` changed, which seemed to make it extraneous. it's clear this is duplicative against `DirectSum`'s LC impl which raw field element `F` can use directly without this kind of indirection.

PR #534, opened 2026-03-06
core: cleanups

ebfull:
Big thing here is that `Consistent` is moved to `ragu_primitives`. I don't think users will need this outside of internal usage in Ragu. Can revisit this later. I take the liberty of removing the book entry too, which is nice.

The rest of the changes are automated documentation improvements and added test coverage. I'll take a look at the generated tests because I suspect a lot of them are useless or have copious amounts of type annotations... maybe that's not a big deal, since they're not really hurting anything.


TalDerei (approved):
ACK – pure structural migration, no logic changes from https://github.com/tachyon-zcash/ragu/pull/388. modulo potentially extraneous tests (maybe consider consolidating in a proptest for `WireMap`?)

PR #535, opened 2026-03-06
Minor PCD API refactor

alxiong:
I extract the less contentious changes in #392 out:

- `Application::fuse() -> Result<(Pcd, Aux)>` instead of returning `Proof` and forcing the caller to manually (and immediately `proof.carry()` to get `pcd`)
  - Tal [expressed his support](https://github.com/tachyon-zcash/ragu/pull/392#pullrequestreview-3753099210) for such elimination.
  - keeping `Aux` as Sean [suggested](https://github.com/tachyon-zcash/ragu/pull/392#issuecomment-3806050310) for future use 
- In `Step::witness() -> (headers, output, aux)`, I added a `output: DriverValue<<Output as Header>::Data>` so that we don't abuse the `aux` field for step output. 
  - mostly in tests like `ragu_testing/src/pcd/nontrivial.rs`: `WitnessLeaf` and `Hash2`; `tests/rerandomization.rs`: `StepWithData`.

### DIFF
--- a/crates/ragu_core/src/drivers.rs
+++ b/crates/ragu_core/src/drivers.rs
@@ -278,7 +278,7 @@ pub trait Driver<'dr>: DriverTypes<ImplWire = Self::Wire, ImplField = Self::F> +
     /// Convenience method returning a unit [`DriverValue`]. Equivalent to
     /// `D::just(|| ())`.
     fn unit() -> DriverValue<Self, ()> {
-        <DriverValue<Self, ()> as Maybe<()>>::unit()
+        Self::just(|| ())
     }
 
     /// Executes a routine with this driver.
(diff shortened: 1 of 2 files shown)


crates/ragu_pcd/src/step/mod.rs
```diff
@@ -167,7 +163,15 @@ pub trait Step<C: Cycle>: Sized + Send + Sync {
     /// The header produced during this step.
     type Output: Header<C::CircuitField>;
 
+    /// Auxiliary information produced during circuit synthesis. This may be
+    /// necessary to construct the [`Header::Data`] for the resulting proof,
+    /// or to pipeline witness data to future steps.
+    type Aux<'source>: Send;
```
ebfull: Comment is stale, `Aux` isn't used to help make the `Header::Data` anymore.

Issue #501, opened 2026-02-20
Document `ragu_primitives` as a standard library


TalDerei:
Architecture docs (https://tachyon.z.cash/ragu/implementation/arch.html) describe the `ragu_primitives` crate as providing "implementations of many algorithms and abstractions that use the API in `ragu_core`, mainly providing gadget implementations that are useful for building circuits". It should eventually be documented as modeling a kind of standard library that will be used by circuit developers, and docs should eventually reflect that before any kind of release.

PR #540, opened 2026-03-08
chore(book): standard library docs and no_std crate standardization 

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/501.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -7,7 +7,6 @@
 
 - [Part I: User Guide]()
   - [Requirements](guide/requirements.md)
-  - [The Primitives Standard Library](guide/standard_library.md)
   - [Getting Started](guide/getting_started.md) <!-- todo -->
   - [Drivers](guide/drivers/index.md)
     - [Witness Data](guide/drivers/witness.md)
@@ -19,6 +18,7 @@
     - [The Kind! Macro](guide/gadgets/kind.md)
     - [Conversion](guide/gadgets/conversion.md)
   - [Routines](guide/routines.md)
+  - [The Primitives Standard Library](guide/standard_library.md) <!-- todo -->
   - [Writing Circuits](guide/writing_circuits.md) <!-- todo -->
   - [Configuration](guide/configuration.md) <!-- todo -->
 - [Part II: Protocol Design]()


book/src/guide/requirements.md
```diff
@@ -5,5 +5,68 @@
...
+
+A small number of crates fall outside this constraint:
+
+* `ragu_macros` is a procedural-macro crate, which always runs on the host at
+  compile time and therefore requires `std`.
+* `ragu_testing` is a test-only utility crate and does not need to be
+  `no_std`-compatible.
+* `ragu_pcd` conditionally enables `std` during tests via
+  `#![cfg_attr(not(test), no_std)]`, while remaining `no_std` in release
+  builds.
```
ebfull: I would frame this differently: principles, then dive into specifics if necessary.

* Ragu's crates _all_ do not require `std` by default. (Enabled via `std` feature in each crate.)
* Ragu's crates _all_ do require `alloc` (for access to `Box`/`Vec`)
* Optional performance features can be enabled via crate feature flags. These may depend on `std`.
* Ragu requires `std` on the host for compilation (build scripts, procedural macros) and tests/benchmarks. _This is a pretty standard requirement even for common `no_std` libraries in the Rust ecosystem._

Shouldn't need to identify specific crates and how they might deviate in practice as long as they follow these principles. Also hhelps because we might want to add optional `std` features to some of them later.

ebfull (changes requested):

book/src/SUMMARY.md
```diff
@@ -7,6 +7,7 @@
 
 - [Part I: User Guide]()
   - [Requirements](guide/requirements.md)
+  - [The Primitives Standard Library](guide/standard_library.md)
```
ebfull: I think this should be mentioned later in the guide, since the pages logically expand on their previous pages. This one mentions (without crosslinking) all sorts of concepts like gadgets and stuff that aren't discussed yet.

ebfull (changes requested):

book/src/SUMMARY.md
TalDerei: moving after the "routines" page and before the "writing circuits" page, we can play around with book placements later in a more hollistic pass.

TalDerei: https://github.com/tachyon-zcash/ragu/pull/540/changes/28645645c656ebf177cc69c24815c0f91cc2c984

PR #541, opened 2026-03-08
`ragu_core` API hardening

ebfull:
**API improvements:**
- **Seal the `Mode` trait** to prevent external implementations
- **Make `DirectSum<F>` fields private** with accessor methods, preserving encapsulation
- **Make `Gadget::num_wires` return `Result<usize>`** instead of panicking, propagating errors from inner gadget queries
- **Improve `Error` enum** — switch from positional tuple variants to named fields, add `#[source]` attributes for error chaining, and fix `CircuitBoundExceeded` to report the correct limit

**Documentation:**
- Document `Routine::predict` error semantics (`Err` vs `Unknown`)
- Document `DriverTypes` trait rationale and the lifetime-free design
- Document `Driver` closure purity requirements (`Fn` vs `FnOnce`/`FnMut`)
- Clarify that fungibility is a correctness contract, not a `GadgetKind` safety invariant
- Expand `GadgetKind` doc comment explaining the `Any` supertrait bound
- Book: misc fixups (cross-links, section anchors, footnotes, admonish blocks)

**Tests:**
- Add `DirectSum` coverage for zero-element and gain-interaction cases

### DIFF
--- a/crates/ragu_core/src/drivers/emulator.rs
+++ b/crates/ragu_core/src/drivers/emulator.rs
@@ -84,9 +84,13 @@ use crate::{
     routines::{Prediction, Routine},
 };
 
+mod sealed {
+    pub trait Sealed {}
+}
+
 /// Mode that an [`Emulator`] may be running in; usually either [`Wired`] or
 /// [`Wireless`].
-pub trait Mode {
+pub trait Mode: sealed::Sealed {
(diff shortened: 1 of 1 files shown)


TalDerei (commented):
these kinds of changes tend to leave behind stale book references and missing tests for new error paths, can we check for those?

ebfull:
@TalDerei Tried to test those with `/book-review` as I went.

PR #542, opened 2026-03-08
book: add demotion page

LucidSamuel:
Follow-up to #526.
- Add \`book/src/guide/gadgets/demotion.md\` covering the demote/promote lifecycle
- Convert demotion paragraph in \`conversion.md\` to an admonish block linking to the new page
- Add SUMMARY.md entry after Conversion


ebfull:
Nice! I touched this in https://github.com/tachyon-zcash/ragu/pull/541 I think, so this should be rebased on that when it merges.

Issue #333, opened 2026-01-16
Amortize base-point reads across MSMs sharing generators


TalDerei:
Noted in https://github.com/tachyon-zcash/ragu/pull/324/files#r2699100324. 

During proof generation, we perform many MSMs using the same generator bases but different scalar vectors, and currently each MSM performs independent bucket accumulation. Specifically inside stages, we could also expose a batched MSM API that shares bucket accumulation work across scalar vectors (similar to something like `VariableBaseMSM::multi_scalar_mul` in arkworks). This could live as a standalone utility in `ragu_arithmetic`.

PR #546, opened 2026-03-08
feat(pcd): add fuse criterion bench

TalDerei:
This adds a criterion bench to measure wall-clock time between sequential and parallel fuse. In thinking about https://github.com/tachyon-zcash/ragu/issues/333, I realized this actually wouldn't offer much benefit because you can only apply this base-point read amortization for MSMs clustered in a _single_ stage, and this only partially benefits some stages. Instead, I tried parallelizing the MSM calls in the prover, but realized our [existing](https://github.com/tachyon-zcash/ragu/pull/510) window-level parallelism within each MSM already likely saturates a normal devices available cores. Halo2 also doesn't implement a multi-core prover (batching MSMs calls) because their prover is seemingly fast enough.

`cargo bench --bench fuse_criterion -p ragu_pcd`
`cargo bench --bench fuse_criterion -p ragu_pcd --features multicore`

-----------

**sequential** 

Screenshot 2026-03-07 at 7 24 21 PM

**Intra-MSM parallelism**

Screenshot 2026-03-08 at 9 12 44 AM

**Inter-MSM parallelism**

Screenshot 2026-03-08 at 9 12 48 AM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -812,8 +812,10 @@ dependencies = [
 name = "ragu_pcd"
 version = "0.0.0"
 dependencies = [
+ "criterion",
  "ff",
  "gungraun",
+ "maybe-rayon",
  "pasta_curves",
  "ragu_arithmetic",
  "ragu_circuits",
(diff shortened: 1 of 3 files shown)


TalDerei:
closing in favor of https://github.com/tachyon-zcash/ragu/pull/549

Issue #545, opened 2026-03-08
Batch-normalize projective to affine conversions


TalDerei:
We should use batch normalize (montgomery's trick) to convert N projective points (from commits) to affine with 1 inversion + 3(N-1) multiplications instead of N inversions.

PR #549, opened 2026-03-08
chore(arithmetic): batch normalize in fuse

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/545, an obvious _minor_ optimization that hooks into the existing `batch_normalize` utility.

This also adds a criterion bench to measure wall-clock time between sequential and parallel fuse. In thinking about https://github.com/tachyon-zcash/ragu/issues/333, I realized this actually wouldn't offer much benefit because you can only apply this base-point read amortization for MSMs clustered in a _single_ stage, and this only partially benefits some stages. Instead, I tried parallelizing the MSM calls in the prover, but realized our [existing](https://github.com/tachyon-zcash/ragu/pull/510) window-level parallelism within each MSM already likely saturates a normal devices available cores. Halo2 also doesn't implement a multi-core prover (batching MSMs calls) because their prover is seemingly fast enough.

`cargo bench --bench fuse_criterion -p ragu_pcd`
`cargo bench --bench fuse_criterion -p ragu_pcd --features multicore`

-----------

**sequential** 

Screenshot 2026-03-07 at 7 24 21 PM

**Intra-MSM parallelism**

Screenshot 2026-03-08 at 9 12 44 AM

**Inter-MSM parallelism**

Screenshot 2026-03-08 at 9 12 48 AM

### DIFF
--- a/crates/ragu_circuits/benches/circuits.rs
+++ b/crates/ragu_circuits/benches/circuits.rs
@@ -27,7 +27,7 @@ fn commit_structured(
         (structured::Polynomial<Fp, ProductionRank>, Fp),
     ),
 ) {
-    black_box(poly.commit(generators, blind));
+    black_box(poly.commit_to_affine(generators, blind));
 }
 
 #[library_benchmark(setup = setup_with_rng)]
@@ -38,7 +38,7 @@ fn commit_unstructured(
         (unstructured::Polynomial<Fp, ProductionRank>, Fp),
     ),
 ) {
-    black_box(poly.commit(generators, blind));
+    black_box(poly.commit_to_affine(generators, blind));
 }
 
 library_benchmark_group!(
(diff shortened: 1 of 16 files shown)


crates/ragu_pcd/benches/criterion/fuse.rs
```diff
@@ -0,0 +1,59 @@
+use criterion::{Criterion, criterion_group, criterion_main};
+use ragu_arithmetic::Cycle;
+use ragu_circuits::polynomials::ProductionRank;
+use ragu_pasta::{Fp, Pasta};
+use ragu_pcd::ApplicationBuilder;
+use ragu_testing::pcd::nontrivial;
+use rand::SeedableRng;
+use rand::rngs::StdRng;
+
+fn fuse_bench(c: &mut Criterion) {
```
TalDerei: same commit from https://github.com/tachyon-zcash/ragu/pull/546

ebfull (approved):

PR #552, opened 2026-03-09
More tests for transcript

alxiong:
Follow-up for #422 

I [made a mistake ](https://github.com/tachyon-zcash/ragu/pull/422#discussion_r2897552800)to allow `resume().absorb()` in earlier version of Transcript api design, which has since been fixed by Tal.

This PR add tests and clarification to sponge and transcript:
- assertion of post-condition of absorb, which explain with `values.is_empty()` is enough as a predicate to detect `NothingAbsorb` in the `Absorb` mode
- disallow squeeze when nothing absorbed. Current logic allows it, which squeeze out zero elements. (I believe this is not desirable), thus I return error in that case
- add a test to showcase the "bad state" Sean mentioned, when we absorb immediately after resuming. i.e. wrong cutoff point for state saving.
- turn existing unit tests in `transcript.rs` into proptests.

### DIFF
--- a/crates/ragu_primitives/src/poseidon.rs
+++ b/crates/ragu_primitives/src/poseidon.rs
@@ -137,6 +137,7 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
     }
 
     /// Get the current pending values in the sponge.
+    #[inline(always)]
     fn values(&self) -> &[Element<'dr, D>] {
         match &self.mode {
             Mode::Squeeze { values, .. } => values,
@@ -145,6 +146,7 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
     }
 
     /// Get the current internal state of the sponge.
+    #[inline(always)]
     fn state(&self) -> &SpongeState<'dr, D, P> {
         match &self.mode {
             Mode::Squeeze { state, .. } => state,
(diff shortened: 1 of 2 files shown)


crates/ragu_primitives/src/poseidon.rs
```diff
@@ -135,6 +135,22 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
...
+        }
+    }
+
+    /// Get the current internal state of the sponge.
+    fn state(&self) -> &SpongeState<'dr, D, P> {
+        match &self.mode {
+            Mode::Squeeze { state, .. } => state,
+            Mode::Absorb { state, .. } => state,
+        }
+    }
```
TalDerei: can we inline these at the call site without the helper to keep minimal API surface on `Sponge`?

crates/ragu_primitives/src/poseidon.rs
```diff
@@ -147,17 +163,25 @@ impl<'dr, D: Driver<'dr>, P: ragu_arithmetic::PoseidonPermutation<D::F>> Sponge<
...
+                    ));
+                } else {
+                    // Before we can switch to squeeze mode, we need to permute
+                    // to absorb the pending values into the state.
+                    self.permute(dr)?;
+
+                    // NOTE: breakpoint for state saving would be here:
+                    // On save, we absorb pending values as the last permute;
+                    // On resume, we enter squeeze mode with squeezable values
+                    // copied from the state.
```
TalDerei:
```suggestion
  // This is the same state boundary that save_state/resume operates on:                                                                                                                                                                             
  // save_state permutes pending values, then resume enters squeeze mode                                                                                                                                                                             
  // with rate values extracted from the post-permutation state.

File: crates/ragu_primitives/src/poseidon.rs [hunk 3]
```diff
@@ -443,6 +475,20 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_squeeze_before_any_absorb() -> Result<()> {
```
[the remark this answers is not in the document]
alxiong: okay! i will add one test that expect a panic to ensure our assertion is there.

PR #556, opened 2026-03-10
chore: shared proptest strategy

LucidSamuel:
- Shared proptest strategies: Centralized field element, coefficient, and
polynomial root strategies in `ragu_testing::strategies`, replacing
duplicated per-crate implementations across arithmetic/core/primitives.
Added CI job (`proptests-fast`) and justfile recipes (`proptests_fast`,
`proptests_heavy`) with `--release` for practical runtimes.


TalDerei (commented):
let's remove the third bullet from scope here (we'll [revisit digest security](https://github.com/tachyon-zcash/ragu/issues/78) down the line) and just focus on the shared proptest scaffolding. I'd really like to land something like this in isolation!

crates/ragu_testing/src/strategies.rs
```diff
@@ -0,0 +1,135 @@
+//! Shared proptest strategies for Ragu crates.
+
+use ff::PrimeField;
+use proptest::{prelude::*, sample::select, strategy::BoxedStrategy};
+use ragu_arithmetic::{Coeff, Domain};
+
+fn edge_u64() -> impl Strategy<Value = u64> + Clone {
+    prop_oneof![
+        select(vec![
```
TalDerei: nit; formatting, collapse list.

TalDerei (approved):
see nit observations; In favor of proptest consolidation.

TalDerei (approved):

ebfull (changes requested):
Rebase please and I think I'm fine to merge this. As long as there isn't an issue with cyclic dependencies in the workspace?

ebfull (approved):

Issue #553, opened 2026-03-09
Distinguish shallow vs deep routine fingerprints for memoization


TalDerei:
Note on some necessary work that needs to either be folded into https://github.com/tachyon-zcash/ragu/pull/503 or split out as a separate precursor based on async discussion. There’s a coverage gap in #511 that necessitates distinguishing between what we’re terming “shallow” versus “deep” fingerprints. The context is two routines with identical constraints that return different wires – imagine an outer routine calling two nested routines that have the same internal constraints and return the same wires in swapped order. During the metrics pass, the remap step in some sense anonymizes the output wires of all its subroutines (by discarding the child wire value), which destroys information about the swap. Even with a single output wire, two routines with identical constraints can return _different_ wires. Without tracking this, downstream consumers like memoization can treat them as interchangeable when they're not.

Currently, the computation that produces the fingerprint (the horner scalar) only tracks `enforce_zero` calls and doesn’t feed the output wire into the accumulator. The original suggestion was to fold the output wires of the nested routines into the horner accumulator, specifically after the routine executes and before extracting the fingerprinting, which would’ve sufficiently captured the ordering and addressed this. Instead, we'd like to separate concerns on more defined lines and we’re leaning towards a cleaner _structural_ identity fix where the `RoutineIdentity` struct includes an additional  'output_scalar' field. This field captures which wire flows to which output slot, capturing routine information. While the ‘scalar’ fingerprint just tells you whether routines impose the same constraints, the ‘output_scalar’ tells you whether the routines return the same wires. 

Then, the identity system produces two things: (1) “shallow” fingerprint: a scalar that corresponds to the s(x, y) contribution, and (2) “deep” fingerprint which is a hash of all nested routines’ deep fingerprints + their output fingerprints + our shallow fingerprint + number of child routines written to the hash.

In that context, floor planning only needs shallow fingerprints, but memoization requires deep fingerprints. We'll need to wire our memoization invariants over deep fingerprints, which requires some work.

alxiong:
> floor planning only needs shallow fingerprints, but memoization requires deep fingerprints.

This is only true for intra-circuit planning. once we want to plan across a registry/mesh, we need to align memoizable segments thus need deep fingerprints. 
Plus, current floor planning don't even need any fingerprint, it's a straightforward DFS arrangement, only shape info needed, not even the shadow fingerprint.

alxiong:
[...]

PR #557, opened 2026-03-10
chore(circuits): distinguish base vs deep fingerprints

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/553 and precursor to https://github.com/tachyon-zcash/ragu/pull/503. Aligns with assumptions made in the book entry in https://github.com/tachyon-zcash/ragu/pull/599. Should also sufficiently remediate the conflating concerns flagged in https://github.com/tachyon-zcash/ragu/issues/598.

- `BaseFingerprint`: (`num_gates, num_constraints, eval)`. The routine's local `s(X, Y)` polynomial contribution — no `TypeId`s, since type identity does not affect the polynomial. Used by the floor planner to group segments with the same polynomial shape.
- `DeepFingerprint`: (`base, deep`). The full recursive identity for memoization, where deep is a BLAKE2b hash over all `BaseFingerprint` fields, child count, and children's deep hashes. Used by the memo cache to ensure safe routine substitution.

These are _deliberately_ separate to prevent conflating polynomial equivalence (floor planning) with memoization safety (cache keys). 

~75% of diff is testing. #503 will have real oracle proptests as a ground truth for detecting memoization bugs.


crates/ragu_circuits/src/metrics.rs
```diff
@@ -89,35 +101,94 @@ pub enum RoutineIdentity {
...
+/// Computes the recursive deep hash from a shallow fingerprint, output
+/// evaluation, and child deep hashes.
+///
+/// Uses BLAKE2b to fold: `eval`, constraint counts, `output_eval`, child
+/// count, and each child's deep hash into a single 64-bit digest.
+/// [`TypeId`] fields are intentionally excluded — they are already captured
+/// in [`ShallowFingerprint`]'s `Eq`/`Hash` impl.
+///
+/// [`TypeId`]: core::any::TypeId
+fn deep_hash(shallow: &ShallowFingerprint, output_eval: u64, children: &[u64]) -> u64 {
```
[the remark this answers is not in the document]
TalDerei: couldn't find a way around a custom hasher since `TypeId` is opaque.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -89,35 +102,119 @@ pub enum RoutineIdentity {
...
+    h.finish()
+}
+
+/// Computes the recursive deep hash from all [`ShallowFingerprint`] fields
+/// (including [`TypeId`] pairs), `output_eval`, child count, and each
+/// child's deep hash, producing a single 64-bit BLAKE2b digest.
+///
+/// [`TypeId`]: core::any::TypeId
+fn deep_hash(shallow: &ShallowFingerprint, output_eval: u64, children: &[u64]) -> u64 {
+    let mut state = blake2b_simd::Params::new().personal(b"FIXME").to_state();
```
TalDerei, line 186: temp placeholder like other personalization strings / domain separation, eg. `RAGU_TAG`. `SegmentRecord::identity()` is also still dead code with marked todo.

crates/ragu_circuits/src/tests/identity.rs
```diff
@@ -1829,3 +1832,469 @@ fn test_typeid_one_wire_constraint() {
...
+                fingerprint_elem(&r),
+                fingerprint_via_eval(&r),
+                "cross-path mismatch"
+            );
+        }
+
+        /// Changing a single op in the sequence changes the fingerprint.
+        /// Schwartz-Zippel: collision probability per pair is negligible.
+        #[test]
+        fn interpreted_mutation_sensitivity(
```
TalDerei: `interpreted_mutation_sensitivity` test failure per proptest expansion in https://github.com/tachyon-zcash/ragu/pull/557/commits/e4374e7ff138db12cf9715c21e2ad438f434f4cf (which augments existing identity tests into a proptest with a wider constraint topology). It accidentally asserted a property that doesn't hold, which is that different ops can produce identical constraint structures.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -89,35 +102,125 @@ pub enum RoutineIdentity {
...
+fn type_id_u64(id: TypeId) -> u64 {
+    struct PassU64(u64);
+    impl core::hash::Hasher for PassU64 {
+        fn finish(&self) -> u64 {
+            self.0
+        }
+        fn write(&mut self, _: &[u8]) {
+            // TypeId::Hash calls write_u64; if this fallback is ever
+            // reached the assumption no longer holds.
+            unreachable!("TypeId::Hash should call write_u64");
```
alxiong:
```suggestion
              // Fallback if TypeId::Hash ever stops calling write_u64.
              for (i, &b) in bytes.iter().enumerate().take(8) {
                  self.0 |= (b as u64) << (i * 8);
              }
```

what if future `impl Hash for TypeId` changed to not call `write_u64()`? to be more robust, we may have this fallback instead.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -10,11 +10,22 @@
...
-//! (Schwartz–Zippel).
+//! Routine fingerprints come in two levels:
+//!
+//! - **Shallow** ([`ShallowFingerprint`]): the tuple `(TypeId(Input),
+//!   TypeId(Output), eval, num_mul, num_lc)`. Captures only the routine's own
+//!   constraint shape — type pairs, constraint counts, and the Schwartz–Zippel
+//!   evaluation scalar. Two routines with the same shallow fingerprint impose
+//!   the same local constraints (with overwhelming probability) but may differ
+//!   in output wire mapping or nested subtree structure. Floor planning uses
+//!   shallow fingerprints to group routines with the same constraint shape.
```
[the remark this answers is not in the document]
TalDerei: some supplemental context; on current main, identity is `#[allow(dead_code)]` as you pointed out, but the floor planning in #503 has been [rebased](https://github.com/tachyon-zcash/ragu/pull/503#issuecomment-4029314935) to consume `RoutineFingerprint` in multiple places: `FloorPlan` module, memoization cache, synthesis driver. Inspecting that diff should satisfy your first point. 

regarding the second point, there's a strong semantic argument to keeping them separate which motivated this PR. **Shallow** (renamed to `RoutineFingerprint`) is "segment-level distinction" (whether two individual segments basically impose the same local constraints and could be aligned regardless of their subtree, ie. padding one circuit's segment against another circuits offset or something). **Deep** (renamed to `MemoFingerprint`) is "subtree-level equivalence" (whether entire routine invocations are interchangeable and can be used for memoization). there's a semantic distinction here that gets lost unfortunately when consolidating into a single identity type: a segment can be shallow-equivalent (same local constraints, worth aligning) but deep-inequivalent (different children so can't memoize). I'd like to maintain the architectural separation in that context. cc @ebfull

alxiong (approved):

TalDerei:
Tagging this as ready for another review, aligned with the book entry in #599.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -306,6 +441,13 @@ struct Counter<F> {
     /// `h * y^n + c_1 * y^{n-1} + …`, making the leading power of `y`
     /// always visible.
     h: F,
```
alxiong, line 443: imo this defense-in-depth is speculative without material benefits. Such degenerated case will be reliably caught by the `num_constraint` count. 

i propose we remove this for simplicity.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -506,6 +668,39 @@ impl<F: FromUniformBytes<64>> WireMap<F> for Counter<F> {
     }
 }
```
alxiong, line 669: revisiting this, i recommend we define a `struct ReinitWires` (or `RestartWires` or similar), 

```rust
struct ReinitWires<F> {
  init: F,
  cur: F,
}

impl WireMap for ReinitWires {}
```

the logic of (and the safety guarantee) is more localized, instead of interweaved with `Counter`'s own logic. 
the `next_remap()` should be part of `ReinitWires`'s logic. 

e.g. currently making sure that `self.x_remap` is not muted requires checking lots of places, but if we localize the management to this type of wire conversion, then easier to audit imo.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -92,34 +109,149 @@ pub enum RoutineIdentity {
...
+/// used for security-critical decisions, store the full field
+/// representation (`[u8; 32]`) instead — the cost is negligible.
+///
+/// [`TypeId`]: core::any::TypeId
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct MemoFingerprint {
+    routine: RoutineFingerprint,
+    output_eval: u64,
+    deep: u64,
+}
```
alxiong, line 142: can drop `output_eval`, already captured in `deep` hash value. that's just an intermediate value to compute the deep hash.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -469,17 +615,33 @@ impl<'dr, F: FromUniformBytes<64>> Driver<'dr> for Counter<F> {
...
+        // which wires flow to which output slots.
+        let output_eval = {
+            let mut acc = OutputEvaluator {
+                current_z: self.z,
+                z: self.z,
+                result: F::ZERO,
+                one: self.one,
+            };
+            let _ = Ro::Output::map_gadget(&output, &mut acc)?;
+            acc.result
```
alxiong, line 628: i believe we don't need extra `z` to encode positional value of the output wires of subroutines, instead just use the x-power `WireEval` used in calculating shallow fingerprint, (which encodes the relative position of their allocation), hash their actual `x_0/1/2/3`-powered value into the `deep_hash`, that should suffice.

alxiong (dismissed):

alxiong (commented):
ACK. I pushed the last few commits so that `metrics.rs` is what I had in mind, and aligns with my book entry in #669 

(all of my last round review comments are resolved)

crates/ragu_circuits/src/metrics.rs
TalDerei: this is a good localized change (invariant is auditable in one place).

crates/ragu_circuits/src/metrics.rs
TalDerei: yeah this was an interstitial defense in depth measure, think i mentally also flagged removing it as some point.

crates/ragu_circuits/src/metrics.rs
TalDerei: verified that's equivalent, don't need to materialize an explicit SZ output scalar there. we're hashing more bytes, which is probably negligible compared to hashing a single field element (that's been accumulated).

crates/ragu_circuits/src/metrics.rs
TalDerei: ACK per previous observation.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -65,61 +74,172 @@ use super::{
...
+    fn new<F: PrimeField, Ro: Routine<F>>(
+        base: BaseFingerprint,
+        output_wires: &[F],
+        children: &[u64],
+    ) -> Self {
+        let mut state = blake2b_simd::Params::new()
+            .personal(DEEP_HASH_PERSONAL)
+            .to_state();
+        state.update(&type_id_u64(TypeId::of::<Ro::Input>()).to_le_bytes());
+        state.update(&type_id_u64(TypeId::of::<Ro::Output>()).to_le_bytes());
```
TalDerei: should we hash the raw 128-bits of each type id instead of truncating to 64-bits first?

TalDerei:
> ACK. I pushed the last few commits so that `metrics.rs` is what I had in mind, and aligns with my book entry in #669

reviewed @alxiong's  latest batch of cleanup commits against the [book entry](https://github.com/tachyon-zcash/ragu/pull/669), strictly an improvement (think we can lift the requested changes here)!

crates/ragu_circuits/src/metrics.rs
alxiong: good point, yes we could. 
(I will wait for Sean's review and address these in batch)

TalDerei:
@alxiong optimistically pushed some commits based on discussions in https://github.com/tachyon-zcash/ragu/pull/669.

crates/ragu_core/src/gadgets.rs
```diff
@@ -209,7 +209,7 @@ pub trait Gadget<'dr, D: Driver<'dr>>: Clone {
 /// trait is `unsafe` to implement and the implementor must ensure that this
 /// property holds. The [`Gadget`](derive@Gadget) derive macro ensures that this
 /// is the case.
-pub unsafe trait GadgetKind<F: Field>: core::any::Any {
```
TalDerei, line 253: drops `TypeId` field on deep hash and relaxes `Any` trait bound on `GadgetKind` (keeps it as `'static`).

alxiong:
ACK on the last three commits. They do align with what I have in mind.

Will give another pass when discussions in 669 got settled.

ebfull (approved):

PR #560, opened 2026-03-11
Macro cleanup

alxiong:
A cleanup for #499.
(basically my belated review comment on it, since it was merged over the weekend and don't have a chance to review back then)

- removed `unified_coverage_cover!` and `unified_coverage_assert_complete!`, turn them into `Coverage::cover()` and `Coverage::assert_covered()` plain methods instead.
  - previously their first matched pattern `field_type` is unused, but now they are pruned altogether 
- add some clarification on the `T, W` type param on `Slot` struct, explicit about what count as "covered"


TalDerei (approved):
looks good to me

TalDerei (approved):

PR #564, opened 2026-03-12
feat(circuits): expose free c wire in circuit api

TalDerei:
Paired allocation already computes `c = a * b` when two values share a gate, but the c wire handle was discarded at the circuit API boundary. `Element::alloc_mul` primitive exposes all three wires from a single mul gate (`c` wire now returns). For consecutive allocations that _then_ need to be multiplied, this collapses 2 allocs (1 gate / pairing) + 1 mul (1 gate + 2 LCs) = 2 gates + 2 LCs to 1 gate + 0 LCs. This only saves a few multiplication gates in our internal recursion circuits, but application circuits can lean into this more heavily.

### DIFF
--- a/crates/ragu_primitives/src/element.rs
+++ b/crates/ragu_primitives/src/element.rs
@@ -74,7 +74,7 @@ impl<'dr, D: Driver<'dr>> Element<'dr, D> {
         })
     }
 
-    /// Allocates two elements and returns them along with their product.
+    /// Allocates two fresh elements and returns them along with their product.
     ///
     /// This costs one multiplication constraint but produces three usable
     /// elements. The product comes from the free c wire of the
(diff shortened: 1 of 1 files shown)


crates/ragu_pcd/src/circuits/native/compute_v.rs
```diff
@@ -181,14 +181,24 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
...
+                let (mu_prime, nu_prime, mu_prime_nu_prime) = Element::alloc_mul(
+                    dr,
+                    unified_output.mu_prime.instance(),
+                    unified_output.nu_prime.instance(),
+                )?;
+                unified_output.mu_prime.fill(mu_prime.clone());
+                unified_output.nu_prime.fill(nu_prime.clone());
                 let mu_inv = mu.invert(dr)?;
                 let mu_prime_inv = mu_prime.invert(dr)?;
-                let munu = mu.mul(dr, &nu)?;
```
TalDerei: `mu * nu` was recomputed immediately after allocating mu and nu.

PR #566, opened 2026-03-13
chore(pcd): k(y) / claim builder alignment and registration assertions

TalDerei:
Rather than just checking circuit domain registration, this enforces a stricter check (both in standalone verifier and in-circuit) that the circuit was registered at that specific domain slot, ruling out any padding-slot circuit ID (where `padding slots = domain_size - num_registered_circuits`). This may be overly defense-in-depth, but intuitively seems better. 

This also tightens assertions between components. For instance, the claim builder and ky can silently go out of sync where if either `build` / `ky_values` is updated without the other, and extra claims get matched against zeros instead of their real `k(y)`, causing verification to fail _opaquely_. The new assertion catches this divergence earlier and more transparently.

### DIFF
--- a/crates/ragu_pcd/src/internal/tests.rs
+++ b/crates/ragu_pcd/src/internal/tests.rs
@@ -203,7 +203,7 @@ fn test_native_registry_digest() {
         .finalize(pasta)
         .unwrap();
 
-    let expected = fp!(0x0ce6a164b606bae6f89ade834fe5ebb1e496faf14891e5ab2d35bf40d856a00f);
+    let expected = fp!(0x0910d36955bb611c18d288a10325c3059446eb04ea6679f55176f6756fb4cf66);
 
     assert_eq!(
         app.native_registry.digest(),


crates/ragu_pcd/src/components/claims/mod.rs
```diff
@@ -121,6 +124,7 @@ impl<'m, 'rx, F: PrimeField, R: Rank> Builder<'m, 'rx, F, R> {
 
         self.a.push(a);
         self.b.push(Cow::Owned(sy));
+        self.num_stages += 1;
```
TalDerei: builder now tracks `num_stages` as claims are added via `stage_impl`.

crates/ragu_pcd/src/components/claims/native.rs
```diff
@@ -285,11 +285,32 @@ pub trait KySource {
     fn zero(&self) -> Self::Ky;
 }
 
+/// Returns the number of concrete (non-zero) k(y) values produced by
+/// [`ky_values`], i.e. the count before the infinite zero tail begins.
+///
+/// This must equal the number of non-stage claims produced by [`build`]
+/// for the same source shape. Use this in the verification path to assert
+/// alignment between `ky_values` and `build`.
+pub fn num_concrete_ky<S: KySource>(source: &S) -> usize {
```
TalDerei: returns the count of non-zero k(y) values.

PR #568, opened 2026-03-14
chore: update readme and default multicore feature

TalDerei:
Inspired by https://github.com/zcash/halo2, flushes the readme with MSRV and parallelism sections. Also makes `multicore` the default feature.


alxiong (changes requested):

alxiong (approved):

ebfull (changes requested):
Some of the crates for which this `alloc` feature was added did not have `compiler_error`s thrown into their `lib.rs`.

PR #569, opened 2026-03-14
chore(pcd): ZST markers for typed FS challenges

TalDerei:
Another stab at https://github.com/tachyon-zcash/ragu/pull/521 (the main source of inspiration for this work was Halo2's challenge handling, and _sparsely_ references the circuit crate recommendation from https://github.com/QED-it/ragu/issues/53). The previous impl attempt conflated two orthogonal concerns: (1) protocol identity (a value came from fiat-shamir) and (2) domain restriction (a value must be non-zero).

This approach instead follow's Halo2's phantom-type pattern to distinguish particular challenges by type _without_ imposing the aforementioned domain constraints. The motivation is to prevent subtle bugs from challenge misuse.

```ts
  struct Challenge<F, T> {
      inner: F,
      _marker: PhantomData<T>,
  }
  impl Deref for Challenge<F, T> { type Target = F; ... }
```

- ZST markers (`ChallengeW`, `ChallengeX`, etc.),
- Transcript squeezes typed challenges via `challenge::<ZST>()`,
- Fuse signatures use typed challenges,
- Deref internally converts to raw `Element` / `F`


crates/ragu_pcd/src/proof/components.rs
```diff
@@ -12,6 +15,86 @@ use ragu_primitives::Element;
...
+        &self.inner
+    }
+}
+
+// ZST markers for each challenge role in the protocol. These are phantom type
+// parameters only — never instantiated as values.
+
+/// Registry evaluation point $w$, squeezed after the preamble commitment.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct ChallengeW;
```
TalDerei: zero-sized type (ZST) which adds a _compile_-time tag.

crates/ragu_pcd/src/circuits/native/hashes_1.rs
```diff
@@ -234,16 +235,16 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
...
         unified_output.w.set(w.clone());
 
         // Derive (y, z) by absorbing nested_s_prime_commitment and squeezing twice
         let (y, z) = {
             let nested_s_prime_commitment = unified_output.nested_s_prime_commitment.verify(dr)?;
             nested_s_prime_commitment.write(dr, &mut transcript)?;
-            let y = transcript.challenge(dr)?;
-            let z = transcript.challenge(dr)?;
+            let y = transcript.challenge::<ChallengeY>(dr)?.into_inner();
+            let z = transcript.challenge::<ChallengeZ>(dr)?.into_inner();
```
TalDerei: Imposes stricter API constraints for type safety.

alxiong (commented):
I'm not convinced all these verbosity for the marginal benefit of avoiding accidental challenge misuse. 

my interpretation of QEDIT's main concern was the potential zero-value challenge. not really mismatch or misuse. these are all internal logic, sutble bug should be caught by existing unit tests? (if I feed `mu` to `alpha`, then the verification would immediately fail, right?)

ebfull:
Going to defer to @alxiong on this PR. It does seem like something that would maybe help as a later refactoring once the protocol is feature-complete though...

TalDerei:
update: will keep this as a draft for now and rebase when appropriate, but this would be nice to eventually land.

TalDerei:
closing for now; filed https://github.com/tachyon-zcash/ragu/issues/822 for tracking this kind of change.

PR #574, opened 2026-03-15
chore(circuits): exploit ILP for revdot 

TalDerei:
Revdot had a single serial accumulator (chained iterator) across all four vector pairs. Now, hardware scheduler can more _effectively_ exploit ILP by seeing 4 independent multiply–accumulate (MAC) chains and interleaving their instructions as separate executions. Scheduler probably did this before to some lesser degree, but this provides stronger hints.

### DIFF
--- a/crates/ragu_circuits/src/polynomials/structured.rs
+++ b/crates/ragu_circuits/src/polynomials/structured.rs
@@ -122,13 +122,16 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
 
     /// Inner product of `self` with the reversed `other`.
     pub fn revdot(&self, other: &Self) -> F {
-        self.u
-            .iter()
-            .zip(other.v.iter())
-            .chain(self.v.iter().zip(other.u.iter()))
-            .chain(self.w.iter().zip(other.d.iter()))
-            .chain(self.d.iter().zip(other.w.iter()))
-            .fold(F::ZERO, |acc, (a, b)| acc + (*a * *b))
+        fn dot<F: Field>(a: &[F], b: &[F]) -> F {
+            a.iter()
+                .zip(b.iter())
+                .fold(F::ZERO, |acc, (a, b)| acc + (*a * *b))
+        }
+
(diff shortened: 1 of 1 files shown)


crates/ragu_circuits/src/polynomials/structured.rs
```diff
@@ -122,13 +122,16 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
...
-            .chain(self.w.iter().zip(other.d.iter()))
-            .chain(self.d.iter().zip(other.w.iter()))
-            .fold(F::ZERO, |acc, (a, b)| acc + (*a * *b))
+        fn dot<F: Field>(a: &[F], b: &[F]) -> F {
+            a.iter()
+                .zip(b.iter())
+                .fold(F::ZERO, |acc, (a, b)| acc + (*a * *b))
+        }
+
+        dot(&self.u, &other.v)
```
TalDerei: `dot` is a multiply-add (MAD, not MAC) chain that itself is also a viable target. there's an observation that we can treat each dot call as 2 independent accumulators (`acc0`, `acc1`).

```ts
fn dot<F: Field>(a: &[F], b: &[F]) -> F {
    let mut accs = [F::ZERO; 2];
    for (i, (a, b)) in a.iter().zip(b.iter()).enumerate() {
        accs[i & 1] += *a * *b;
    }
    accs[0] + accs[1]
}
```

ebfull:
There's another optimization that's worth tracking somewhere: multiplications in our fields perform reductions immediately after computing the product, and the reductions cost about as much as the multiplications themselves. But the reductions are only needed to recover the final element in Montgomery form, and the unreduced forms are algebraically related (up to residue class or something) and so can be summed. In other words, we should be able to perform inner products without reducing until the very end, due to the low multiplicative depth of inner products.

TalDerei:
@alxiong haven't yet reviewed the PR, but since it's accompanied by a more extensive `pasta_curves` diff, merging this in the interim and we can iterate.

PR #578, opened 2026-03-16
A commitment optimization

ebfull:
This entire PR was written by Claude.

----

During fuse, `a_commitment` is computed by committing to the fully-assembled $A(X)$ polynomial — an MSM of size $O(\text{degree})$. But $A(X)$ is just a linear combination of polynomials already committed in the child proofs. Naively we could track each claim's contribution independently, but many claims reference the same underlying child-proof polynomial (e.g. both the circuit claim and a stage claim for the same circuit share an rx polynomial). So we track the decomposition as `(source_key, coefficient)` pairs keyed by `(Side, RxComponent)`, thread that tracking through the entire claims pipeline and both fold layers, then deduplicate by key at the end — summing coefficients for terms that share a commitment.

The final MSM size is $O(\text{commitments})$, not $O(\text{claims})$ or $O(\text{degree})$. $B(X)$ does not qualify for this optimization because `circuit_impl` dilates the rx polynomial by $z$ before adding $s(y)$ and $t(z)$ — the result is not a linear combination of anything already committed.

The claims pipeline assembles $A(X)$ from child-proof rx polynomials through three operations — direct inclusion, polynomial sums, and Horner folds — then applies two layers of Horner folding ($M$, then $N$) to reduce to a single polynomial. Each constituent $a_i$ decomposes over child-proof sources:

| Claim type | Decomposition |
|---|---|
| raw / circuit | $a_i = r_j$ (single source, coefficient 1) |
| internal circuit | $a_i = \sum_j r_j$ (sum, all coefficients 1) |
| stage | $a_i = \sum_j z^{n-1-j} \cdot r_j$ (Horner fold over $z$) |

Horner folding is linear, so the decomposition is preserved through both layers. Since `commit` is also linear:

$$\text{commit}(A, b) = \sum_j c_j \cdot \text{commit}(r_j, b_j) + \left(b - \sum_j c_j \cdot b_j\right) \cdot H$$

After deduplication, this is an MSM of size $|\text{distinct keys}| + 1$ — bounded by the number of distinct `(Side, RxComponent)` pairs across both child proofs, regardless of how many claims or fold steps reference them.

### Implementation

The main obstacles are that `fold_polys_m`/`fold_polys_n` only know about polynomials, `Builder` hardcodes `Cow<Polynomial>` for the a-vector, and there's no way to resolve a source identity back to its `(commitment, blind)` pair. The commits build toward the optimization one layer at a time:

1. **`Foldable` trait** — Abstracts the two Horner-fold operations (`scale`, `add_assign`) so the fold functions accept any foldable type, not just `structured::Polynomial`.

```diff
-pub fn fold_polys_m<F: Field, R: Rank, P: Parameters>(
-    source: &[impl Borrow<structured::Polynomial<F, R>>],
+pub fn fold_polys_m<T: Foldable<F>, F: Field, P: Parameters>(
+    source: &[impl Borrow<T>],
```

2. **Extract `FuseProofSource`** — Moves the fuse source into `fuse/claims.rs` and hoists construction to the caller, preparing for atom-tagged accessors.

3. **Generic `Builder<A>`** — Parameterizes `Builder` over `A` so the a-vector can hold either `Cow<Polynomial>` (verify/nested) or `TrackedPoly` (fuse). Splits `stage_impl` into `fold_stage_polys` + `stage_impl` so callers control what gets pushed.

```diff
-pub struct Builder<'m, 'rx, F: PrimeField, R: Rank> {
-    pub a: Vec<Cow<'rx, structured::Polynomial<F, R>>>,
+pub struct Builder<'m, 'rx, A, F: PrimeField, R: Rank> {
+    pub a: Vec<A>,
```

4. **Decomposition tracking** — Introduces `TrackedPoly<K, F, R>` (polynomial + `CommitmentDecomposition<K, F>`) with a `Foldable` impl that folds both in lockstep. Adds `FuseAtom = (Side, RxComponent)` keys and `rx_commitment_blind` on `Proof` for resolution.

5. **The payoff** — `_06_ab` destructures the final `TrackedPoly`, deduplicates terms by key, and computes `a_commitment` via small MSM:

```diff
-        let [a_commitment, b_commitment] = ragu_arithmetic::batch_to_affine([
-            a_poly.commit(host_gen, a_blind),
-            b_poly.commit(host_gen, b_blind),
-        ]);
+        let a_commitment_proj = {
+            // deduplicate decomposition terms by FuseAtom key, summing coefficients
+            ...
+            for (key, coeff) in entries {
+                let (commitment, blind) = source.resolve_atom(key);
+                accumulated_blind += coeff * blind;
+                msm.push((coeff, commitment));
+            }
+            msm.push((a_blind - accumulated_blind, *host_gen.h()));
+            ragu_arithmetic::mul(...)
+        };
```

6. **Cleanup** — Minor improvements from code review.

### DIFF
--- a/crates/ragu_pcd/src/fuse/claims.rs
+++ b/crates/ragu_pcd/src/fuse/claims.rs
@@ -126,9 +126,9 @@ impl<K, F: Field, R: Rank> Borrow<structured::Polynomial<F, R>> for TrackedPoly<
 ///
 /// The fuse pipeline threads these through the claims builder so that each
 /// `A` polynomial retains a link back to its commitment.
-pub struct Atom<'rx, K, F: Field, R: Rank> {
-    pub key: K,
-    pub poly: &'rx structured::Polynomial<F, R>,
+pub(super) struct Atom<'rx, K, F: Field, R: Rank> {
+    pub(super) key: K,
+    pub(super) poly: &'rx structured::Polynomial<F, R>,
 }
 
 // Manual Copy/Clone: derive(Copy) would add spurious F: Copy and R: Copy bounds.
@@ -142,21 +142,21 @@ impl<K: Copy, F: Field, R: Rank> Copy for Atom<'_, K, F, R> {}
 
 /// Identifies which of the two child proofs a polynomial came from.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum Side {
+pub(super) enum Side {
     Left,
     Right,
 }
 
 /// Key identifying a polynomial and its corresponding commitment within the
 /// fuse pipeline: which child proof, and which component of that proof.
-pub(crate) type FuseAtom = (Side, RxComponent);
+pub(super) type FuseAtom = (Side, RxComponent);
 
 /// The two child proofs being fused. Provides [`Atom`]-tagged rx values
 /// for claim building, and resolves [`FuseAtom`] keys back to their
 /// `(commitment, blind)` pairs for the MSM in `_06_ab`.
-pub(crate) struct FuseProofSource<'rx, C: Cycle, R: Rank> {
-    pub(crate) left: &'rx Proof<C, R>,
-    pub(crate) right: &'rx Proof<C, R>,
+pub(super) struct FuseProofSource<'rx, C: Cycle, R: Rank> {
+    pub(super) left: &'rx Proof<C, R>,
+    pub(super) right: &'rx Proof<C, R>,
 }
 
 impl<'rx, C: Cycle, R: Rank> FuseProofSource<'rx, C, R> {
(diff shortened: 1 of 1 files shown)


crates/ragu_pcd/src/fuse/_06_ab.rs
```diff
@@ -87,16 +93,49 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
+
+            let mut accumulated_blind = C::CircuitField::ZERO;
+            let mut msm: Vec<(C::CircuitField, C::HostCurve)> =
+                Vec::with_capacity(entries.len() + 1);
+            for (key, coeff) in entries {
+                let (commitment, blind) = source.resolve_atom(key);
+                accumulated_blind += coeff * blind;
+                msm.push((coeff, commitment));
+            }
+            msm.push((a_blind - accumulated_blind, *host_gen.h()));
```
TalDerei: ACK – blinding term correction.

crates/ragu_pcd/src/proof/components.rs
```diff
@@ -17,14 +17,28 @@ use rand::CryptoRng;
 
 use alloc::vec::Vec;
 
+/// A grouped `(rx, blind, commitment)` triple for a native-field rx polynomial.
+///
+/// The `commitment` is the Pedersen commitment to `rx` under blinding scalar
+/// `blind`.
+#[derive(Clone)]
+pub struct RxTriple<C: Cycle, R: Rank> {
```
TalDerei:
```suggestion
#[derive(Clone)]
pub(crate) struct RxTriple<C: Cycle, R: Rank> {

PR #575, opened 2026-03-15
pcd refactoring: make circuit handling exhaustive in native code, other reorganizations and cleanups

ebfull:
This branch reorganizes `ragu_pcd` internals around two exhaustive enums — `InternalCircuitIndex` and `RxIndex` — with canonical `ALL` arrays that drive circuit registration, proof field access, and the fuse pipeline. Paired container structs (`InternalCircuitValues<T>`, `RxValues<T>`) replace scattered manual indexing, and the old `circuits/`/`components/` modules merge into a single `internal/` namespace. The result: adding or modifying an internal circuit now produces compile errors everywhere it needs handling, making the dual-pipeline code significantly cleaner.

Lots of other reorgs and cleanups and refactorings here as well, all intended to make the nested mirror of the protocol behavior simple to incrementally review.

PR #582, opened 2026-03-16
chore(pcd): refactoring suggestions

TalDerei:
Suggestions for https://github.com/tachyon-zcash/ragu/pull/575 for reducing some verbosity on call sites using the existing patterns that PR established.

### DIFF
--- a/crates/ragu_pcd/src/internal/native/mod.rs
+++ b/crates/ragu_pcd/src/internal/native/mod.rs
@@ -28,8 +28,7 @@ pub mod unified;
 
 /// Identifies one of the 13 internal circuits registered by [`register_all`].
 ///
-/// Variants are grouped into native circuits, native stage masks, and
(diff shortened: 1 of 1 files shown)


crates/ragu_pcd/src/internal/native/mod.rs
```diff
@@ -290,6 +290,77 @@ pub enum RxComponent {
...
+    Hashes1,
+    Hashes2,
+    PartialCollapse,
+    FullCollapse,
+    ComputeV,
+}
+
+const NUM_CIRCUIT_PROOFS: usize = 5;
+
+impl CircuitProofIndex {
```
TalDerei: `CircuitProofIndex / CircuitProofValues` is an optimistic impl, references https://github.com/tachyon-zcash/ragu/pull/575#discussion_r2937290966.

PR #584, opened 2026-03-17
Single responsibility claim order

alxiong:
Here's my stab at #566. (only the ky/claimer part, since that PR contains changes for orthogonal issue like `circuit.is_registered()` check)

Instead of tracking `num_stages` to assert ky value alignment, I'd like a centrally declared claim order that is singularly responsible for the order in both `build()` and `ky_values()`. 

### Problem Statement

`build()` and `ky_values()` must iterate claims in the same order. Previously they were independent functions that implicitly agreed; if they drifted, verification would fail silently. Now both are driven by `claim_order(),` making drift impossible.

### Changes Made

- Add `ClaimOrder` + `KyKind` enums (native) and `ClaimOrder` enum (nested) that define canonical claim iteration order
- Add `KyValues<I>` structs that flatten per-group iterators in `claim_order()` sequence via `into_values()`
- Add shared `KySource` trait `(internal/claims.rs)` with `ky_values()` and `padded_ky_values()` provided method
- Rewrite `build()` in both native and nested to iterate `claim_order()`
- Delete `NUM_UNIFIED_CIRCUITS` constant

### DIFF
--- a/crates/ragu_pcd/src/fuse/_05_error_n.rs
+++ b/crates/ragu_pcd/src/fuse/_05_error_n.rs
@@ -23,6 +23,7 @@ use rand::CryptoRng;
 use crate::{
     Application,
     internal::{
+        claims::KySource,
         fold_revdot, native,
         native::stages::error_n::{ChildKyValues, KyValues},
         nested,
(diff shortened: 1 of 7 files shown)


TalDerei:
advocating for closing this; rounds of refactors (#632 for instance) will probably make this moot, and the worst case from the divergence this PR sets out to resolve is proof failure.

alxiong:
agree, at this stage, my changes here feels superfluous. Sean's `InternalCircuitIndex::ALL` effort already eliminates lots of the concern. 
Even though we still have a `native::claims::build()` and `ky_values()` split that might diverge, inspecting/spotting such drift is significantly easier thanks to the explicit and centralized order in `ALL`.

PR #585, opened 2026-03-18
feat(pcd): `#[application]` and `#[header]` macros for app definitions

alxiong:
Closes #330

Adds the `#[application]` and `#[header]` proc-macros and a `ragu_pcd::app` module: a higher-level API for defining PCD applications.

### Why

Writing a PCD app today means hand-managing framework bookkeeping — a unique `Suffix` per header, an `INDEX` per step, and a full `Header`/`Step` impl for each.

### How

`#[application]` reads the step list at compile time and derives suffixes and indices from declaration order — no `TypeId` lookups, no runtime registry, no `'static` bound on step types. It generates the `Header`/`Step` impls plus a wrapper with typed `build()`/`seed()`/`fuse()`/`verify()`/`rerandomize()` methods sealed to the declared steps.

`#[header]` generates the whole `HeaderContent` impl for the common case: a header that allocates a single gadget from its data. A custom `encode()` is still available when you need one.

The raw `ragu_pcd` API is untouched; `app` is a thin layer on top, so framework internals can evolve independently of the developer-facing surface.

### Also in here

- `Suffix::new` now rejects values that would wrap into the reserved internal suffix range.
- `Header` and `Pcd::data` document that only what `encode()` constrains is authenticated — a non-injective encoding leaves the rest as unauthenticated witness data.
- `crates/ragu_pcd/tests/example.rs`: an end-to-end three-step app (hash leaves → merge → endoscalar group scale).

PR #586, opened 2026-03-18
Header macro

alxiong:
A follow up of #585, (blocked by it)
(Only review [8d9bb19](https://github.com/tachyon-zcash/ragu/pull/586/commits/8d9bb196e04a09994ded5d0a5940af386be2489e) onward, everything prior are from 585)

This PR introduce a new `#[header]` proc-macro that auto generate the `Header` implementation for trivial header (that directly allocate everything in the `H::Data` into its `H::Output` gadget. We still leave the freedom to circuit dev to define their own custom `Header::encode()` logic if needed.

### DIFF
--- a/crates/ragu_macros/src/proc/header.rs
+++ b/crates/ragu_macros/src/proc/header.rs
@@ -1,0 +1,217 @@
+//! Implementation of the `#[header]` proc-macro.
+//!
+//! Generates a `HeaderContent` implementation for single-gadget headers where
+//! `encode()` calls `Gadget::alloc(dr, witness)`.
+//!
+//! # Modes
+//!
+//! - **Generic** (`#[header(data = F, gadget = Element)]`): `data` doubles as
+//!   the field type parameter, producing `impl<F: Field> HeaderContent<F>`.
+//! - **Concrete** (`#[header(data = EpAffine, gadget = Point<EpAffine>, field = Fp)]`):
+//!   an explicit `field` pins the impl to a specific field type.
+
+use proc_macro2::TokenStream;
+use quote::quote;
+use syn::{Error, Ident, ItemStruct, Result, Token, Type, parse::Parse, parse::ParseStream};
+
+use crate::path_resolution::RaguAppPath;
+
+/// Generates the `HeaderContent` impl for a `#[header]`-annotated struct.
+pub fn evaluate(attr: HeaderAttr, item: ItemStruct) -> Result<TokenStream> {
+    let app = RaguAppPath::resolve()?;
+
+    if !item.generics.params.is_empty() || item.generics.where_clause.is_some() {
+        return Err(Error::new_spanned(
+            &item.generics,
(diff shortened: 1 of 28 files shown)


TalDerei (commented):
ACK – proc macro for trivial headers

crates/ragu_macros/src/proc/header.rs
```diff
@@ -0,0 +1,217 @@
...
+    let struct_ident = &item.ident;
+    let struct_attrs = &item.attrs;
+    let data_ty = &attr.data;
+    let prelude = quote!(#app::__macro_internal);
+    let kind_expr = make_kind_expr(&attr.gadget)?;
+    let gadget_base = extract_base_path(&attr.gadget)?;
+
+    // When `field` is absent, `data` is both the field parameter and the data
+    // type, yielding a generic impl. When present, the impl is concrete.
+    let field_ty = attr.field.as_ref().unwrap_or(data_ty);
```
LucidSamuel, line 47: in generic mode (no `field`), `data` is used directly as the type parameter name. so If someone writes:
  `#[header(data = Vec<u8>, gadget = Element)]`

  this tries to generate `impl<Vec<u8>: Field> ...`, which is invalid.
  could validate that data is a simple ident when field is absent:

```
 if attr.field.is_none() {
      match &attr.data {
          Type::Path(p)
              if p.qself.is_none()
                  && p.path.segments.len() == 1
                  && p.path.segments[0].arguments ==
  syn::PathArguments::None => {}
          _ => {
              return Err(Error::new_spanned(
                  &attr.data,
                  "when `field` is not specified,
  `data` must be a type parameter name (e.g. `F`). For
  concrete data types, add `field = YourField`",
              ));
          }
      }
  }
```

crates/ragu_macros/src/proc/header.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// Returns the inner [`syn::TypePath`], or an error if `ty` is not a path type.
+fn as_type_path(ty: &Type) -> Result<&syn::TypePath> {
+    match ty {
+        Type::Path(p) => Ok(p),
+        _ => Err(Error::new_spanned(ty, "expected a path type for gadget")),
+    }
+}
+
+/// Builds the `Kind!` gadget expression by prepending `'_, _` (the lifetime
```
LucidSamuel, line 162: would be worth adding a short user-facing note on
  `#[header]` / `HeaderContent` saying that custom gadgets
  need to follow the `<'dr, D: Driver, ...>` convention to work with this macro.

LucidSamuel (commented):
GG, suggested minor improv

TalDerei:
@alxiong cherry-picked commit into https://github.com/tachyon-zcash/ragu/pull/585 so there's one reviewable set of changes for macros, and modified PR title and description to reflect these header macro changes accordingly. thought this is the right direction to start getting things merged and reduce PR queue pressure.

PR #580, opened 2026-03-16
perf: Use lazy Montgomery reduction during revdot

alxiong:
Close #579, Pair with https://github.com/ebfull/pasta_curves/pull/1.
subsume changes in ILP improvement in #574 (cc @TalDerei, but feel free to merge that in, it's originally orthogonal improvements, but it's clearer for me to just follow the same inspiration here and break up into 4 `n`-sized revdot)

The benchmark verifies **a 47% improvement on `revdot`** against `main`:

```
circuits::poly_ops::revdot revdot:setup_rng((rand_structured_poly, rand_structured_p...
  Instructions:                     1093515|2075253              (-47.3069%) [-1.89778x]
  L1 Hits:                          1315245|2395959              (-45.1057%) [-1.82168x]
  LL Hits:                             8256|8243                 (+0.15771%) [+1.00158x]
  RAM Hits:                              63|80                   (-21.2500%) [-1.26984x]
  Total read+write:                 1323564|2404282              (-44.9497%) [-1.81652x]
  Estimated Cycles:                 1358730|2439974              (-44.3138%) [-1.79578x]

```

PR #589, opened 2026-03-18
chore: expanding lazy mont reduction

TalDerei:
Builds on https://github.com/tachyon-zcash/ragu/pull/580 by extending `inner_product` usage at callsites. The first commit duplicates that diff and will be subsumed after rebase once @alxiong's PR lands. CI currently failing because of the `needs_reduction` bug identified in https://github.com/zcash/pasta_curves/pull/88.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -604,7 +604,7 @@ checksum = "d6790f58c7ff633d8771f42965289203411a5e5c68388703c06e14f24770b41e"
 [[package]]
 name = "pasta_curves"
 version = "0.5.1"
-source = "git+https://github.com/ebfull/pasta_curves?branch=ff-0.14-with-rand-0.10#ad4c7a24d9b0829f9da5f305aab6b4ab97a97900"
+source = "git+https://github.com/alxiong/pasta_curves?branch=v0.5.2-rc.1#59e30c5377191c6ef10f5051c3d95da43eb1081e"
 dependencies = [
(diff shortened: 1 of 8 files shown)


alxiong:
Re-reviewed, since i just updated #580, as claude summaries the changes here:

- Plumbing / trait bounds / `F::inner_product` usage — already subsumed by the `lazy-mont-reduction` commit, and on a cleaner API `(DeferredField` vs the now-removed `MontgomeryRepr).`
- `util::dot` rewrite — skipped. Only used in tests/bench, not on the hot path. If it ever becomes hot, the better form is zip-into-`mul_accumulate` (no Vec), not the PR's materialize-to-Vec pattern.
- ILP-unrolled nested `Horner` in `_08_f.rs`. Loop-order swap (iterators outer, slots inner) + 4-wide destructure per chunk exposes independent mul-adds across slots, orthogonal to lazy reduction (the α·acc+next pattern must reduce between steps anyway).

Only the 3rd point remains to be valid consideration, first two are either subsumed or irrelevant. 
I personally find the ILP-unrolling in horner bringing too little benefits to justify the unrolled logic and various `.unwrap()`. 
empirically, I tried to copy that ILP verbatim on my local branch, and criterion says there is -0.1445% change in `fuse`, which is negligible and within noise. 

I'd suggest we close this PR unless @TalDerei thinks we should port some ideas (that I missed) over?

Issue #593, opened 2026-03-19
Generalize stage mask infrastructure to "bonding polynomials"


ebfull:
## Motivation

Routing polynomials — wiring polynomials that enforce value equality between wire positions in different stages — are needed to connect stages that share data. The current infrastructure would treat these as ordinary circuits, requiring witnesses for every stage they touch. This is the core problem described in #592: witness data for different stages may not all be available at the same time, and forcing circuits to fully witness every stage they reference is too coarse.

But routing polynomials only enforce linear constraints — they don't multiply or allocate wires. Any circuit with only linear constraints has no witness and no need for a trace polynomial of its own. Its wiring polynomial is checked against traces produced by other circuits or stages, and because it encodes only linear constraints, the revdot check is linear in those traces — so multiple traces fold into one claim. This means routing circuits don't need witnesses at all, which eliminates the witness availability problem that #592's composite staging system was designed to solve.

Stage masks are already an instance of this same pattern — linear-only circuits with no witness, whose wiring polynomials are batchable across traces. But the infrastructure supporting them (`internal_masks`, `register_internal_mask`, `Processor::stage()`, `stage_impl`, `fold_stage_polys`) is named and framed as stage-mask-specific.

## Proposal

Introduce the concept of a **bonding polynomial**: a wiring polynomial encoding only linear constraints, with no witness and no need for a trace polynomial of its own.[^1] Its revdot check is linear in the traces it constrains, making it batchable. Refactor the existing stage mask infrastructure to use this general framing. Stage masks and routing polynomials become different factories producing the same kind of object, processed through the same claim machinery.

[^1]: Bonding polynomials are also distinguished from circuit polynomials by having a zero constant term, which prevents them from being substituted for a circuit polynomial that anchors the ONE wire.

ebfull:
These special pattern circuits that only can run `enforce_zero` do not need witnesses at all.

TalDerei:
closed by https://github.com/tachyon-zcash/ragu/pull/596, https://github.com/tachyon-zcash/ragu/pull/602, https://github.com/tachyon-zcash/ragu/pull/610, https://github.com/tachyon-zcash/ragu/pull/600.

PR #594, opened 2026-03-19
chore(circuits): Introduce bonding polynomials

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/593 and to be rebased over https://github.com/tachyon-zcash/ragu/pull/596. Introduces a general-purpose bonding polynomial factory alongside the existing stage mask component. Downstream consumers like routing circuits will impl `BondingCircuit` that basically provides a path for arbitrary linear constraints that goes through synthesis (ie. for enforcing equality between wires in different stages across fuse steps).

### DIFF
--- a/crates/ragu_circuits/src/staging/bonding.rs
+++ b/crates/ragu_circuits/src/staging/bonding.rs
@@ -62,7 +62,7 @@ impl<'a, 'dr, D: Driver<'dr>> BondingDriver<'a, 'dr, D> {
 ///
 /// To reference a specific stage's wires, index into `gates` using the
 /// stage's [`skip_multiplications`](super::Stage::skip_multiplications) offset.
-pub trait BondingCircuit<F: Field>: Sized + Send + Sync {
+pub trait BondingCircuit<F: Field, R: Rank>: Sized + Send + Sync {
     /// How many gates to pre-allocate wire handles for.
     fn num_gates(&self) -> usize;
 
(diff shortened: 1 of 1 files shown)


crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -0,0 +1,254 @@
...
+        gates: &[(D::Wire, D::Wire, D::Wire)],
+    ) -> Result<()>;
+
+    /// Produce a [`CircuitObject`] for registry registration via
+    /// [`register_internal_bonding`](registry::RegistryBuilder::register_internal_bonding).
+    fn into_bonding_object<'a, R: Rank>(self) -> Result<Box<dyn CircuitObject<F, R> + 'a>>
+    where
+        Self: 'a,
+        F: FromUniformBytes<64>,
+    {
```
TalDerei: `into_bonding_object` produces the `Box<dyn CircuitObject>`

```ts
registry.register_internal_bonding(X.into_bonding_object::<R>()?)?

File: crates/ragu_circuits/src/staging/bonding.rs
Comment (TalDerei) [inline crates/ragu_circuits/src/staging/bonding.rs]: not sure where this bonding factory live?

File: crates/ragu_circuits/src/staging/bonding.rs [hunk 2]
```diff
@@ -0,0 +1,305 @@
...
+            adapter,
+            metrics,
+        })))
+    }
+}
+
+/// Bridges [`BondingCircuit`] to [`Circuit`] so we can reuse the standard
+/// synthesis drivers. Pre-allocates gate wire handles, then hands the
+/// restricted [`BondingDriver`] to the bonding circuit's `witness`.
+struct Adapter<B>(B);
```
ebfull: Building on the above: if bonding polynomial synthesis uses a standard `Driver` (with a `mul`-intercepting decorator) and locates wires through the existing `MultiStage`/`StageBuilder` APIs, then `Adapter` and `Processed` can go away entirely. Currently they exist to bridge `BondingCircuit` back to `Circuit` so `s::*::eval` can be reused, but eval unconditionally injects the ONE wire — so `Processed` has to strip it back out in three independent, representation-dependent places (`sxy` subtracts `x^(4n-1)`, `sx` zeroes `coeffs[0]`, `sy` subtracts from `backward().c[0]`). Each depends on eval internals like where Horner places `enforce_one`. `StageMask` avoids this entirely by never producing the ONE wire contribution — correct by construction rather than by subtraction.

ebfull:
I think with #600 merged (or something equivalent) this PR will be an unnecessary generalization for now. Suggesting to close.

Issue #573, opened 2026-03-15
Avoiding fine-grained parallelism 


TalDerei:
In some cases, fine-grained parallelism is actually an _anti-pattern_ (found this Halo2 issue: https://github.com/zcash/halo2/issues/360) where thread overhead is greater than actual work. Spawning thread pools inside every polynomial operation, as that issue described, would slow things down.

Instruction-level parallelism / SIMD combined with coarse-grained parallelism (multi-core / rayon) are cases where parallelism actually speeds up proving and justifies the threading overhead.

TalDerei:
closed by https://github.com/tachyon-zcash/ragu/pull/595, cc @ebfull.

PR #595, opened 2026-03-19
chore(pcd): parallelize fuse evals

TalDerei:
Inspired by https://github.com/tachyon-zcash/ragu/issues/573, closes TODOs in our fuse pipeline by evaluating fuse polynomial evaluations in query and eval stages using _coarse_-grained parallelism. Unlike MSMs which use `maybe::into_par_iter` (which is for data-parallelism when you need to split a collection across threads), this implements `maybe::join` (which is for task parallelism since we're working with independent closures). It doesn't use `scope` which needs channels or shared state. This parallelizes:

- 13 independent poly evals at omega
- 6 independent poly evals at u
- Left/right child witness construction

------------

Benches against main suggest **~5%** improvement in fuse. 

```
git stash && cargo bench --bench fuse_criterion -p ragu_pcd --features multicore
git stash pop && cargo bench --bench fuse_criterion -p ragu_pcd --features multicore 
```

Screenshot 2026-03-19 at 4 51 20 PM

### DIFF
--- a/crates/ragu_arithmetic/src/lib.rs
+++ b/crates/ragu_arithmetic/src/lib.rs
@@ -83,6 +83,10 @@ pub use coeff::Coeff;
 pub use domain::Domain;
 use ff::{Field, FromUniformBytes, WithSmallOrderMulGroup};
 pub use fft::{Ring, bitreverse};
+/// Internal re-export consumed by [`par_join!`]. `par_join!` expands to a path
+/// through `$crate`, so the helper must be reachable at the crate root.
+#[doc(hidden)]
+pub use multicore::join;
 pub use pasta_curves::{
     arithmetic::{Coordinates, CurveAffine, CurveExt},
     deferred::DeferredField,
@@ -103,6 +107,21 @@ pub use u128 as Uendo;
 pub use util::{
     batch_to_affine, dot, eval, factor, factor_iter, geosum, low_u64, mul, poly_with_roots,
 };
+
+/// N-way parallel join for coarse-grained task parallelism.
+///
+/// Like rayon's `join` for more than two closures: nests internally and
+/// flattens the result into a single tuple. Each closure may return a
+/// different type. Supports 2..=4 closures — for higher arities prefer a
+/// data-parallel iterator.
+///
+/// ```
+/// use ragu_arithmetic::par_join;
+///
+/// let (a, b, c) = par_join!(|| 1u32, || "two", || 3.0_f64);
+/// assert_eq!((a, b, c), (1, "two", 3.0));
+/// ```
+pub use crate::__ragu_arithmetic_par_join as par_join;
 
 /// Represents a "cycle" of elliptic curves where the scalar field of one curve
 /// is the base field of the other, and vice-versa.
--- a/crates/ragu_arithmetic/src/lib.rs
+++ b/crates/ragu_arithmetic/src/lib.rs
@@ -107,20 +107,6 @@ pub use util::{
     poly_mul, poly_with_roots,
 };
 
-/// N-way parallel join for coarse-grained task parallelism.
-///
-/// Like [`join`] for more than two closures: nests internally and flattens the
-/// result into a single tuple. Each closure may return a different type.
-/// Supports 2..=4 closures — for higher arities prefer a data-parallel
-/// iterator.
-///
-/// ```
-/// use ragu_arithmetic::par_join;
-///
-/// let (a, b, c) = par_join!(|| 1u32, || "two", || 3.0_f64);
-/// assert_eq!((a, b, c), (1, "two", 3.0));
-/// ```
-pub use crate::__ragu_arithmetic_par_join as par_join;
 use crate::ff::{Field, FromUniformBytes, PrimeFieldBits, WithSmallOrderMulGroup};
 pub use crate::pasta_curves::{
     arithmetic::{Coordinates, CurveAffine, CurveExt},
--- a/crates/ragu_arithmetic/src/lib.rs
+++ b/crates/ragu_arithmetic/src/lib.rs
@@ -90,6 +90,13 @@ pub extern crate rand;
 pub use coeff::Coeff;
 pub use domain::Domain;
 pub use fft::{Ring, bitreverse};
+/// Runs two closures, potentially in parallel, and returns both of their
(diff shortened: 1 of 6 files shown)


crates/ragu_pcd/src/fuse/_07_query.rs
```diff
@@ -79,29 +79,47 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
+            || {
+                (
+                    native::InternalCircuitValues::from_fn(|id| {
+                        registry_xy_poly.eval(id.circuit_index().omega_j())
+                    }),
+                    registry_xy_poly.eval(w),
+                )
+            },
+            || {
+                maybe_rayon::join(
```
TalDerei: nested / outer joins was the idiomatic way to do this.

ebfull (changes requested):

alxiong:
- rebased against main (preserving a commit for current work)
- introduce `par_join!()` macro for a flatten join ergonomic
- resolve TODO mentioned in https://github.com/tachyon-zcash/ragu/pull/595/#discussion_r2971983762

Overall result: **6.6%** improvement over main

```
$ cargo bench --bench fuse_criterion -p ragu_pcd --features multicore
fuse                    time:   [313.81 ms 314.64 ms 315.50 ms]
                        change: [-9.3159% -6.6572% -4.5933%] (p = 0.00 < 0.05)
                        Performance has improved.

```

crates/ragu_arithmetic/src/lib.rs
```diff
@@ -104,6 +108,21 @@ pub use util::{
...
+/// different type. Supports 2..=4 closures — for higher arities prefer a
+/// data-parallel iterator.
+///
+/// ```
+/// use ragu_arithmetic::par_join;
+///
+/// let (a, b, c) = par_join!(|| 1u32, || "two", || 3.0_f64);
+/// assert_eq!((a, b, c), (1, "two", 3.0));
+/// ```
+pub use crate::__ragu_arithmetic_par_join as par_join;
```
TalDerei: > introduce par_join!() macro for a flatten join ergonomic

pushing back on this macro change, think it should have been raised as a separate comment. generally not a fan of introducing unnecessary macros, and in this context I don't think the callsites warrant the abstraction (the tuple-flattening sugar in some ways actually works against what I was trying to convey here: coarse-grained parallelism with an explicit join structure, and the macro hides that. I don't think saving a few lines in the nesting structure requires this abstraction anyway).

alxiong: okay, i'm fine dropping it. just feel like simple win. it's not something complicated or hard-to-audit. 
the macro shouldn't diminish the parallelism boundary, it's syntactic sugar and just equivalent. 
personally i like it but will go with whatever reviewers want.

crates/ragu_arithmetic/src/lib.rs
```diff
@@ -83,6 +83,10 @@ pub use coeff::Coeff;
 pub use domain::Domain;
 use ff::{Field, FromUniformBytes, WithSmallOrderMulGroup};
 pub use fft::{Ring, bitreverse};
+/// Internal re-export consumed by [`par_join!`]. `par_join!` expands to a path
+/// through `$crate`, so the helper must be reachable at the crate root.
+#[doc(hidden)]
+pub use multicore::join;
```
ebfull: Nah, something wrong about this. Just make it doc visible? What would be the problem with that?

TalDerei: there wouldn't be an issue, just need to add doc comments.

crates/ragu_circuits/src/registry.rs
```diff
@@ -436,32 +436,66 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+    /// with no registered circuit hold it alone), and masking circuits also
+    /// carry the shared global term.
+    pub fn wxy_over_domain(&self, x: F, y: F) -> Vec<F> {
+        // The key term k * (XY)^{4n-1} has no W factor, so it adds the same
+        // scalar to every domain evaluation.
         let key_scalar = self.key_sxy(x, y);
-        let mut coeffs = alloc::vec![F::ZERO; R::num_coeffs()];
+        let global_xy = crate::staging::mask::global_mask::<F, R>(x, y);
 
+        let mut evals = alloc::vec![key_scalar; self.domain.n()];
```
TalDerei, line 488: realized the old code was allocating the full poly capacity, and only ifft'd `domain.n()` with a bunch of padding, so let's just alloc the domain.

crates/ragu_pcd/src/fuse/_07_query.rs
```diff
@@ -37,31 +37,42 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
                 &registry_xy_poly,
                 &registry_wy.poly,
             ),
+        );
+
+        let query_witness = native::stages::query::Witness {
+            fixed_registry,
+            registry_wxy,
+            left: left_witness,
+            right: right_witness,
```
TalDerei: basically before, it did 13 poly evals at domain points, eval the poly at random W, and built left / right child witnesses all sequentially. now, the 13 calls are direct lookups into the vector, and the remaining work is concurrent.

crates/ragu_circuits/src/registry.rs
```diff
@@ -436,32 +436,66 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
         // global term to each mask slot inline.
-        let global_xy = crate::staging::mask::global_mask::<F, R>(x, y);
         for (i, circuit) in self.circuits.iter().enumerate() {
             let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
             let mut v = circuit.sxy(x, y, &self.floor_plans[i]);
             if circuit.is_mask() {
                 v += global_xy;
             }
-            coeffs[j] = v;
+            evals[j] += v;
```
TalDerei, line 497: per-circuit `circuit.sxy` loop remains serial (maybe worth parallelizing, reference https://github.com/tachyon-zcash/ragu/issues/573).

alxiong (approved):

PR #596, opened 2026-03-20
ragu_circuits: stop exposing CircuitObject in public API

ebfull:
This is a change in service of #594.

`CircuitObject` doesn't need to be exposed in the public API. There are some cases where this is necessary, mainly for tests, but we can rectify those later. In the mean time, we can introduce a wrapper around a boxed `CircuitObject` called a `BondingPolynomial` (appropriate, because that's the only case right now that an API consumer will deal with) and modify the registry and claims code to work generically over this wrapper type instead.

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -312,6 +312,26 @@ where
 /// An evaluable bonding object $s(X, Y)$ used to enforce well-formedness
 /// of a staged trace.
 ///
+/// A bonding polynomial is the wiring polynomial of a bonding circuit —
+/// a circuit that has only linear constraints (no multiplication gates).
+/// This gives bonding polynomials three properties that general wiring
+/// polynomials lack:
+///
+/// 1. **No dilation term.** Because the underlying circuit has no
+///    multiplication gates there is no $t(z)$, and the revdot identity
+///    simplifies to $b = s\_{y}$.
+///
+/// 2. **$k(y) = 0$ and zero constant term.** Bonding claims require
+///    $k(y) = 0$; to enforce this the constant-term position of $s(X, Y)$
+///    carries no linear constraint, making the constant term zero. General
+///    wiring polynomials always have a constant term of one (constraining
+///    the ONE wire), so the zero constant term distinguishes bonding
+///    polynomials and prevents substitution attacks.
+///
+/// 3. **Batchable.** Without multiplication gates the revdot identity is
+///    linear in the trace, so multiple traces can be folded with a random
+///    challenge and verified in a single claim.
+///
 /// Constructed via [`StageExt::mask`] and [`StageExt::final_mask`]; the
 /// underlying implementation is private to this crate.
 ///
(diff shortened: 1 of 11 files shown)


crates/ragu_circuits/src/lib.rs
```diff
@@ -244,9 +232,145 @@ pub trait CircuitObject<F: Field, R: Rank>: Send + Sync {
...
+/// Wraps a circuit into a boxed [`CircuitObject`] that can evaluate the
+/// $s(X, Y)$ polynomial.
+pub(crate) fn into_circuit_object<'a, F, C, R>(
+    circuit: C,
+) -> Result<Box<dyn CircuitObject<F, R> + 'a>>
+where
+    F: FromUniformBytes<64>,
+    C: Circuit<F> + 'a,
+    R: Rank,
+{
```
TalDerei: ACK – supersedes `into_bonding_object` in https://github.com/tachyon-zcash/ragu/pull/594#discussion_r2962799337.

crates/ragu_circuits/src/lib.rs
```diff
@@ -244,9 +232,145 @@ pub trait CircuitObject<F: Field, R: Rank>: Send + Sync {
...
+    pub(crate) fn new(inner: Box<dyn CircuitObject<F, R> + 'a>) -> Self {
+        Self { inner }
+    }
+
+    pub(crate) fn into_inner(self) -> Box<dyn CircuitObject<F, R> + 'a> {
+        self.inner
+    }
+
+    /// Evaluates the polynomial $s(x, y)$ for some $x, y \in \mathbb{F}$.
+    pub fn sxy(
```
TalDerei: are these supposed to be public? thought `BondingPolynomial` was supposed to be an opaque handle that registry immediately unwraps with `into_inner`.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -673,30 +661,6 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_constraint_counts_matches_metrics() {
```
TalDerei: can we supplement this test? don't think we should remove test coverage since this was validating hand-written constraint counts in `StageMask`.

TalDerei (commented):
In favor of `CircuitObject` API hardening, left some design observations.

crates/ragu_circuits/src/lib.rs
```diff
@@ -244,9 +232,145 @@ pub trait CircuitObject<F: Field, R: Rank>: Send + Sync {
...
+/// of a staged trace.
+///
+/// Constructed via [`StageExt::mask`] and [`StageExt::final_mask`]; the
+/// underlying implementation is private to this crate.
+///
+/// [`StageExt::mask`]: crate::staging::StageExt::mask
+/// [`StageExt::final_mask`]: crate::staging::StageExt::final_mask
+pub struct BondingPolynomial<'a, F: Field, R: Rank> {
+    inner: Box<dyn CircuitObject<F, R> + 'a>,
+}
```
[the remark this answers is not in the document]
ebfull: However, `BondingObject` is a superior name, so I'll switch to that.

Issue #598, opened 2026-03-20
Fingerprint model conflates polynomial equivalence with memoization safety


TalDerei:
Raised by @ebfull OOB and had claude file this issue from our investigation. The current `RoutineFingerprint` bundles TypeIds with the Schwartz-Zippel eval scalar, conflating two distinct invariants.

   ## Investigation

   ### 1. TypeIds may not need to be in the fingerprint

   The `RoutineFingerprint` is the tuple `(TypeId(Input), TypeId(Output), eval, num_mul, num_lc)`.

   The `eval` scalar is a Schwartz-Zippel evaluation of the routine's constraints at canonical positions. It captures the routine's internal constraint structure independently of caller context. Two routines with the same `eval` + same counts
    produce the same s(x,y) contribution when placed at the same floor plan position.

   TypeIds do not affect the polynomial. They capture Rust-level type information (input/output arity). They're necessary for memoization safety (ensuring you don't substitute a 1-input routine for a 2-input one), not for polynomial
   equivalence.

   ### 2. The tests are checking the wrong invariant

   There are 12 `test_typeid_necessary_*` tests. Each constructs a pair of routines where:
   - `eval` values are identical (confirmed with `assert_eq!(a.eval(), b.eval())`)
   - constraint counts are identical
   - only TypeIds differ
   - the test asserts `a != b`

   These tests are correct for **memoization safety** — without TypeIds, you could falsely match an `(Element) → Element` routine with an `(Element, Element) → Element` routine. But they're not testing polynomial equivalence. For the
   invariant "same s(x,y) contribution," these routine pairs genuinely ARE equivalent — they produce the same polynomial contribution.

   The test names (`test_typeid_necessary_for_input_discrimination`) suggest TypeIds are inherently necessary for fingerprint correctness, when they're actually only necessary for memoization.

   ### 3. The fingerprint conflates memoization with polynomial correctness 

   Key evidence from the code:
   - The `Counter` (fingerprinting) initializes `result = h` (nonzero Horner seed) and does an uncounted input wire remap (`metrics.rs:452`)
   - The SXY evaluator initializes `result = F::ZERO` (`sxy.rs:268`) and passes caller wire values directly — no remap

   These compute different things. The fingerprint eval is a context-independent structural hash, not the actual s(x,y) contribution. It's designed to answer "same internal structure?" rather than "same polynomial output at this circuit
   position?"

   But both concerns are bundled into one `RoutineFingerprint` type.

[...]

PR #599, opened 2026-03-20
book: routine implementation page

ebfull:


TalDerei (commented):
first pass read-through, will be taking a turn expanding.

book/src/implementation/routines.md
```diff
@@ -152,6 +214,116 @@ during assembly. **As a simplification, we assume all routine invocations are
...
+
+Each segment occupies a **non-overlapping** range of gates and constraints in
+the polynomial layout, assigned by the floor planner. Currently, the floor
+planner preserves DFS synthesis order and computes offsets via a simple prefix
+sum, but a future implementation could reorder segments to co-locate equivalent
+routines or improve memory access patterns. The only constraint is that the root
+segment remains pinned at the polynomial origin (both offsets zero), and
+no two segments may overlap.
+
+### Testability
```
TalDerei, line 308: written in present tense assuming https://github.com/tachyon-zcash/ragu/pull/503 will land in some form where these kinds of soundness tests are already implemented; future tense would require another book refactor once it's implemented.

book/src/implementation/routines.md
```diff
@@ -152,6 +214,116 @@ during assembly. **As a simplification, we assume all routine invocations are
...
+type pairs, and counts all match, the two invocations are structurally
+equivalent with overwhelming probability.
+
+Fingerprints capture only the **local** constraint structure. When entering a
+routine, the `Counter` resets its geometric sequences and Horner accumulator to
+a fixed initial state, and output wires from nested routine calls are remapped
+to fresh positions in the caller's sequence space. This ensures the fingerprint
+is independent of calling context and of any sub-routines the routine invokes.
+
+### Shallow vs. Deep Fingerprints
```
TalDerei: written in present tense assuming https://github.com/tachyon-zcash/ragu/pull/557 will land in some form; future tense here would require another book refactor once it's implemented.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+segments impose the same local constraints and could be aligned in the
+polynomial layout regardless of their subtrees — for instance, padding
+one circuit's segment against another circuit's offset. A deep
+fingerprint captures **subtree-level equivalence**: whether entire routine
+invocations are interchangeable and can be substituted via memoization.
+A segment can be shallow-equivalent to another (same local constraints,
+worth aligning) but deep-inequivalent (different children, so the
+cached subtree cannot be reused). Consolidating both levels into a single
+identity type loses this distinction and forces the floor planner to
+treat segments as different when only their subtrees differ.
```
TalDerei: language largely lifted from https://github.com/tachyon-zcash/ragu/pull/557#discussion_r2942082544.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+
+[^root-segment]: The root segment is not repositioned. It contains the special
+    `ONE` wire and is where all stage wires are located.
+
+Each segment has its own [contribution](#algebraic-description) to $s(X, Y)$. A
+leaf routine invocation — one with no nested calls — contributes exactly one
+segment. When a routine nests further calls, its **total contribution** is the
+sum of its own segment's contribution and every descendant segment's. These
+segments are not independent: routines send and receive wires through [`Input`]
+and [`Output`] gadgets, and those wires are often allocated in different
```
TalDerei, line 183: there's a lot of broken links in this book entry that need fixing.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+rearranging the wiring polynomials of the circuits in some way. The result
+is a floor plan, which is maintained by the registry.
+
+The wiring polynomial evaluators use the floor plan to properly align and
+memoize arithmetic to reduce the cost of evaluating the registry polynomial. The
+trace polynomial assembly process produces an unassembled trace of execution for
+a given witness, and the registry uses the floor plan to translate this to the
+actual trace polynomial $r(X)$.
+
+## Invocations
```
TalDerei, line 105: invocations section traces the lifetime of a driver making a routine call.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+one circuit's segment against another circuit's offset. A deep
+fingerprint captures **subtree-level equivalence**: whether entire routine
+invocations are interchangeable and can be substituted via memoization.
+A segment can be shallow-equivalent to another (same local constraints,
+worth aligning) but deep-inequivalent (different children, so the
+cached subtree cannot be reused). Consolidating both levels into a single
+identity type loses this distinction and forces the floor planner to
+treat segments as different when only their subtrees differ.
+
+### Repositioning {#repositioning}
```
TalDerei, line 298: repositioning section describes the non-overlapping segment invariant.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+The [algebraic description](#algebraic-description) decomposes a routine's
+contribution into **internal** polynomials $g_x, g_{x^{-1}}$ that depend only
+on the routine's constraint structure, and **interface** terms $h_\ell$ that
+depend on wires crossing the routine boundary. Memoization caches the internal
+part: if two invocations produce the same constraint structure, their $g_x$ and
+$g_{x^{-1}}$ are identical and can be reused. The interface terms and
+[repositioning](#repositioning) factors $X^i, Y^j$ must still be computed
+per-invocation.
+
+### Fingerprinting
```
TalDerei, line 226: the fingerprinting section can be expanded with the possible conflations we're trying to avoid, described in https://github.com/tachyon-zcash/ragu/issues/598.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,339 @@
...
+## Algebraic Description {#algebraic-description}
+
+Routines exploit the fact that contiguous sections of code tend to have
+algebraically convenient structure in the resulting wiring polynomial $s(X, Y)$:
+
+* [`gate`] advances an $i$ counter and returns $(X^{2n - 1 - i}, X^{2n + i}, X^{4n - 1 - i}, X^i)$ wires.
+* `enforce_zero` advances a counter $j$ and adds a fresh linear combination of previous wires multiplied by $Y^j$.
+
+Most of the linear constraints created within a routine consist purely of wires
+created within that routine, meaning that the **internal** contribution of a
```
ebfull:
```suggestion
Most of the linear constraints created within a routine consist purely of wires also
created within that routine, meaning that the **internal** contribution of a
```

TalDerei: folding into staged changes

book/src/implementation/routines.md
```diff
@@ -318,10 +319,11 @@ in a single check. The non-overlapping invariant
 ensures that any disagreement is a real bug rather than a masking
 coincidence. Because the native path does not depend on the
 fingerprinting model at all, this comparison catches errors in the model
-itself — a shallow fingerprint that over-groups, or a deep fingerprint
+itself — a routine fingerprint that over-groups, or a memo fingerprint
 that omits a necessary field — not only bugs in the memoization code.
 
 [`RoutineFingerprint`]: ragu_circuits::metrics::RoutineFingerprint
+[`MemoFingerprint`]: ragu_circuits::metrics::MemoFingerprint
```
TalDerei, line 342: https://github.com/tachyon-zcash/ragu/pull/599/commits/7b42be4fc55f3d867ac07fe8d8ee9f80342d83e9 changes terminology from shallow / deep to routine / memo fingerprints to align with https://github.com/tachyon-zcash/ragu/pull/557 impl. fixes some other stale code / diagrams refs. the motivation there is consistent code review.

book/src/implementation/routines.md
TalDerei: https://github.com/tachyon-zcash/ragu/pull/599/commits/094b3f195dc004f12d3db4c562a8b9d8f1c3600b

TalDerei:
pushed some more updates, releasing lock again.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,355 @@
...
+method, but this almost always defeats the purpose of the abstraction. Instead,
+they are meant to be invoked through the [`Driver::routine`] method, which hands
+control and visibility of its execution to the driver.
+
+```admonish info
+Drivers do not learn the [`TypeId`] of a [`Routine`] they are asked to `execute`,
+since [`Routine`] does not implement [`Any`] / `'static`. Even if they could
+learn this, routines of the same type are allowed to diverge in their actual
+behavior because they can be arbitrarily configured, so long as they remain
+deterministic.
```
alxiong, line 20: this reads cryptic (or too abstract) to me.
it's not clear how `Any`/`static` is necessary, it's also ambiguous what does "arbitrarily configured" mean.
 
I'd recount the 3 caveat points in https://github.com/tachyon-zcash/ragu/issues/62#issuecomment-3963126884, expanded each with an example. 

- 2 routines with the same Input/Output type: `Txz` and `ScaledTxz` (link to existing [paragraph](https://tachyon.z.cash/ragu/guide/routines.html#parameterization)), but they are diff routines. 
- Even Routine with the same type, among all possible `ScaledTxz` or maybe `HashChain` routine where `output = H(...H(input)...)`, where the `scale`/`num_iter` is configurable.
  - given this point and the last, it should be useful and clear what we mean by "Routines is not fungible" and "behavior can be arbitrarily configured".
- same shape (num_constraint + num_gate): just place those constraints on different wires, they lead to different routines.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,164 @@
...
+routine, such as the input gadget wires or the fixed `ONE` wire. This is less
+amenable to memoization because these wires can vary between invocations; we
+must generally maintain separate polynomials $h_\ell$ for every interface wire.
+
+Taken together, the **contribution** of a routine invocation to $s(X, Y)$ can be
+written as
+
+$$
+Y^j \Big( X^i g_x + X^{-i} g_{x^{-1}} + \sum_\ell h_\ell \Big)
+$$
```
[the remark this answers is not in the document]
alxiong: agree, too specific will lead to very complicated closed-form expression. 

but i wonder if we can:
- instead of using subscript, clarity the indeterminant in poly: $g_1(X, Y), g_2(X^{-1}, Y)$
  - it's not immediately clear that $g_x$ is actually a bivariate poly. 
  - The important thing to highlight here is that **$g_1, g_2$ are invariant to the routine** no matter when or how many times this routine is invoked
- clarify that here $(i, j)$ is the starting `gate` index and `constraint` index when entering the routine, and they are common factors for all internal contribution. 
- completely drop the $h_\ell$, I think we should _only express the internal contribution_, since this is the part where we memoize and shift anyway. only mention that overall contribution = internal + interface is enough.
- add another sentence to explain why interface contribution cannot be factored this way (i.e. be memorized): 
  - unknown location of the allocated input gadget -> unknown greatest common $X^i$ factor
  - worse: input gadget can have different relative position to the start of the routine for different invocations. -> no longer something that's invariant to the routine, which is the core reason why it's not memoized.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,355 @@
...
+    W --> TE
+    metrics --> FP
+    FP --> ED
+    trace --> TP["r(X)"]
+    FP --> TP
+```
+
+Ragu learns about invocations by executing circuit code in an analysis pass that
+collects metrics about each routine invocation. Because circuit synthesis is
+deterministic, invocations appear in a canonical **DFS order** during
```
alxiong, line 86: clearer if we first mention "nested routine is possible" before "DFS order".

I'd also link to [our rustdoc](https://tachyon.z.cash/ragu/internal/ragu_circuits/floor_planner/index.html), an example is better than a paragraph of abstract description.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,164 @@
...
+        MP[metrics pass] --> metrics
+        ED[eval drivers] --> SP["s(X,Y)"]
+    end
+    C --> rx
+    C --> MP
+    trace --> R
+    metrics --> R
+    FP --> ED
+    FP --> TP
+```
```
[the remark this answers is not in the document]
alxiong: question/comments:
- why are "metric pass", "eval driver", and "trace eval" (or in Tal's revision, "rx") boxes? imo, we can remove them. 
  - metric pass should be dropped, eval drivers should be merged with s(X,Y), rx should be dropped
- I would pull "metrics" object out of "Eval path" and become a common inputs to two regions: "eval path" and "metric pass" (which replace the current "registry" label)

book/src/implementation/routines.md
```diff
@@ -0,0 +1,355 @@
...
+All of the metrics for every wiring polynomial added to the registry are fed
+into a **floor planner** that is responsible for making scheduling and
+relocation decisions in advance of future circuit operations, such as wiring
+polynomial evaluation or trace polynomial assembly. The floor planner's goal is
+to maximize the optimization opportunities of those operations, usually by
+rearranging the wiring polynomials of the circuits in some way. The result
+is a floor plan, which is maintained by the registry.
+
+The wiring polynomial evaluators use the floor plan to properly align and
+memoize arithmetic to reduce the cost of evaluating the registry polynomial. The
```
alxiong, line 100: clarify that registry maintain not only the floor plan, but also routine collections (also produced during "metric pass") -- their identities/fingerprints, shape. 

my main point here being that current phrasing wasn't clear about who builds up the "routine registry", we should clarify that it's not the evaluators/drivers.

alxiong (commented):
From "Invocation" subsection onward, it doesn't read like a spec to reason along with, but a colloquial reiteration of what our code does. which I hope gets revised, because it adds (arguably) little value to just have a text description of the implementation logic that already has good code doc. 

I was hoping this page is more about first-principle description/rationale of the routine memoization/identification algorithms, so that we can audit and check our reasoning. Lots of these rationales are scattered across Sean and Tal's github comments. 

----

okay, maybe my comments is little unfair, we do have lots of useful explanation so far, but it doesn't flow smoothly yet, there's no logic design flow, just loosely ordered collection of code pointers and inter-related concepts.

PR #600, opened 2026-03-21
chore(circuits): add routing method to `MultiStageCircuit`

TalDerei:
Just a draft that exercises a slightly different shape than https://github.com/tachyon-zcash/ragu/pull/594. Can be used as a reference to source from.

### DIFF
--- a/crates/ragu_circuits/src/staging/bonding.rs
+++ b/crates/ragu_circuits/src/staging/bonding.rs
@@ -181,7 +181,7 @@ impl<'dr, F: Field> Driver<'dr> for BondingValidator<F> {
 
 /// Wraps a [`CircuitObject`] and strips the `enforce_one` contribution,
 /// giving a zero constant term in $Y$.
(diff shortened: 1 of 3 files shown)


crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -0,0 +1,307 @@
...
+        adapter,
+        metrics,
+    }))))
+}
+
+/// Bridges [`MultiStageCircuit::routing`](super::MultiStageCircuit::routing)
+/// to [`Circuit`] so we can reuse the standard synthesis drivers.
+/// Pre-allocates gate wire handles with the real driver, then hands a
+/// [`RoutingDriver`] decorator to the routing method.
+struct Adapter<S, F, R>(core::marker::PhantomData<(S, F, R)>);
```
ebfull: The `Adapter + s::sx::eval/s::sy::eval` indirection isn't needed, you can synthesize the circuit into a `CircuitObject` and then delegate `sx`/`sy` to it directly. (Stripping `enforce_one` afterwards before returning of course.) That removes the `Processed` wrapper, the early-return zero guards, and most of this file.

PR #601, opened 2026-03-21
chore: structurally couple compute_f and poly_queries query count

LucidSamuel:
'`compute_f` (`fuse/_08_f.rs`) and
   `poly_queries` (`compute_v.rs`) independently enumerate 20+
  polynomial queries whose α-weighting must agree. Adds a shared
  `NUM_POLY_QUERIES` constant and runtime assertions on both sides so
   that additions or removals to either query list are caught
  immediately.


ebfull:
I'll have to revisit this, I'm scared of adding more constants that we have to keep track of. This one seems slightly arbitrary; I wish we could base it on an existing constant so we don't have to visit it whenever we update the f(X) poly again.

TalDerei:
closing per comment above, we should probably avoid introducing arbitrary new constants.

PR #605, opened 2026-03-21
Unified, sparse structured polynomials

ebfull:
Replaces the separate `structured::Polynomial` and `unstructured::Polynomial` types with a single `sparse::Polynomial` backed by block storage for sparsity encoding.

## Motivation

The existing structured/unstructured polynomial split doesn't expose the `d` wire within gates, which is needed for upcoming optimizations (#604, #565, #606). Those optimizations will also produce very sparse `d` vectors. The current structured polynomial only supports "tail-sparse" subvectors and can't exploit the contiguous empty blocks that already appear (e.g. in stage trace polynomials).

## Approach

- **Unify the two types.** In practice, the internal structure of a structured polynomial is only used during construction — after that, it's just coefficients. A new `sparse::view::View` builder captures the construction-time structure and produces a single `sparse::Polynomial` with no retained internal layout. Unlike the old `View` (a borrowing reference into an existing polynomial's fields), the new `View` is an owned builder — you fill it, then call `.build()`.
- **Block-based sparsity.** The new representation stores sorted, non-overlapping blocks of contiguous nonzero coefficients. Operations like `commit`, `eval`, `dilate`, and `revdot` skip gaps entirely, so cost scales with stored coefficients rather than total degree.
- **Expose the `d` wire.** The old `View` only provided `a`, `b`, `c` — the `d` vector was internal to the structured polynomial and only accessible indirectly via the backward view's `c` field. The new `View` exposes all four wires in both forward and backward perspectives. No consumer uses `d` yet, but this lays the groundwork for #604 / #565 / #606.
- **Stricter bounds.** The old `View` left bound enforcement to the caller, with violations caught by assertions later during `eval`/`commit`/`dilate`. The new `View::build()` panics immediately if any wire buffer exceeds `R::n()`.
- **Eliminate cross-type glue.** The old `unstructured::Polynomial` had special methods like `add_structured` and `add_unstructured`, and the structured polynomial needed `.unstructured()` for many operations. With a single type, all of this is gone.

## Testing

The old structured/unstructured types had hand-written unit tests that exercised specific configurations. The new `sparse` module replaces these with a comprehensive proptest suite covering all construction paths (forward view, backward view, sparse wire vectors, `from_coeffs`), algebraic operations (`add`, `sub`, `scale`, `negate`, `fold`), evaluation (`eval`, `dilate`, `revdot`), iteration (forward, reverse, interleaved), commitment consistency against a dense reference, and ring-FFT roundtrips. Dedicated edge-case tests cover zero polynomials, degree-boundary coefficients, the `d` wire in isolation, and the alloc-optimization sparsity pattern (mostly-zero `d` with scattered nonzero entries).

## Changes

- Delete `structured.rs` and `unstructured.rs`
- Add `polynomials/sparse/` with block-compressed polynomial, owned view builder, and proptest-based test suite
- Update all consumers in `ragu_circuits` and `ragu_pcd`

### DIFF
--- a/crates/ragu_circuits/src/polynomials/sparse/mod.rs
+++ b/crates/ragu_circuits/src/polynomials/sparse/mod.rs
@@ -104,29 +104,11 @@ impl<T, R: Rank> Polynomial<T, R> {
 /// within a block rather than triggering a split. Inline zeros waste MSM
 /// slots in [`commit`](Polynomial::commit), so this is kept small. The
 /// tolerance covers only the per-block overhead (allocation, merge
-/// iterations in [`combine_assign`](Polynomial::combine_assign)) — eval and
-/// dilate gap-skip costs are handled by [`pow_gap`], which avoids the
-/// fixed 64-squaring overhead of `pow_vartime` for small exponents.
+/// iterations in [`combine_assign`](Polynomial::combine_assign)) — each
+/// extra block requires a `pow_vartime` call to skip the gap.
 ///
 /// TODO(#608): benchmark to determine the optimal value.
 const GAP_TOLERANCE: usize = 4;
-
-/// Computes `z^gap` using chained multiplication for small gaps and
-/// square-and-multiply for large ones. `ff::Field::pow_vartime` iterates
-/// all 64 bits of a `u64` exponent unconditionally, so for gaps below 64
-/// chained multiplication is significantly cheaper.
-#[inline]
-fn pow_gap<F: Field>(z: F, gap: usize) -> F {
-    if gap < 64 {
-        let mut acc = F::ONE;
-        for _ in 0..gap {
-            acc *= z;
-        }
-        acc
-    } else {
-        z.pow_vartime([gap as u64])
-    }
-}
 
 /// Splits `data` into runs of coefficients and appends each run to `out` as
 /// `(base + run_offset, run_values)`. Zero gaps of up to [`GAP_TOLERANCE`]
@@ -380,7 +362,7 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
         for (start, data) in self.blocks.iter().rev() {
             let gap = prev_start - (start + data.len());
             if gap > 0 {
-                result *= pow_gap(z, gap);
+                result *= z.pow_vartime([gap as u64]);
             }
             for coeff in data.iter().rev() {
                 result = result * z + *coeff;
@@ -388,7 +370,7 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
             prev_start = *start;
         }
         if prev_start > 0 {
-            result *= pow_gap(z, prev_start);
+            result *= z.pow_vartime([prev_start as u64]);
         }
         result
     }
@@ -401,7 +383,7 @@ impl<F: Field, R: Rank> Polynomial<F, R> {
         for (start, data) in &mut self.blocks {
             let gap = *start - prev_end;
             if gap > 0 {
-                power *= pow_gap(z, gap);
+                power *= z.pow_vartime([gap as u64]);
             }
             for coeff in data.iter_mut() {
                 *coeff *= power;
--- a/crates/ragu_circuits/src/polynomials/sparse/view.rs
+++ b/crates/ragu_circuits/src/polynomials/sparse/view.rs
@@ -177,8 +177,7 @@ impl<T, R: Rank> View<T, R, Backward> {
 
 impl<T, R: Rank, P: Perspective> View<T, R, P> {
     /// Consumes this view, mapping wire buffers to degree positions and
-    /// producing a [`Polynomial`]. The resulting blocks preserve internal
-    /// zeros from the wire buffers.
+    /// producing a [`Polynomial`].
     ///
     /// # Panics
     ///


TalDerei:
https://github.com/tachyon-zcash/ragu/pull/565#issuecomment-4069897341 and https://github.com/tachyon-zcash/ragu/pull/606#discussion_r2970546727 for supplemental context.

crates/ragu_circuits/src/polynomials/sparse/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+        for (start, data) in &self.blocks {
+            dense[*start..*start + data.len()].copy_from_slice(data);
+        }
+        dense
+    }
+
+    /// Iterates over the coefficients of this polynomial in ascending order of
+    /// degree, yielding `F::ZERO` for gaps between blocks.
+    pub fn iter_coeffs(&self) -> impl DoubleEndedIterator<Item = F> + ExactSizeIterator + '_ {
+        self.to_dense().into_iter()
```
ebfull: There must be a more efficient, custom iterator type we can implement here. It would not need to create a temporary dense vector at all, and could yield elements (double-ended) accounting for the gaps using zeroes.

crates/ragu_circuits/src/polynomials/sparse/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+/// store and iterate than the per-block overhead in [`eval`](Polynomial::eval)
+/// and [`dilate`](Polynomial::dilate) — each extra block requires a
+/// `pow_vartime` call to skip the gap.
+const GAP_TOLERANCE: usize = 4;
+
+/// Splits `data` into runs of coefficients and appends each run to `out` as
+/// `(base + run_offset, run_values)`. Zero gaps of up to [`GAP_TOLERANCE`]
+/// consecutive zeros are kept inline within a run; longer gaps cause a split.
+/// Leading and trailing zeros are always trimmed.
+fn extend_runs<F: Field>(out: &mut Vec<(usize, Vec<F>)>, base: usize, data: Vec<F>) {
```
TalDerei: ACK – greedily splits dense vector into blocks.

crates/ragu_circuits/src/polynomials/sparse/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+    pub fn fold<E: Borrow<Self>>(polys: impl IntoIterator<Item = E>, scale_factor: F) -> Self {
+        polys.into_iter().fold(Self::default(), |mut acc, poly| {
+            acc.scale(scale_factor);
+            acc.add_assign(poly.borrow());
+            acc
+        })
+    }
+
+    /// Evaluates this polynomial at `z` using reverse Horner's method by block.
+    pub fn eval(&self, z: F) -> F {
```
TalDerei: ACK – diverges from old impl; reverse horner's method with `pow_vartime` for gaps between blocks.

crates/ragu_circuits/src/polynomials/sparse/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+impl<T, R: Rank> Polynomial<T, R> {
+    /// Panics if the block list violates any structural invariant: blocks must
+    /// be sorted by start index, non-empty, non-overlapping, and each block
+    /// must fit within `[0, R::num_coeffs())`. Adjacent blocks are permitted.
+    fn assert_invariants(&self) {
+        let mut prev_end: usize = 0;
+        for (i, (start, data)) in self.blocks.iter().enumerate() {
+            debug_assert!(!data.is_empty(), "block {i} is empty");
```
TalDerei: promote these to assertions?

crates/ragu_circuits/src/polynomials/sparse/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+/// `pow_vartime` call to skip the gap.
+const GAP_TOLERANCE: usize = 4;
+
+/// Splits `data` into runs of coefficients and appends each run to `out` as
+/// `(base + run_offset, run_values)`. Zero gaps of up to [`GAP_TOLERANCE`]
+/// consecutive zeros are kept inline within a run; longer gaps cause a split.
+/// Leading and trailing zeros are always trimmed.
+fn extend_runs<F: Field>(out: &mut Vec<(usize, Vec<F>)>, base: usize, data: Vec<F>) {
+    let mut run_start: Option<usize> = None;
+    let mut run = Vec::new();
```
TalDerei:
```suggestion
let mut run = Vec::with_capacity(data.len());

File: crates/ragu_circuits/src/polynomials/sparse/mod.rs [hunk 6]
```diff
@@ -0,0 +1,574 @@
...
+                for (d, r) in data[off..off + rd.len()].iter_mut().zip(rd) {
+                    op(d, r);
+                }
+            }
+
+            extend_runs(&mut out, cluster_start, data);
+        }
+
+        out.shrink_to_fit();
+        self.blocks = out;
```
TalDerei: path that bypasses invariant checks from `from_blocks`.

```suggestion
self.blocks = out;
self.assert_invariants();

File: crates/ragu_circuits/src/polynomials/sparse/mod.rs [hunk 7]
```diff
@@ -0,0 +1,574 @@
...
+        if self.blocks.is_empty() {
+            let mut out = Vec::new();
+            for (s, d) in &other.blocks {
+                let mut v = alloc::vec![F::ZERO; d.len()];
+                for (o, r) in v.iter_mut().zip(d) {
+                    op(o, r);
+                }
+                extend_runs(&mut out, *s, v);
+            }
+            self.blocks = out;
```
TalDerei: path that bypasses invariant checks from `from_blocks`.

```suggestion
self.blocks = out;
self.assert_invariants();

File: crates/ragu_circuits/src/polynomials/sparse/mod.rs [hunk 8]
```diff
@@ -0,0 +1,574 @@
...
+    fn sub_assign(r: &mut Self, other: &Self) {
+        Polynomial::sub_assign(r, other);
+    }
+}
+
+impl<F: Field, R: Rank> core::ops::AddAssign<&Self> for Polynomial<F, R> {
+    fn add_assign(&mut self, rhs: &Self) {
+        Polynomial::add_assign(self, rhs);
+    }
+}
```
TalDerei: need `SubAssign` for symmetry?

PR #606, opened 2026-03-21
feat(circuits): cheaper gate layout for alloc

TalDerei:
Optimization vector originally suggested by @ebfull and to be rebased over #605 and orthognal to https://github.com/tachyon-zcash/ragu/issues/607. We knew each nonzero coefficient in the trace polynomial constituted a scalar multiplication in the commitment. The previous layout `(a, b, a*b, 0)`  had 3 nonzero coefficients / alloc gate (3 scalars muls). Internally this behaved like `mul()`, where the first alloc in the pair called mul internally to create the gate, the second alloc filled in the b-wire, and an extraneous c-wire was computed and discarded. 

The new paired allocation pattern `(0, a, 0, b)` lives in `alloc` and has 2 nonzero coefficients / alloc gate (2 scalars muls), which saves 1 scalar mul per _paired_ alloc. It's an allocation trick where you don't need a real multiplication, and instead of the normal gate layout `(a, b, c, 0)` which wastes the allocation, we shift the representation to `(0, a, 0, b)`. I modeled this as a dedicated API extension in #604 to separate concerns before attempting any unification with our existing alloc method.

I reference "layout" here, the structured polynomial and the gate wire layouts remain unchanged. This only affects which wire positions the alloc pairing uses within that layout.

Moreover, this doesn't necessarily superceed https://github.com/tachyon-zcash/ragu/pull/565 because they do different things:
- That construction exposes 3 usable wires (a, b, d) at zero-product constraints (ie. when a * b = 0), which could save gates by packing a third allocation in the d-wire,
- This construction only does 2 allocations per gate, but saves scalar muls per paired allocs in the multiexp. Every gate is a * b = c and c * d = 0,  and alloc gates set c = 0, and mul gates set d = 0.

### DIFF
--- a/crates/ragu_circuits/src/s/sx.rs
+++ b/crates/ragu_circuits/src/s/sx.rs
@@ -158,13 +158,13 @@ struct Evaluator<'fp, F: Field, R: Rank> {
     /// for the $c$ wire.
     base_c_x: F,
 
-    /// Inverse of `base_b_x`: $(x^{2n})^{-1}$. Used in [`gate`](DriverTypes::gate)
-    /// to derive the $d$-wire monomial $x^i$ from `current_b_x` ($x^{2n+i}$)
-    /// without tracking a separate running monomial.
+    /// Correction factor $(x^{-2n})$ that converts a $b$-wire monomial
+    /// $x^{2n+i}$ into the corresponding $d$-wire monomial $x^i$.
     ///
-    /// This field is only read by `gate`, not by [`mul`](Driver::mul), so the
-    /// extra multiplication is skipped when callers don't need the $d$ wire.
-    base_b_x_inv: F,
+    /// Only read by [`gate`](DriverTypes::gate), not by [`mul`](Driver::mul),
+    /// so the extra multiplication is skipped when callers don't need the
+    /// $d$ wire.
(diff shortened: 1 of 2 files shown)


TalDerei:
reminder to also address feedback like https://github.com/tachyon-zcash/ragu/pull/612#discussion_r2977605243 and update stale references here.

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -160,9 +158,13 @@ struct Evaluator<'fp, F: Field, R: Rank> {
...
 
-    /// Base monomial $x^0$, used to compute routine starting monomials for the
+    /// Correction factor $(x^{-2n})$ that converts a $b$-wire monomial
+    /// $x^{2n+i}$ into the corresponding $d$-wire monomial $x^i$.
+    ///
+    /// Only read by [`gate`](DriverTypes::gate), not by [`mul`](Driver::mul),
+    /// so the extra multiplication is skipped when callers don't need the
     /// $d$ wire.
-    base_d_x: F,
+    b_to_d: F,
```
alxiong: by the same logic, there's a fixed `2^2n` offset between `a` and `c`.

PR #610, opened 2026-03-23
chore(circuits): bonding polynomial suggestions

TalDerei:
Reviewed #600 post-hoc — ACK on architectural approach, with some suggestions.

### DIFF
--- a/crates/ragu_circuits/src/staging/bonding.rs
+++ b/crates/ragu_circuits/src/staging/bonding.rs
@@ -63,6 +63,7 @@ where
     where
         Self: 'a,
     {
+        // Validate: run synthesis with a driver that rejects mul and ONE usage.
         let mut validator = BondingValidator::<F>::new();
         self.witness(&mut validator, Empty)?;
         if let Some(msg) = validator.error {


crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -65,6 +65,10 @@ where
         Self: 'a,
     {
         // Validate: run synthesis with a driver that rejects mul and ONE usage.
+        // TODO: we may want to fuse this into the metrics pass inside
+        // `into_circuit_object` to eliminate the separate synthesis and the
+        // two-pass divergence risk (a circuit with interior mutability could
+        // theoretically behave differently across passes).
```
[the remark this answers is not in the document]
ebfull: Interesting. I don't think we should try too hard to defend against circuit developers that break determinism expectations, there's no purity guarantee that we can perfectly ensure in the language.

crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -221,7 +231,9 @@ impl<F: Field, R: Rank> CircuitObject<F, R> for Stripped<'_, F, R> {
     }
 
     fn constraint_counts(&self) -> (usize, usize) {
-        self.0.constraint_counts()
+        let (mul, lin) = self.0.constraint_counts();
+        // The inner object includes the `enforce_one` constraint that we strip.
+        (mul, lin - 1)
```
ebfull: I want to remove this `constraint_counts` method or at least revisit it. It's only load bearing for testing like our constraint tests in `ragu_pcd`. It's hard to judge what it means or what the caller expects it to mean because of these special "system constraints" (linear or multiplication). Can you open an issue to revisit this entire `constraint_counts` method?

crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -63,7 +63,6 @@ where
     where
         Self: 'a,
     {
-        // Validate: run synthesis with a driver that rejects mul and ONE usage.
```
ebfull: Not sure why this comment got deleted? It seems accurate.

TalDerei: accidentally dropped during rebase, restored.

PR #620, opened 2026-03-26
test(pcd): end-to-end d-wire alloc-pairing regression

LucidSamuel:
PR #606 changed alloc() pairing from (a, b, 0, 0) to (0, b, 0, d), placing the second allocation in each pair on
the d-wire. added a dedicated integration test that exercises this path through the full proof lifecycle: seed → fuse → verify.


TalDerei:
this can be closed; we have comprehensive proptests exercising this.

PR #622, opened 2026-03-28
Make constant() and add() fallible

alxiong:
Addressing my own comment: https://github.com/tachyon-zcash/ragu/pull/600#discussion_r3004316063

Change `Driver::constant()` and `Driver::add()` to return `Result<Self::Wire>,` aligning them with `mul(),` `alloc(),` `gate(),` and `enforce_zero()` which were already fallible. This lets `BondingValidator` return `Err` immediately on constraint violations instead of accumulating errors in an `Option` field and checking after the witness completes.

### DIFF
--- a/crates/ragu_primitives/src/boolean.rs
+++ b/crates/ragu_primitives/src/boolean.rs
@@ -58,11 +58,11 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
     }
 
     /// Computes the NOT of this boolean. This is "free" in the circuit model.
-    pub fn not(&self, dr: &mut D) -> Self {
+    pub fn not(&self, dr: &mut D) -> Result<Self> {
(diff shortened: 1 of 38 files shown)


crates/ragu_primitives/src/boolean.rs
```diff
@@ -58,11 +58,11 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
     }
 
     /// Computes the NOT of this boolean. This is "free" in the circuit model.
-    pub fn not(&self, dr: &mut D) -> Self {
+    pub fn not(&self, dr: &mut D) -> Result<Self> {
```
ebfull: It requires things that are currently infallible to have fallible code paths, even though our standard drivers (trace, sx/sy/sxy, etc.) do not have failure conditions for virtual wires.

PR #625, opened 2026-03-29
[WIP] Loading / Copying Circuits

ebfull:

PR #627, opened 2026-03-31
chore(pcd): loading / copying suggestions

TalDerei:
Draft; Extension of https://github.com/tachyon-zcash/ragu/pull/625. Not sure yet if the last commit is neccesary, thinking it through.

**Update:** This actually appears to be mostly duplicative because when computing `ProofInputs::unified_bridge_ky` for the left/right child, `self.unified.write()` serializes the child's own unified instance which already contains nearly all the same bridge commitments. Also, this change clearly bloats the unified instance across all circuits. cc @ebfull


TalDerei:
see readme on why this is closeable, cc @ebfull.

PR #628, opened 2026-04-02
Bundle HEADER_SIZE const generic into PcdConfig trait

### LINKED ISSUE #243: Bundle parameterized const generics into a config trait
Two const generics (and potentially more in the future) propagate through our methods and impl blocks, cascading through method calls in `ragu_pcd`:

```ts
pub struct Circuit<'params, C: Cycle, R, const HEADER_SIZE: usize, const NUM_REVDOT_CLAIMS: usize>
pub struct Witness<'a, C: Cycle, R: Rank, const HEADER_SIZE: usize, const NUM_REVDOT_CLAIMS: usize>
pub struct Stage<C: Cycle, R, const HEADER_SIZE: usize, const NUM_REVDOT_CLAIMS: usize>
pub struct Output<'dr, D: Driver<'dr>, C: Cycle, const HEADER_SIZE: usize>
pub struct Application<'params, C: Cycle, R: Rank, const HEADER_SIZE: usize>
```

We may want to consider using a `PcdConfig` trait that bundles these const generics to simplify type signatures:

```ts
  trait PcdConfig {
      const HEADER_SIZE: usize;
      const NUM_REVDOT_CLAIMS: usize;
  }

  struct DefaultConfig;
  impl PcdConfig for DefaultConfig {
      const HEADER_SIZE: usize;
      const NUM_REVDOT_CLAIMS: usize;
  }

  pub struct Circuit<'params, C: Cycle, R, Cfg: PcdConfig> {}
  pub struct Stage<C: Cycle, R, Cfg: PcdConfig> {}
```

MavenRain:
Replace the cascading `const HEADER_SIZE: usize` parameter with a `PcdConfig` trait whose `HeaderSize` associated type implements `Len`, following the precedent set by `fold_revdot::Parameters`. Internally, stages and circuits take `HS: Len`; the `Application` and fuse pipeline use `Cfg: PcdConfig` and extract `Cfg::HeaderSize`.

Closes #243

### DIFF
--- a/crates/ragu_pcd/src/fuse/_01_application.rs
+++ b/crates/ragu_pcd/src/fuse/_01_application.rs
@@ -11,11 +11,11 @@ use ragu_core::Result;
 use rand::CryptoRng;
 
 use crate::{
-    Application, Header, Pcd, Proof, proof,
+    Application, Header, Pcd, PcdConfig, Proof, proof,
     step::{Step, internal::adapter::Adapter},
 };
 
-impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_SIZE> {
+impl<C: Cycle, R: Rank, Cfg: PcdConfig> Application<'_, C, R, Cfg> {
     pub(super) fn compute_application_proof<'source, RNG: CryptoRng, S: Step<C>>(
         &self,
         rng: &mut RNG,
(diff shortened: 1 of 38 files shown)


TalDerei:
thanks for this PR! this is directionally good, but we're going to defer this kind of config trait change for now.

PR #629, opened 2026-04-02
Add MaybeArith extension trait for Maybe<T> arithmetic

### LINKED ISSUE #570: `Maybe<T>` arithmetic methods
In some places (like `Element`), witness arithmetic on `Maybe<T>` can be simplified from say

```ts
    pub fn mul(&self, dr: &mut D, other: &Self) -> Result<Self> {
        let product = D::just(|| {
            let a = *self.value.snag();
            let b = *other.value.snag();
            a * b
        });
```

to 

```ts
    pub fn mul(&self, dr: &mut D, other: &Self) -> Result<Self> {
        let product = self.value.maybe_mul(&other.value);
``` 

where you don't need to call `just` or `snag`, rather the method accesses `self.0` directly. This would require expanding `Maybe<T>`:

```ts
    /// Adds two `Maybe<T>` values element-wise, returning a new `Maybe<T>`.
    /// For [`Always`], performs the addition; for [`Empty`], returns [`Empty`].
    fn maybe_add(&self, other: &Perhaps<Self::Kind, T>) -> Perhaps<Self::Kind, T>
    where
        T: core::ops::Add<Output = T> + Copy + Sync;

    /// Subtracts two `Maybe<T>` values element-wise, returning a new `Maybe<T>`.
    /// For [`Always`], performs the subtraction; for [`Empty`], returns [`Empty`].
    fn maybe_sub(&self, other: &Perhaps<Self::Kind, T>) -> Perhaps<Self::Kind, T>
    where
        T: core::ops::Sub<Output = T> + Copy + Sync;

    /// Multiplies two `Maybe<T>` values element-wise, returning a new `Maybe<T>`.
    /// For [`Always`], performs the multiplication; for [`Empty`], returns [`Empty`].
    fn maybe_mul(&self, other: &Perhaps<Self::Kind, T>) -> Perhaps<Self::Kind, T>
    where
        T: core::ops::Mul<Output = T> + Copy + Sync;

    /// Negates a `Maybe<T>` value, returning a new `Maybe<T>`.
    /// For [`Always`], performs the negation; for [`Empty`], returns [`Empty`].
    fn maybe_neg(&self) -> Perhaps<Self::Kind, T>
    where
        T: core::ops::Neg<Output = T> + Copy + Sync;
```

Not sure this is worth it given the expanded trait surface area? I suspect there are driver-level ergonomics improvements like this as well. The scope here can probably be expanded to implement synthetic sugar in the `primitives` crate to reduce verbosity at the user-level.

MavenRain:
Introduce maybe_add, maybe_sub, maybe_mul, and maybe_neg combinators that lift standard arithmetic into the Maybe abstraction, eliminating the just/snag boilerplate pattern.  Apply to Element's add, sub, and mul methods as the first consumer.

Closes #570

### DIFF
--- a/crates/ragu_core/src/maybe/arith.rs
+++ b/crates/ragu_core/src/maybe/arith.rs
@@ -1,0 +1,160 @@
+//! Arithmetic combinators for [`Maybe<T>`] values.
+//!
+//! [`MaybeArith`] is a blanket extension trait that lifts standard arithmetic
+//! operations (`Add`, `Sub`, `Mul`, `Neg`) into the [`Maybe`] abstraction.
+//! For [`Always`] values the operation executes; for [`Empty`] values the
+//! closure is dead-code eliminated, exactly like [`Maybe::just`].
+//!
+//! # Example
+//!
+//! ```ignore
+//! // Before:
+//! let sum = D::just(|| *a.snag() + *b.snag());
+//!
+//! // After:
+//! let sum = a.maybe_add(&b);
+//! ```
+//!
+//! [`Always`]: super::Always
+//! [`Empty`]: super::Empty
+
+use core::ops::{Add, Mul, Neg, Sub};
+
+use super::{Maybe, MaybeKind, Perhaps};
+
+/// Arithmetic combinators for [`Maybe<T>`] values.
+///
+/// This trait is automatically implemented for every type that implements
+/// [`Maybe<T>`].  It provides convenience methods that combine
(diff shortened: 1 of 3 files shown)


ebfull:
Why can't e.g. `Maybe<T>: Add<Self>` when `T: Add<Self>`?

grr. alright, I see how this isn't really possible to bound or express...

TalDerei:
we're going to defer this kind of change now as well.

Issue #218, opened 2025-12-10
Enforce consistent import ordering


TalDerei:
We should adhere to consistent import ordering:

```ts
// 1. External crates (std, third-party)
// 2. Core / alloc (when in no_std context)
// 3. Local crate imports
```

There’s a way to enforce this via clippy, but the lint is only available on _nightly_. We should look for a viable workaround so we can enforce this in stable linting.

PR #632, opened 2026-04-06
Preliminaries for Loading / Copying circuits

ebfull:
Collections of commits working to improve and clean things prior to #625.


TalDerei:
just saw this merged, will review post-hoc.

alxiong:
browse through the commits. only nit comments:

- can we version control the `rustfmt.toml` so that we are consistent? (we can add `+nightly fmt` in CI etc.)
- why drop the `define_bridge_stage!` and expand them in [6687e78](https://github.com/tachyon-zcash/ragu/pull/632/commits/6687e7892b89568c96aed91b16b17fc8571b211a). isn't macro better for avoiding accidental drift?
- In `ProofBuilder`'s lazily computed commitment, shouldn't they be `OnceCell` instead of `Cell`?

ebfull:
@alxiong

* happy to merge such a change to CI if you'd like to open it! I don't want to run tests or lints themselves using nightly though. The problem with doing this is that it will require thinking about how (and to what extent) to cache a nightly toolchain... I would really like nightly only used to invoke `cargo fmt --check` using those flags, and I don't want developers to run into problems locally running `fmt` on stable.
* for the macro, all of these stages will end up being more complicated and the macro (as-is) becomes useless. We'll need to rewrite and reintroduce macros to simplify and refactor things after it's all finished.
* Yes, I _think_ so, but I might have ran into some kind of multithreading issue with it at the time.

crates/ragu_pcd/src/internal/native/claims.rs
```diff
@@ -69,10 +69,26 @@ pub trait Processor<Rx, AppCircuitId> {
...
+        id: InternalCircuitIndex,
+        groups: impl Iterator<Item = impl Iterator<Item = Rx>>,
+    ) -> Result<()>;
+
+    /// Process a masking claim (fold of rxs, $k(y) = 0$).
+    ///
+    /// Default wraps each rx as a single-element group.
+    fn masking(&mut self, id: InternalCircuitIndex, rxs: impl Iterator<Item = Rx>) -> Result<()> {
+        self.bonding(id, rxs.map(core::iter::once))
+    }
```
TalDerei, line 91: ACK – `masking` folds individual rxs (each rx has it's own slot); `bonding` folds group-sums of rxs (each grouped rx has it's own slot). the zero call sites for specifically the bonding impl implies to me that it'll be used by the copying circuit, which is the shape i'd expect for these kind of copy-style constraints.

minor nit; calling these "masking" and "bonding" feels overloaded and confusing semantically with the property of the rxs it's processing since these are just generic rx claims under the hood. not trying to be pedantic, but i'd replace the naming and doc comments with a more mechanical description of the claim shape and what the methods _actually_ do. reading something like `processor.masking(id, source.rx(Rx(ComputeV)))` is confusing and requires a mental context switch. 

the different processor claims we have are `raw_claim, circuit, internal_circuit, bonding, masking`. I'd tighten these up and align with a `_claim` convention more explicitly. that said, maybe keep bonding since these are technically `revdot(a, sy) = 0` claims.

```ts
  fn raw_claim(&mut self, a, b);                                                                                                                                                                                
  fn circuit_claim(&mut self, app_id, rx);                                                                                                                                                              
  fn internal_circuit_claim(&mut self, id, rxs);                                                                                                                                                              
  fn bonding_claim(&mut self, id, rxs);                                                                                                                                                       
  fn grouped_bonding_claim(&mut self, id, groups);               
```

crates/ragu_pcd/src/fuse/_08_f.rs
```diff
@@ -70,7 +71,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             )?,
         );
 
-        Ok(proof::F { native, bridge })
```
TalDerei, line 73: ACK – stripping ephemerals from proof (child proofs and verifier only reference the f bridge). same pattern for `NativeSPrime`, `NativeInnerError`, `NativeAB::c`, `NativeP::v`.

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -18,13 +17,24 @@ use ragu_circuits::{
...
 
+pub(crate) use builder::ProofBuilder;
+
+/// A newtype marking a field as derived/cacheable.
+///
+/// Wraps a value that can be recomputed from primary proof data. Used to
+/// distinguish commitment caches from primary polynomial fields at the type
+/// level. Immutable once constructed.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct Cached<T>(T);
```
TalDerei, line 36: ACK – starts partially addressing the cached structure in https://github.com/tachyon-zcash/ragu/issues/191.

crates/ragu_pcd/src/fuse/_02_preamble.rs
```diff
@@ -21,52 +24,48 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
-        proof::Preamble<C, R>,
-        native::stages::preamble::Witness<'a, C, R, HEADER_SIZE>,
-    )> {
-        let (native, preamble_witness) =
-            self.compute_native_preamble(rng, left, right, application)?;
+        builder: &mut ProofBuilder<'_, C, R>,
+    ) -> Result<native::stages::preamble::Witness<'a, C, R, HEADER_SIZE>> {
+        let (rx, preamble_witness) = self.compute_native_preamble(rng, left, right, builder)?;
+
+        builder.set_native_preamble_rx(rx);
```
TalDerei, line 31: nit; since we're already passing builder into `compute_native_preamble`, it can internally call `set_native_preamble_rx`.

crates/ragu_pcd/src/fuse/_02_preamble.rs
```diff
@@ -21,52 +24,48 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
+        let bridge_rx = nested::stages::preamble::Stage::<C::HostCurve, R>::rx(
+            C::ScalarField::random(&mut *rng),
+            &nested::stages::preamble::Witness {
+                native_preamble: builder.native_preamble_commitment(),
+                left: nested::stages::preamble::ChildWitness::from_proof(left),
+                right: nested::stages::preamble::ChildWitness::from_proof(right),
+            },
+        )?;
+        let bridge_commitment = bridge_rx.commit_to_affine(C::nested_generators(self.params));
+        builder.set_bridge_preamble_rx(bridge_rx, bridge_commitment);
```
TalDerei, line 42: nit; instead of inlining this, what if we symmetrize by having a `compute_bridge_preamble` that also internally calls `set_bridge_preamble_rx`?

crates/ragu_pcd/src/verify.rs
```diff
@@ -109,28 +109,12 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
-
-        // Check P commitment corresponds to polynomial.
-        let p_commitment_claim = pcd
-            .proof()
-            .p
-            .native
-            .poly
-            .commit_to_affine(C::host_generators(self.params))
-            == pcd.proof().p.native.commitment;
-
```
TalDerei, line 127: ACK – tautological checks removed.

crates/ragu_primitives/src/lib.rs
```diff
@@ -31,15 +31,13 @@ mod simulator;
...
-use io::{Buffer, Write};
-use promotion::Demoted;
-
 pub use boolean::{Boolean, multipack};
 pub use element::{Element, multiadd};
 pub use endoscalar::{Endoscalar, extract_endoscalar, lift_endoscalar};
+use io::{Buffer, Write};
 pub use point::Point;
+use promotion::Demoted;
+use ragu_core::{Result, drivers::Driver, gadgets::Gadget};
```
TalDerei, line 40: nit; breaks import ordering https://github.com/tachyon-zcash/ragu/issues/218.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+                            .as_ref()
+                            .expect(concat!(stringify!($poly), " not set"))
+                            .commit_to_affine(host_gen),
+                    concat!(stringify!($cache), " does not match ", stringify!($poly))
+                );
+            };
+        }
+        verify!(native_a_commitment, native_a_poly);
+        verify!(native_b_commitment, native_b_poly);
+        verify!(native_p_commitment, native_p_poly);
```
TalDerei, line 677: hmm the setter already stored a pre-computed commitment that the fuse pipeline trusts, so these `commit_to_affine` calls seem extraneous.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+
+    // Cached bridge rx + commitments (lazily computed from bridge_alpha + native commitments)
+    bridge_outer_error_rx: Option<sparse::Polynomial<C::ScalarField, R>>,
+    bridge_outer_error_commitment: Option<C::NestedCurve>,
+    bridge_ab_rx: Option<sparse::Polynomial<C::ScalarField, R>>,
+    bridge_ab_commitment: Option<C::NestedCurve>,
+    bridge_query_rx: Option<sparse::Polynomial<C::ScalarField, R>>,
+    bridge_query_commitment: Option<C::NestedCurve>,
+    bridge_eval_rx: Option<sparse::Polynomial<C::ScalarField, R>>,
+    bridge_eval_commitment: Option<C::NestedCurve>,
```
TalDerei, line 262: why aren't cached bridges using `Cell<Option<T>>` like native commitment caches? can probably fold into https://github.com/tachyon-zcash/ragu/pull/638. cc @alxiong

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+        }
+    };
+}
+
+/// Builder for incremental [`Proof`] construction.
+///
+/// Native commitment caches are computed lazily from polynomials on first
+/// access. Special commitments (`a`, `p`) must be provided explicitly because
+/// they are computed via non-standard techniques.
+pub(crate) struct ProofBuilder<'params, C: Cycle, R: Rank> {
```
TalDerei, line 216: ACK – really like this proof builder construction!

crates/ragu_pcd/src/verify.rs
```diff
@@ -79,7 +81,11 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
         // Check all native revdot claims.
         let native_revdot_claims = {
             let ky_source = native::SingleProofKySource {
-                raw_c: pcd.proof().ab.native.c,
+                // NOTE: `raw_c` is now computed as `revdot(a, b)` rather
+                // than stored in the proof, so this claim is tautological
+                // in the verifier. It remains meaningful inside the circuit
+                // where `c` is an independently allocated witness element.
+                raw_c: pcd.proof().c(),
```
TalDerei, line 88: yeah we can drop the tautological `raw_c` check in verify.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+    bridge_pair!(
+        set_bridge_f_rx,
+        bridge_f_commitment,
+        bridge_f_rx,
+        bridge_f_commitment
+    );
+
+    /// Returns the derived alpha for a cached bridge, as a distinct power of
+    /// `bridge_alpha`.
+    fn bridge_alpha_power(&self, idx: nested::RxIndex) -> C::ScalarField {
```
TalDerei, line 507: nit; can probably inline `self.bridge_alpha.pow_vartime` in the `cached_bridge` macro body and delete `bridge_alpha_power`.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+        self.native_p_poly().eval(self.u())
+    }
+
+    /// Build the proof. All polynomial fields must have been set. Commitment
+    /// caches that haven't been accessed yet are computed now.
+    pub(crate) fn build(mut self) -> Result<Proof<C, R>> {
+        let host_gen = C::host_generators(self.params);
+
+        // Ensure all native commitment caches are populated.
+        macro_rules! resolve {
```
TalDerei, line 634: observation; lots of new macros, and intuitively feels like there's room to consolidate and reduce.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -0,0 +1,804 @@
...
+            self.$commitment_field = Some(c);
+            Ok(c)
+        }
+    };
+}
+
+/// Builder for incremental [`Proof`] construction.
+///
+/// Native commitment caches are computed lazily from polynomials on first
+/// access. Special commitments (`a`, `p`) must be provided explicitly because
```
TalDerei, line 214:
```suggestion
/// access. Special commitments (`a`, `b`, `p`) must be provided explicitly because

TalDerei (commented):
reviewed and left some post-hoc observations.

PR #633, opened 2026-04-07
chore: dependency audit 

TalDerei:
`cargo audit` checks deps against https://rustsec.org/ for known CVEs, and came back clean except for one unmaintained dependency: bincode 1.3.3 was flagged as unmaintained ([RUSTSEC-2025-0141](https://github.com/bitcoindevkit/bdk/issues/2092)), pulled in transitively through [gungraun 0.17.0](https://github.com/gungraun/gungraun/releases). 

Inspired by https://github.com/str4d/rage/pull/608, also ran `cargo vet init` to create a `supply-chain/` directory and then `cargo vet`. Here are the findings after multiple iterations of pruning.

-------

Screenshot 2026-04-08 at 8 31 15 AM

--------

Not implying something like this should land, but opening for visibility into our dep graph. cc @ebfull @alxiong


ebfull (changes requested):
Let's move this to `qa/supply-chain`.

TalDerei:
> Let's move this to `qa/supply-chain`.

https://github.com/tachyon-zcash/ragu/pull/633/commits/dd85914b64e967412fcbdc296dde7274914345fa.

noticed that cargo-vet defaults to looking for `supply-chain/` at the workspace root, so we'll need to pass in an explicit path `--store-path qa/supply-chain`.

ebfull (approved):

ebfull (approved):

PR #634, opened 2026-04-08
Cache global mask and projection for bonding polynomials

### LINKED ISSUE #623: Factor global stage mask out of StageMask into Registry
Follow-up for #621's stage mask changes, see also #611.

`StageMask`'s `CircuitObject` impl computes `global(xy) - notch(xy)`, but `global` is invariant across all `StageMask` instances for a given `(R, F)`. When the registry accumulates circuit contributions at a fixed W point, every mask circuit redundantly evaluates the same `global` term.

The registry should own the global mask evaluation and compute it once. Each `StageMask` returns only `-notch` (the per-instance part) and sets a flag on the `CircuitObject` evaluation methods signaling that the global mask should be applied. The registry sums Lagrange coefficients for flagged circuits and multiplies by a single global evaluation.

This also makes the polynomial restriction paths (`sx`/`sy`) cheaper: notch projections touch only `4 * num_gates` wire positions vs `4 * n` for the full mask, so the sparse polynomial operations (`scale`, `add_assign`) during accumulation work on smaller blocks.

**This should not change the circuit digest and is purely an optimization.**

alxiong:
Closes #623 

Factor the invariant $s_{global}(x,y)$ global stage mask out of `StageMask` to `Registry` so that we only have to compute it once for all the stage mask/bonding polynomials, instead of doing them redundantly per stage mask.

Similarly, the projection for `sx()/sy()` is also split into a global projection and per-stage mask notch projection which is the negation of the active regions (since global projection is roughly the diagonal $p^i$ formatted/partially-reversed like a structured vector. Previously, we zero-out the active region on the `View::wiring()`, but here we defer them to registry for negation/offset.

### DIFF
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -207,6 +207,15 @@ pub(crate) trait CircuitObject<F: Field, R: Rank>: Send + Sync {
     /// These records serve as input to [`floor_planner::floor_plan`] for
     /// computing absolute constraint offsets.
     fn segment_records(&self) -> &[SegmentRecord];
+
+    /// Returns `true` if this circuit is a masking polynomial whose
+    /// `sxy`/`sx`/`sy` return only the $-\text{notch}$ term, expecting
+    /// the [`Registry`] to add the shared $S_{\text{global}}$ once.
+    ///
+    /// [`Registry`]: registry::Registry
+    fn is_mask(&self) -> bool {
+        false
+    }
 }
 
 /// Wraps a circuit into a boxed [`CircuitObject`] that can evaluate the
--- a/crates/ragu_circuits/src/lib.rs
+++ b/crates/ragu_circuits/src/lib.rs
@@ -210,7 +210,7 @@ pub(crate) trait CircuitObject<F: Field, R: Rank>: Send + Sync {
 
     /// Returns `true` if this circuit is a masking polynomial whose
     /// `sxy`/`sx`/`sy` return only the $-\text{notch}$ term, expecting
-    /// the [`Registry`] to add the shared $S_{\text{global}}$ once.
+    /// the [`Registry`] to add the shared $S\_{\text{global}}$ once.
     ///
     /// [`Registry`]: registry::Registry
     fn is_mask(&self) -> bool {
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -16,6 +16,7 @@
 //! be efficiently evaluated at different restrictions.
 
 use alloc::{boxed::Box, collections::btree_map::BTreeMap, vec::Vec};
+use core::ops::Range;
 
 use blake2b_simd::Params;
 use ff::{Field, FromUniformBytes, PrimeField};
@@ -177,6 +178,9 @@ impl<'params, F: FromUniformBytes<64>, R: Rank> RegistryBuilder<'params, F, R> {
         let log2_circuits = self.log2_circuits();
         let domain = Domain::<F>::new(log2_circuits);
 
+        let bonding_start = self.internal_circuits.len();
+        let bonding_end = bonding_start + self.bonding.len();
+
         let circuits: Vec<_> = self
             .internal_circuits
             .into_iter()
@@ -214,6 +218,7 @@ impl<'params, F: FromUniformBytes<64>, R: Rank> RegistryBuilder<'params, F, R> {
             circuits,
             floor_plans,
             omega_lookup,
+            bonding_range: bonding_start..bonding_end,
             key: Key::default(),
         };
         registry.key = Key::new(registry.compute_registry_digest());
@@ -306,6 +311,9 @@ pub struct Registry<'params, F: PrimeField, R: Rank> {
     /// Maps from the OmegaKey (which represents some `omega^j`) to the index `i`
     /// of the circuits vector.
     omega_lookup: BTreeMap<OmegaKey, usize>,
+
+    /// Index range of bonding polynomials (stage masks) in `circuits`.
+    bonding_range: Range<usize>,
 
     /// Registry key used to bind circuits to this registry.
     key: Key<F>,
@@ -412,10 +420,22 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
     pub fn xy(&self, x: F, y: F) -> sparse::Polynomial<F, R> {
         let key_scalar = self.key_sxy(x, y);
         let mut coeffs = alloc::vec![F::ZERO; R::num_coeffs()];
+
         for (i, circuit) in self.circuits.iter().enumerate() {
             let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
             coeffs[j] = circuit.sxy(x, y, &self.floor_plans[i]);
         }
+
+        // Bonding polynomials return only -notch; add the shared global
+        // term to each bonding slot to reconstruct the full mask value.
+        if !self.bonding_range.is_empty() {
+            let global_xy = crate::staging::mask::global_mask::<F, R>(x, y);
+            for i in self.bonding_range.clone() {
+                let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
+                coeffs[j] += global_xy;
+            }
+        }
+
         // Convert from the Lagrange basis.
         let domain = &self.domain;
         domain.ifft(&mut coeffs[..domain.n()]);
@@ -505,6 +525,34 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
         result
     }
 
+    /// Sums Lagrange coefficients for bonding polynomials at the given $W$ point.
+    fn bonding_coeff_sum(&self, cache: &LagrangeCache<F>) -> F {
+        if self.bonding_range.is_empty() {
+            return F::ZERO;
+        }
+
+        match cache {
+            LagrangeCache::Interpolate(coeffs) => {
+                let mut sum = F::ZERO;
+                for i in self.bonding_range.clone() {
+                    let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
+                    sum += coeffs[j];
+                }
+                sum
+            }
+            LagrangeCache::Direct(i) => {
+                // W is exactly omega^i, so the Lagrange coefficient for
+                // circuit i is ONE and all others are ZERO.
+                if self.bonding_range.contains(i) {
+                    F::ONE
+                } else {
+                    F::ZERO
+                }
+            }
+            LagrangeCache::Empty => F::ZERO,
+        }
+    }
+
     /// Bind the registry to a specific $W$ point, caching Lagrange coefficients.
     ///
     /// Returns a [`RegistryAt`] that can be used to evaluate the registry
@@ -528,7 +576,7 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
 }
 
 impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
-    /// Evaluate the registry polynomial restricted at $W$, unrestricted at $Y$.
+    /// Evaluate the registry polynomial restricted at $W$ and $Y$, unrestricted at $X$.
     pub fn y(&self, y: F) -> sparse::Polynomial<F, R> {
         let mut poly = self.registry.w_cached(
             &self.cache,
@@ -553,7 +601,7 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
         poly
     }
 
-    /// Evaluate the registry polynomial restricted at $W$, unrestricted at $X$.
+    /// Evaluate the registry polynomial restricted at $W$ and $X$, unrestricted at $Y$.
     pub fn x(&self, x: F) -> sparse::Polynomial<F, R> {
         let mut poly = self.registry.w_cached(
             &self.cache,
@@ -582,13 +630,20 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
 
     /// Evaluate the registry polynomial at the point ($W$, $X$, $Y$).
     pub fn xy(&self, x: F, y: F) -> F {
-        let result: F = self.registry.w_cached(
+        let mut result: F = self.registry.w_cached(
             &self.cache,
             || F::ZERO,
             |circuit, floor_plan, coeff, result| {
                 *result += circuit.sxy(x, y, floor_plan) * coeff;
             },
         );
+
+        // Bonding polynomials return only -notch(x, y). Apply the shared
+        // global stage mask scalar once.
+        if !self.registry.bonding_range.is_empty() {
+            result += self.registry.bonding_coeff_sum(&self.cache)
+                * crate::staging::mask::global_mask::<F, R>(x, y);
+        }
 
         // Add the registry key contribution.
         result + self.registry.key_sxy(x, y)
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -426,8 +426,8 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
             coeffs[j] = circuit.sxy(x, y, &self.floor_plans[i]);
         }
 
-        // Bonding polynomials return only -notch; add the shared global
-        // term to each bonding slot to reconstruct the full mask value.
+        // The sxy evaluation of bonding polynomials returns only -notch;
+        // add the shared global term to each bonding slot.
         if !self.bonding_range.is_empty() {
             let global_xy = crate::staging::mask::global_mask::<F, R>(x, y);
             for i in self.bonding_range.clone() {
@@ -588,6 +588,14 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
             },
         );
 
+        // The sy projection of bonding polynomials returns only -notch(X, y).
+        // Add the shared global once, scaled by the sum of Lagrange coefficients.
+        if !self.registry.bonding_range.is_empty() {
+            let mut global = crate::staging::mask::global_project::<F, R>(y);
+            global.scale(self.registry.bonding_coeff_sum(&self.cache));
+            poly.add_assign(&global);
+        }
+
         // Add the registry key contribution k * (XY)^{4n-1}.  Restricted
         // at Y, this is k * y^{4n-1} at X^{4n-1} (c-wire of the SYSTEM gate in
         // the wiring layout).
@@ -612,6 +620,14 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
                 poly.add_assign(&tmp);
             },
         );
+
+        // The sx projection of bonding polynomials returns only -notch(x, Y).
+        // Add the shared global once, scaled by the sum of Lagrange coefficients.
+        if !self.registry.bonding_range.is_empty() {
+            let mut global = crate::staging::mask::global_project::<F, R>(x);
+            global.scale(self.registry.bonding_coeff_sum(&self.cache));
+            poly.add_assign(&global);
+        }
 
         // Add the registry key contribution k * (XY)^{4n-1}.  Restricted
         // at X, this is k * x^{4n-1} at Y^{4n-1}.
@@ -638,8 +654,8 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
             },
         );
 
-        // Bonding polynomials return only -notch(x, y). Apply the shared
-        // global stage mask scalar once.
+        // The sxy evaluation of bonding polynomials returns only -notch(x, y).
+        // Apply the shared global stage mask scalar once.
         if !self.registry.bonding_range.is_empty() {
             result += self.registry.bonding_coeff_sum(&self.cache)
                 * crate::staging::mask::global_mask::<F, R>(x, y);
--- a/crates/ragu_circuits/src/registry.rs
+++ b/crates/ragu_circuits/src/registry.rs
@@ -337,6 +337,7 @@ enum LagrangeCache<F> {
 pub struct RegistryAt<'a, F: PrimeField, R: Rank> {
     registry: &'a Registry<'a, F, R>,
     cache: LagrangeCache<F>,
+    bonding_coeff_sum: F,
 }
 
 /// Represents a key for identifying a unique $\omega^j$ value where $\omega$ is
@@ -541,8 +542,8 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
                 sum
             }
             LagrangeCache::Direct(i) => {
-                // W is exactly omega^i, so the Lagrange coefficient for
-                // circuit i is ONE and all others are ZERO.
+                // W is exactly omega^bitreverse(i), so the Lagrange
+                // coefficient for circuit i is ONE and all others are ZERO.
                 if self.bonding_range.contains(i) {
                     F::ONE
                 } else {
@@ -568,9 +569,11 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
             // w is in the domain but no circuit registered at that index.
             LagrangeCache::Empty
         };
+        let bonding_coeff_sum = self.bonding_coeff_sum(&cache);
         RegistryAt {
             registry: self,
             cache,
+            bonding_coeff_sum,
         }
     }
 }
@@ -592,7 +595,7 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
         // Add the shared global once, scaled by the sum of Lagrange coefficients.
         if !self.registry.bonding_range.is_empty() {
             let mut global = crate::staging::mask::global_project::<F, R>(y);
-            global.scale(self.registry.bonding_coeff_sum(&self.cache));
+            global.scale(self.bonding_coeff_sum);
             poly.add_assign(&global);
         }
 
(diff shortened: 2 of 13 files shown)


crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -7,6 +7,39 @@ use crate::{
...
 };
 
+/// Evaluates the global stage mask polynomial $S_{\text{global}}(X, Y)$ at
+/// the point $(x, y)$.
+///
+/// The global mask enforces all $4n$ wire slots to zero except the four wires
+/// of the SYSTEM gate (gate 0):
+///
+/// $$S_{\text{global}}(xy) = \sum_{i=0}^{4n-1} (xy)^i
+///     - \bigl((xy)^{2n} + 1\bigr)\bigl((xy)^{2n-1} + 1\bigr)$$
```
TalDerei, line 17: ACK – maximal zeroing mask + carving out the system gate.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -105,36 +138,31 @@ impl<R: Rank> StageMask<R> {
...
+    ///
+    /// The full stage mask is $S_{\text{global}} - \text{notch}$, but
+    /// the invariant $S_{\text{global}}$ term is factored out and applied
+    /// once by the [`Registry`](crate::registry::Registry). This method
+    /// returns only $-\text{notch}$, where
+    ///
+    /// $$\text{notch}(xy) = (1 + (xy)^{2n})\bigl((xy)^g + (xy)^{2n-g-m}\bigr)
+    ///     \cdot \sum_{i=0}^{m-1} (xy)^i$$
+    ///
+    /// with $g = \text{skip\_gates}$ and $m = \text{num\_gates}$.
```
TalDerei, line 187: ACK – notch subtracts out the active-stage gates.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -410,7 +438,9 @@ mod tests {
         let x = Fp::random(&mut rand::rng());
         let y = Fp::random(&mut rand::rng());
 
-        let sxy = stage_mask.sxy(x, y, &[]);
+        // sxy returns -notch; sx/sy return the full mask (global - notch).
+        let global_xy = super::global_mask::<Fp, R>(x, y);
+        let sxy = stage_mask.sxy(x, y, &[]) + global_xy;
```
TalDerei: ACK – global - notch is full stage mask that leaves the active region of the stage mask unconstrained (for the stage trace to populate).

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -40,6 +40,47 @@ pub(crate) fn global_mask<F: Field, R: Rank>(x: F, y: F) -> F {
     geosum(xy, R::n() << 2) - (xy_2n + F::ONE) * (xy_2n * xy_inv + F::ONE)
 }
 
+/// Restricts the global stage mask to a univariate sparse polynomial by
+/// substituting `p` for one variable. Populates `4(n - 1)` wire
+/// positions — all except the four wires of the SYSTEM gate (gate 0).
+///
+/// Used by [`Registry`](crate::registry::Registry) to add the shared global
+/// contribution once, scaled by the sum of bonding Lagrange coefficients.
+pub(crate) fn global_project<F: Field, R: Rank>(p: F) -> sparse::Polynomial<F, R> {
```
TalDerei, line 50: nit; we could minimize the diff by reordering this method down.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -88,49 +129,40 @@ impl<R: Rank> StageMask<R> {
...
-    /// `sx` ($S(x, Y)$) and `sy` ($S(X, y)$). Unconstrained wires
-    /// (the SYSTEM gate and active-stage gates) are zeroed out.
-    fn project<F: Field>(&self, p: F) -> sparse::Polynomial<F, R> {
+    /// Projects only the negated notch entries: the active-stage gate
+    /// wires at positions `skip_gates..skip_gates + num_gates`, negated.
+    /// The SYSTEM gate is excluded (it is already zero in the global mask).
+    ///
+    /// Returns a sparse polynomial with `4 × num_gates` non-zero entries
+    /// (vs `4(n - 1)` for the full mask projection).
+    fn notch_project<F: Field>(&self, p: F) -> sparse::Polynomial<F, R> {
```
TalDerei: ACK – much sparser than populating dense 4n polynomial that then zeroed out entries.

crates/ragu_circuits/src/registry.rs
```diff
@@ -505,6 +525,34 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+            LagrangeCache::Interpolate(coeffs) => {
+                let mut sum = F::ZERO;
+                for i in self.bonding_range.clone() {
+                    let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
+                    sum += coeffs[j];
+                }
+                sum
+            }
+            LagrangeCache::Direct(i) => {
+                // W is exactly omega^i, so the Lagrange coefficient for
```
TalDerei: think the comment means `omega^bitreverse(i)`?

crates/ragu_circuits/src/registry.rs
```diff
@@ -307,6 +312,9 @@ pub struct Registry<'params, F: PrimeField, R: Rank> {
     /// of the circuits vector.
     omega_lookup: BTreeMap<OmegaKey, usize>,
 
+    /// Index range of bonding polynomials (stage masks) in `circuits`.
+    bonding_range: Range<usize>,
```
TalDerei: there's a latent hazard here that deserves a comment; https://github.com/tachyon-zcash/ragu/pull/600 implemented `into_bonding_object` (a builder that takes any `MultiStage` circuit and produces a `BondingObject`) specifically as infrastructure for future routing circuits. the registry assumes everything in `bonding_range` is _currently_ a `StageMask`, but registering a non-mask `BondingObject` will basically add an extra global mask term to that circuit's contribution and corrupt the registry eval.

crates/ragu_circuits/src/staging/mask.rs
TalDerei: should we also add zero check for consistency with `global_mask`?

```ts
    if p == F::ZERO {
        return sparse::Polynomial::default();
    }
```

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -55,102 +129,88 @@ impl<R: Rank> StageMask<R> {
...
-
-        /// Contribution of the `m` active-stage gates starting at gate `g`.
-        /// Subtracted from [`global`] to zero out unconstrained wires.
-        fn notch<F: Field>(xy: F, xy_2n: F, xy_inv: F, g: usize, m: usize) -> F {
-            let gsum = geosum(xy, m);
-            let xy_g = xy.pow_vartime([g as u64]);
-            let xy_h = xy_2n * xy_inv.pow_vartime([(g + m) as u64]);
+        let gsum = geosum(xy, self.num_gates);
+        let xy_g = xy.pow_vartime([self.skip_gates as u64]);
+        let xy_h = xy_2n * xy_inv.pow_vartime([(self.skip_gates + self.num_gates) as u64]);
```
TalDerei: nit; can we use better variable names for `xy_g, xy_h`? also, we can simplify `xy_h` by dropping the inversion

```ts
 let xy_h = xy.pow_vartime([(2 * n - self.skip_gates - self.num_gates) as u64]);

File: crates/ragu_circuits/src/registry.rs [hunk 9]
```diff
@@ -540,6 +588,14 @@ impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
             },
         );
 
+        // The sy projection of bonding polynomials returns only -notch(X, y).
+        // Add the shared global once, scaled by the sum of Lagrange coefficients.
+        if !self.registry.bonding_range.is_empty() {
+            let mut global = crate::staging::mask::global_project::<F, R>(y);
+            global.scale(self.registry.bonding_coeff_sum(&self.cache));
+            poly.add_assign(&global);
```
TalDerei: isn't there a minor optimization here (and other call sites with the same logic)? can't we skip `global_project` when its scaling sum is zero (`bonding_coeff_sum` equals zero), which skips the 4(n-1) field ops + allocation. need to think about situations that'd be applicable.

```suggestion
            let sum = self.registry.bonding_coeff_sum(&self.cache);
            if sum != F::ZERO {
                let mut global = crate::staging::mask::global_project::<F, R>(y);
                global.scale(sum);
                poly.add_assign(&global);
            }
```

crates/ragu_circuits/src/staging/mask.rs
TalDerei, line 1: reviewed the math by hand and looks right, but can we add / expand an existing proptest exercising the global and notch properties in this PR?

crates/ragu_circuits/src/registry.rs
```diff
@@ -412,10 +420,22 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
+
+        // The sxy evaluation of bonding polynomials returns only -notch;
+        // add the shared global term to each bonding slot.
+        if !self.bonding_range.is_empty() {
+            let global_xy = crate::staging::mask::global_mask::<F, R>(x, y);
+            for i in self.bonding_range.clone() {
+                let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
+                coeffs[j] += global_xy;
+            }
+        }
```
TalDerei, line 427: can we fuse this bonding loop with the code above? noticed currently there's two passes over coeffs

```suggestion
        let global_xy = if self.bonding_range.is_empty() {
            F::ZERO
        } else {
            crate::staging::mask::global_mask::<F, R>(x, y)
        };

        for (i, circuit) in self.circuits.iter().enumerate() {
            let j = bitreverse(i as u32, self.domain.log2_n()) as usize;
            let mut v = circuit.sxy(x, y, &self.floor_plans[i]);
            if self.bonding_range.contains(&i) {
                v += global_xy;
            }
            coeffs[j] = v;
        }
```

crates/ragu_circuits/src/registry.rs
```diff
@@ -505,6 +525,34 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
         result
     }
 
+    /// Sums Lagrange coefficients for bonding polynomials at the given $W$ point.
+    fn bonding_coeff_sum(&self, cache: &LagrangeCache<F>) -> F {
```
TalDerei: can we cache `bonding_coeff_sum` on `RegistryAt`?

crates/ragu_circuits/src/registry.rs
TalDerei: I'd also refactor the internals here to use a helper in `crates/ragu_circuits/src/polynomials/sparse/mod.rs`

```ts
pub fn add_assign_scaled(&mut self, other: &Self, scale: F) {
    if bool::from(scale.is_zero()) {
        return;
    }
    self.combine_assign(other, |a, b| *a += *b * scale);
}
```

```ts
if !self.registry.bonding_range.is_empty() {
    let sum = self.registry.bonding_coeff_sum(&self.cache);
    if sum != F::ZERO {
        let global = crate::staging::mask::global_project::<F, R>(y);
        poly.add_assign_scaled(&global, sum);
    }
}
```

TalDerei (commented):
this looks great! noted some suggestions

crates/ragu_circuits/src/registry.rs
alxiong: yeah, actually AI originally wrote it this way, shortcut first on `bonding_range.is_empty()`, then `sum == F::ZERO`, but I deny both. because i know these values are all gonna be non-empty and non-zero. I simply don't see a practical benefit to add a branch in the compiled code. 

I know your protection/shortcut is semantically correct. I'm not religious about my decision here. I'll go with whatever majority decides.

crates/ragu_circuits/src/registry.rs
alxiong: I also considered this, but my intuition is your suggested version has more branch prediction in compile code. (i don't have evidence). 

Thus I decide to go for the two passes (even though I know i recompute `j = bitreverse()` twice. but in my mind, compute is cheap, branching is expensive). 

again happy to change, but just explaining why I did it this way.

crates/ragu_circuits/src/registry.rs
alxiong: yes, this is important. I know this is an implicit assumption I make. 
But I don't know how the routing polynomial (which is the other bonding object) work yet. Do you have suggestion how do refactor this PR to prepare for it? 

cc @ebfull as well

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -7,6 +7,83 @@ use crate::{
...
+///
+/// $$S_{\text{global}}(xy) = \sum_{i=0}^{4n-1} (xy)^i
+///     - \bigl((xy)^{2n} + 1\bigr)\bigl((xy)^{2n-1} + 1\bigr)$$
+///
+/// The polynomial depends only on the product $XY$, so both variables
+/// enter only through $xy = x \cdot y$.
+///
+/// This term is invariant across all [`StageMask`] instances for a given
+/// `(R, F)` — it depends only on `R::n()`. The [`Registry`] computes it once
+/// and scales by the sum of Lagrange coefficients for all bonding polynomials,
```
ebfull:
```suggestion
/// and scales by the sum of Lagrange coefficients for all masking polynomials,
```

ebfull (changes requested):
Bonding polynomials more generally are not always masking polynomials; this PR largely interprets them as the same thing. Masking polynomials will be bonding polynomials of the form `global - notch` and they're notable _kinds_ of bonding polynomials for exactly this reason.

My preference was for this PR to ask `CircuitObject` implementations to tell the registry if they'd like to activate the global mask via a boolean flag or some other interface. The registry can then act appropriately.

TalDerei:
> Bonding polynomials more generally are not always masking polynomials; this PR largely interprets them as the same thing. Masking polynomials will be bonding polynomials of the form `global - notch` and they're notable _kinds_ of bonding polynomials for exactly this reason.

https://github.com/tachyon-zcash/ragu/pull/634#discussion_r3063671974

crates/ragu_circuits/src/registry.rs
alxiong: btw, I fuse the two passes after changing to `CircuitObject::is_mask()` in [4179eec](https://github.com/tachyon-zcash/ragu/pull/634/commits/4179eec627b7af87a966da29ea45a1b849997c5d)

crates/ragu_circuits/src/registry.rs
alxiong: after [4179eec](https://github.com/tachyon-zcash/ragu/pull/634/commits/4179eec627b7af87a966da29ea45a1b849997c5d), we don't have cheap `range.is_empty()`, plus scale by zero is cheap. lmk if you still want to add some empty-guard.

crates/ragu_circuits/src/registry.rs
alxiong: resolved in [4179eec](https://github.com/tachyon-zcash/ragu/pull/634/commits/4179eec627b7af87a966da29ea45a1b849997c5d)

alxiong:
re @ebfull 

Agreed — bonding polynomials aren't necessarily masking polynomials, and the positional `bonding_range` was too implicit. I've added `CircuitObject::is_mask()` (default false, `StageMask` returns true) so circuits self-declare whether they return only `-notch` and expect the Registry to supply the global term. 

This lets us drop `bonding_range` entirely — `mask_coeff_sum` now iterates all circuits checking `is_mask(),` and the IFFT path fuses into a single loop. The `RegistryAt` methods unconditionally add `global_project` scaled by `mask_coeff_sum,` which is zero when no masks are present.

See [4179eec](https://github.com/tachyon-zcash/ragu/pull/634/commits/4179eec627b7af87a966da29ea45a1b849997c5d)

crates/ragu_circuits/src/lib.rs
```diff
@@ -207,6 +207,15 @@ pub(crate) trait CircuitObject<F: Field, R: Rank>: Send + Sync {
...
     fn segment_records(&self) -> &[SegmentRecord];
+
+    /// Returns `true` if this circuit is a masking polynomial whose
+    /// `sxy`/`sx`/`sy` return only the $-\text{notch}$ term, expecting
+    /// the [`Registry`] to add the shared $S_{\text{global}}$ once.
+    ///
+    /// [`Registry`]: registry::Registry
+    fn is_mask(&self) -> bool {
+        false
+    }
```
TalDerei, line 218: suggesting removing default in favor of explicit declarations. 

```ts
fn is_mask(&self) -> bool;
```

and `crates/ragu_circuits/src/lib.rs` would have an explicit false on `Processed` and `crates/ragu_circuits/src/staging/bonding.rs` would have an explicit false on `Stripped`.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -7,6 +7,83 @@ use crate::{
...
+/// This term is invariant across all [`StageMask`] instances for a given
+/// `(R, F)` — it depends only on `R::n()`. The [`Registry`] computes it once
+/// and scales by the sum of Lagrange coefficients for all masking polynomials,
+/// rather than letting each mask evaluate it redundantly.
+///
+/// Returns [`Field::ZERO`] when `x == 0` or `y == 0` (bonding polynomials
+/// have zero constant term in $Y$).
+///
+/// [`Registry`]: crate::registry::Registry
+pub(crate) fn global_mask<F: Field, R: Rank>(x: F, y: F) -> F {
```
TalDerei, line 31: can we benchmark against main? curious what the fuse savings actually are in this PR.

TalDerei (approved):
optimistically 👍

crates/ragu_circuits/src/lib.rs
ebfull: I think letting these defaults be `false` is fine personally.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -7,6 +7,83 @@ use crate::{
     polynomials::{Rank, sparse},
 };
 
+/// Evaluates the global stage mask polynomial $S_{\text{global}}(X, Y)$ at
+/// the point $(x, y)$.
+///
+/// The global mask enforces all $4n$ wire slots to zero except the four wires
+/// of the SYSTEM gate (gate 0):
+///
+/// $$S_{\text{global}}(xy) = \sum_{i=0}^{4n-1} (xy)^i
```
ebfull:
```suggestion
/// $$S_{\text{global}}(x, y) = \sum_{i=0}^{4n-1} (xy)^i
```

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -55,102 +132,87 @@ impl<R: Rank> StageMask<R> {
...
-    /// polynomial by evaluating one variable at `p`. Used by both
-    /// `sx` ($S(x, Y)$) and `sy` ($S(X, y)$). Unconstrained wires
-    /// (the SYSTEM gate and active-stage gates) are zeroed out.
-    fn project<F: Field>(&self, p: F) -> sparse::Polynomial<F, R> {
+    /// Projects only the negated notch entries: the active-stage gate
+    /// wires at positions `skip_gates..skip_gates + num_gates`, negated.
+    /// The SYSTEM gate is excluded (it is already zero in the global mask).
+    ///
+    /// Returns a sparse polynomial with `4 × num_gates` non-zero entries
+    /// (vs `4(n - 1)` for the full mask projection).
```
ebfull: Documentation here is written as if to document the diff rather than the actual method as it's currently implemented. Should be phrased like "Returns the X" or "Computes X given ..." etc.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -55,102 +132,87 @@ impl<R: Rank> StageMask<R> {
...
+        let mut d = p.pow_vartime([g as u64]);
+        let mut a = p.pow_vartime([(2 * n - 1 - g) as u64]);
+        let mut b = p.pow_vartime([(2 * n + g) as u64]);
+        let mut c = p.pow_vartime([(4 * n - 1 - g) as u64]);
+
+        for j in g..g + m {
+            view.d[j] = -d;
+            view.a[j] = -a;
+            view.b[j] = -b;
+            view.c[j] = -c;
```
ebfull: Negations could be folded into the loop; negate each `a/b/c/d` at the start of the loop.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -55,102 +132,87 @@ impl<R: Rank> StageMask<R> {
...
 
 impl<F: Field, R: Rank> CircuitObject<F, R> for StageMask<R> {
+    /// Evaluates the per-instance notch term $-\text{notch}(x, y)$.
+    ///
+    /// The full stage mask is $S_{\text{global}} - \text{notch}$, but
+    /// the invariant $S_{\text{global}}$ term is factored out and applied
+    /// once by the [`Registry`](crate::registry::Registry). This method
+    /// returns only $-\text{notch}$, where
+    ///
+    /// $$\text{notch}(xy) = (1 + (xy)^{2n})\bigl((xy)^g + (xy)^{2n-g-m}\bigr)
```
ebfull:
```suggestion
    /// $$\text{notch}(x, y) = (1 + (xy)^{2n})\bigl((xy)^g + (xy)^{2n-g-m}\bigr)
```

crates/ragu_circuits/src/staging/mask.rs
alxiong: ofc! but I will need to add new benchmark, because current `ragu_circuit/benches` only registered non-bonding polynomials. And I will need the registry with proper multi-stage internal circuits, thus I plan to place this bench in `ragu_pcd` instead.

alxiong:
@TalDerei 
Here's the bench against `main` on the criterion bench introduced in [2de48f0](https://github.com/tachyon-zcash/ragu/pull/634/commits/2de48f04e316bc74bf75873949e915ad324890cd):

```
registry::finalize      time:   [55.729 ms 55.778 ms 55.893 ms]
                        change: [-2.3113% -1.2927% -0.2567%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe

registry::wx            time:   [4.8546 ms 4.8783 ms 4.9012 ms]
                        change: [-25.266% -24.513% -23.884%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high severe

registry::wy            time:   [5.0927 ms 5.1095 ms 5.1402 ms]
                        change: [-23.360% -22.831% -22.356%] (p = 0.00 < 0.05)
                        Performance has improved.

registry::wxy           time:   [2.2267 ms 2.2275 ms 2.2286 ms]
                        change: [-3.7137% -3.4657% -3.1176%] (p = 0.00 < 0.05)
                        Performance has improved.
```

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -7,6 +7,84 @@ use crate::{
...
+}
+
+/// Computes the polynomial restriction $S\_{\text{global}}(p, Y)$ (equivalently
+/// $S\_{\text{global}}(X, p)$ by symmetry) as a sparse univariate polynomial.
+/// Populates `4(n - 1)` wire positions — all except the four wires of the
+/// SYSTEM gate (gate 0).
+///
+/// Used by [`Registry`](crate::registry::Registry) to add the shared global
+/// contribution once, scaled by the sum of masking circuit Lagrange coefficients.
+pub(crate) fn global_project<F: Field, R: Rank>(p: F) -> sparse::Polynomial<F, R> {
```
ebfull, line 50: Note for future purposes: we should _someday_ make available to `CircuitObject::sx`/`CircuitObject::sy` a precomputed power vector of `x` or `y` since we must compute them all here anyway.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -55,102 +133,87 @@ impl<R: Rank> StageMask<R> {
...
+    /// $$\text{notch}(x, y) = (1 + (xy)^{2n})\bigl((xy)^g + (xy)^{2n-g-m}\bigr)
+    ///     \cdot \sum_{i=0}^{m-1} (xy)^i$$
+    ///
+    /// with $g = \text{skip\_gates}$ and $m = \text{num\_gates}$.
     fn sxy(&self, x: F, y: F, _floor_plan: &[crate::floor_planner::ConstraintSegment]) -> F {
         if x == F::ZERO || y == F::ZERO {
-            // If either x or y is zero, the polynomial evaluates to zero
-            // (the constant term of a bonding polynomial is always zero).
             return F::ZERO;
         }
```
ebfull, line 191: What would this function do if this check were not present? I suspect it would return the correct value anyway.

crates/ragu_circuits/src/registry.rs
```diff
@@ -520,15 +558,17 @@ impl<F: PrimeField, R: Rank> Registry<'_, F, R> {
...
             registry: self,
             cache,
+            mask_coeff_sum,
         }
     }
 }
 
 impl<F: PrimeField, R: Rank> RegistryAt<'_, F, R> {
-    /// Evaluate the registry polynomial restricted at $W$, unrestricted at $Y$.
+    /// Evaluate the registry polynomial restricted at $W$ and $Y$, unrestricted at $X$.
```
ebfull, line 571: Hm, nice catch.

alxiong: yeah, there are two such cases i recall, when I try to modify them, and saw the typo in the rustdoc.

crates/ragu_circuits/src/staging/mask.rs
alxiong: it would still be correct, this only shortcuts the rest, purely performance without affecting correctness.
do you still want to preserve this comment?

ebfull (approved):

PR #635, opened 2026-04-10
book: improve clarity on new wiring poly layout

alxiong:
Main two clarifications:
- explain more clearly how (and why) the global mask works. with back-and-forth description of the masking polynomial, and its wiring matrix.
  - also explicitly mention "notch poly"
- clarify the `kappa` in the masking polynomial

### DIFF
--- a/book/src/protocol/local/overview.md
+++ b/book/src/protocol/local/overview.md
@@ -114,12 +114,10 @@ coefficient for $z^0$, where the combined check yields
 $$
 c_0 d_0 + \revdot{\v{r}}{\v{s}} = \dot{\v{k}}{\v{y^{4n}}}.
 $$
-Meanwhile, the $z^{4n - 1}$ coefficient gives
-$$
-\v{d}_0 \v{c}_0 = 0
-$$
-and substituting into the $z^0$ coefficient gives
+Meanwhile, the $z^{4n - 1}$ term independently enforces $d_0 c_0 = 0$.
+Thus, the $z^0$ coefficient equation above simplifies to
 $\revdot{\v{r}}{\v{s}} = \dot{\v{k}}{\v{y^{4n}}}$, which is exactly the
-constraint check. In any case, Ragu defines $\v{s}_0 = 0$ for all circuits (the
-$d_0$ wire does not participate in any constraints), so this boundary condition
-is only theoretical.
+constraint check. In any case, Ragu defines
+$\v{s}_0 = \sum_j y^j\cdot \v{s}_{j,0} = 0$
+for all circuits (the $d_0$ wire does not participate in any constraints),
+so this boundary condition is only theoretical.
--- a/book/src/protocol/local/wiring.md
+++ b/book/src/protocol/local/wiring.md
@@ -61,6 +61,8 @@ wiring polynomials, they must contain the $\kappa$ constraint.
 
 These polynomials are exclusively used in revdots of the form
 $\revdot{\v{r}}{\v{s}} = 0$.
+Revdot checks against a shared bonding polynomial can be batched using a
+challenge $z$: $\revdot{\sum_i z^i \v{r}_i}{\v{s}} = 0$.
 
 ### Masking Polynomials
 
@@ -68,28 +70,43 @@ Masking polynomials are bonding polynomials that are used to enforce that
 partial trace polynomials (stages) only contain assignments at designated
 positions.
 
-We could naively define the simplest possible masking polynomial $\sum_i
-(XY)^i$, since this could be used to enforce a (partial) trace contains no wire
-assignments, and we could define all masking polynomials as a difference between
-this global mask and a sparse polynomial. However, we must satisfy $s(X, 0) = 0$
-for bonding polynomials.
-
-Thus, we could define the global mask
-
+We start with a masking polynomial that enforces the trace polynomial to be the
+zero polynomial (i.e. all wire assignments are zeroes):
 
 $$
-s_\text{global}(X, Y) = \sum_{i=0}^{4n - 1} (XY)^i - \left((XY)^{2n} + 1\right)\left((XY)^{2n-1} + 1\right)
+s_{all}(X, Y) = \sum_i X^i Y^{\sigma(i)}
 $$
 
-which enforces that every wire _except_ the four wires of the SYSTEM gate
-(gate 0) are zero. The SYSTEM gate wires are unconstrained by the mask:
-$b_0$ carries the constant $1$ (the `ONE` wire), $d_0$ carries an arbitrary
-blinding factor $\alpha$, and $a_0$, $c_0$ are zero but left unconstrained.
+where $\sigma$ is a permutation. The resulting wiring layout matrix is a
+_permutation matrix_ where exactly one entry of each row and each columns is $1$
+while the rest is all zeroes. Subsequently,
+$\revdot{\v{r}}{\v{s}} = \sum_i \rv{r}_i \cdot y^{\sigma(i)} = 0$
+enforces $\v{r} = \v{0}$.
+To allow active sub-regions of the trace, we simply drop $X^i Y^{\sigma(i)}$
+terms with the corresponding $X^i$ from the mask. This zeroes out the rows
+of the permutation matrix, leaving those trace positions unconstrained with any
+possible wire assignments. For example, unmasking/activating $\v{b}_{1\ldots 5}$
+region requires $s_{all} -\sum_{i\in [2n+1, 2n+5]} X^i Y^{\sigma(i)}$.
+
+The simplest permutation is $\sigma(i) = i$ which renders a masking polynomial
+$\sum_i (XY)^i$ and the wiring matrix as the anti-diagonal permutation matrix.
+All masking polynomials must satisfy $s(X, 0) = 0$ which leaves the four wires of
+the SYSTEM gate (gate 0) unconstrained. Furthermore, like any wiring polynomial,
+the registry key $\kappa$ needs to be injected in the last (reserved) constraint.
+Thus, we could define a **global mask polynomial**
+
+$$
+\begin{aligned}
+s_\text{global}(X, Y)
+&= \sum_{i=0}^{4n - 1} (XY)^i - \sum_{j\in\{0, 2n-1, 2n, 4n-1 \}} (XY)^j &+ \kappa\cdot(XY)^{4n-1} \\
+&= \sum_{i=0}^{4n - 1} (XY)^i - \left((XY)^{2n} + 1\right)\left((XY)^{2n-1} + 1\right) &+ \kappa\cdot(XY)^{4n-1}
+\end{aligned}
+$$
 
 Given $g = \text{skip\_gates}$ (the starting active gate index) and
 $m = \text{num\_gates}$ (the number of active gates in the stage), we can define
 a specific stage polynomial's masking polynomial as $s_\text{global}$ subtracted
-by
+by a **notch polynomial**
 
 $$
 \sum\limits_{i=0}^{m - 1} \left( (XY)^{g + i} + (XY)^{2n - 1 - g - i} + (XY)^{2n + g + i} + (XY)^{4n - 1 - g - i} \right)


book/src/protocol/local/overview.md
```diff
@@ -114,12 +114,10 @@ coefficient for $z^0$, where the combined check yields
...
+Meanwhile, the $z^{4n - 1}$ term independently enforces $d_0 c_0 = 0$.
+Thus, the $z^0$ coefficient equation above simplifies to
 $\revdot{\v{r}}{\v{s}} = \dot{\v{k}}{\v{y^{4n}}}$, which is exactly the
-constraint check. In any case, Ragu defines $\v{s}_0 = 0$ for all circuits (the
-$d_0$ wire does not participate in any constraints), so this boundary condition
-is only theoretical.
+constraint check. In any case, Ragu defines
+$\v{s}_0 = \sum_j y^j\cdot \v{s}_{j,0} = 0$
+for all circuits (the $d_0$ wire does not participate in any constraints),
+so this boundary condition is only theoretical.
```
TalDerei, line 123: I'd be even more explicit here to avoid conflating distinct facts, grounding each one in the layout table:

```suggestion
Meanwhile, the $z^{4n - 1}$ term independently enforces $d_0 c_0 = 0$. Substituting into the $z^0$ coefficient leaves $\revdot{\v{r}}{\v{s}} = \dot{\v{k}}{\v{y^{4n}}}$, which is exactly the constraint check. In practice this rescue is redundant: the layout fixes $c_0 = 0$, so $d_0 c_0 = 0$ holds trivially, regardless of the blinding value $d_0 = \alpha$. Separately, no circuit constraint touches $d_0$, since the top row of the wiring column in the layout table is empty at every $y^j$ except $j = 4n-1$, which carries only the registry key $\kappa$.

Comment (TalDerei) [reply book/src/protocol/local/overview.md:123]: another idea; similar to the wiring layout table, it'd be nice to visualize how the two independent batching challenges y and z organize the combined check into a diagram / grid, where gate equations run down the y^0 column as powers of z, and constraints run across the z^0 row as powers of y.

File: book/src/protocol/local/wiring.md [hunk 2]
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
 
 Masking polynomials are bonding polynomials that are used to enforce that
 partial trace polynomials (stages) only contain assignments at designated
 positions.
 
-We could naively define the simplest possible masking polynomial $\sum_i
-(XY)^i$, since this could be used to enforce a (partial) trace contains no wire
-assignments, and we could define all masking polynomials as a difference between
-this global mask and a sparse polynomial. However, we must satisfy $s(X, 0) = 0$
-for bonding polynomials.
```
TalDerei, line 75: I'd maintain that definitionally the equation defines the simplest possible masking polynomial. I'd probably keep most of this and just reword in the present tense assuming https://github.com/tachyon-zcash/ragu/pull/634 lands.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
+$\revdot{\v{r}}{\v{s}} = \sum_i \rv{r}_i \cdot y^{\sigma(i)} = 0$
+enforces $\v{r} = \v{0}$.
+To allow active sub-regions of the trace, we simply drop $X^i Y^{\sigma(i)}$
+terms with the corresponding $X^i$ from the mask. This zeroes out the rows
+of the permutation matrix, leaving those trace positions unconstrained with any
+possible wire assignments. For example, unmasking/activating $\v{b}_{1\ldots 5}$
+region requires $s_{all} -\sum_{i\in [2n+1, 2n+5]} X^i Y^{\sigma(i)}$.
+
+The simplest permutation is $\sigma(i) = i$ which renders a masking polynomial
+$\sum_i (XY)^i$ and the wiring matrix as the anti-diagonal permutation matrix.
```
TalDerei, line 92: took me some time to understand why you called this "anti-diagonal" given $\sigma(i) = i$, but realized we're trying to match the degree convention of our layout table. suggesting we be more specific since under standard linear algebra conventions, you'd expect $\sigma(i) = i$ to give the "main diagonal":

"the simplest permutation is $\sigma(i) = i$, giving the masking polynomial $\sum_i (XY)^i$. On the layout table, which draws high X at the top and low Y at the left, the resulting pattern visually traces an anti-diagonal from ($X^{0}$, $Y^{0}$) at the bottom-left to ($X^{(4n-1)}$, $Y^{(4n-1)}$) at the top-right.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
+the registry key $\kappa$ needs to be injected in the last (reserved) constraint.
+Thus, we could define a **global mask polynomial**
 
 $$
-s_\text{global}(X, Y) = \sum_{i=0}^{4n - 1} (XY)^i - \left((XY)^{2n} + 1\right)\left((XY)^{2n-1} + 1\right)
+\begin{aligned}
+s_\text{global}(X, Y)
+&= \sum_{i=0}^{4n - 1} (XY)^i - \sum_{j\in\{0, 2n-1, 2n, 4n-1 \}} (XY)^j &+ \kappa\cdot(XY)^{4n-1} \\
+&= \sum_{i=0}^{4n - 1} (XY)^i - \left((XY)^{2n} + 1\right)\left((XY)^{2n-1} + 1\right) &+ \kappa\cdot(XY)^{4n-1}
+\end{aligned}
```
TalDerei, line 103: I'd revert to the original equation modeling only a pure global mask minus the carve-out for the system gate. assuming the registry key was added here purely pedagogically, but that's not what the code does. inspecting our `global_mask` method, it uses the original formula, and the registry key is added separately by `key_sxy` once at the registry level (added to registry output, not to `s_global`). I'd maintain the separation of concerns.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
-which enforces that every wire _except_ the four wires of the SYSTEM gate
-(gate 0) are zero. The SYSTEM gate wires are unconstrained by the mask:
-$b_0$ carries the constant $1$ (the `ONE` wire), $d_0$ carries an arbitrary
-blinding factor $\alpha$, and $a_0$, $c_0$ are zero but left unconstrained.
-
 Given $g = \text{skip\_gates}$ (the starting active gate index) and
 $m = \text{num\_gates}$ (the number of active gates in the stage), we can define
 a specific stage polynomial's masking polynomial as $s_\text{global}$ subtracted
-by
+by a **notch polynomial**
```
TalDerei, line 109: I'd be interesting to visually display these notch holes in the matrix with some kind of diagram.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
+where $\sigma$ is a permutation. The resulting wiring layout matrix is a
+_permutation matrix_ where exactly one entry of each row and each columns is $1$
+while the rest is all zeroes. Subsequently,
+$\revdot{\v{r}}{\v{s}} = \sum_i \rv{r}_i \cdot y^{\sigma(i)} = 0$
+enforces $\v{r} = \v{0}$.
+To allow active sub-regions of the trace, we simply drop $X^i Y^{\sigma(i)}$
+terms with the corresponding $X^i$ from the mask. This zeroes out the rows
+of the permutation matrix, leaving those trace positions unconstrained with any
+possible wire assignments. For example, unmasking/activating $\v{b}_{1\ldots 5}$
+region requires $s_{all} -\sum_{i\in [2n+1, 2n+5]} X^i Y^{\sigma(i)}$.
```
TalDerei, line 89: Is this range right? this is the b-block

```suggestion
possible wire assignments. For example, unmasking/activating $\v{b}_{1\ldots 5}$
region requires $s_{all} -\sum_{i\in [2n-6, 2n-2]} X^i Y^{\sigma(i)}$.

File: book/src/protocol/local/wiring.md [hunk 7]
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
+
+The simplest permutation is $\sigma(i) = i$ which renders a masking polynomial
+$\sum_i (XY)^i$ and the wiring matrix as the anti-diagonal permutation matrix.
+All masking polynomials must satisfy $s(X, 0) = 0$ which leaves the four wires of
+the SYSTEM gate (gate 0) unconstrained. Furthermore, like any wiring polynomial,
```
TalDerei, line 94: this phrasing feels like a non-sequitur. 

bonding polynomials must satisfy $s(X, 0) = 0$ (i'd call this some kind of like bonding-guard property), and then separately we have the system gate carve out (in the global mask equation where we subtract diagonal entries). 

I'd clarify along the lines of: "all masking polynomials must satisfy the bonding-guard property $s(X, 0) = 0$. Separately, Ragu unmasks the four wires of the system gate by subtracting the four diagonal entries $(XY)^0$, $(XY)^{2n-1}$, $(XY)^{2n}$, and $(XY)^{4n-1}$` from the mask."

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
 
 These polynomials are exclusively used in revdots of the form
 $\revdot{\v{r}}{\v{s}} = 0$.
+Revdot checks against a shared bonding polynomial can be batched using a
+challenge $z$: $\revdot{\sum_i z^i \v{r}_i}{\v{s}} = 0$.
```
TalDerei, line 65: ACK – batching via bilinearity of revdot.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
-blinding factor $\alpha$, and $a_0$, $c_0$ are zero but left unconstrained.
-
 Given $g = \text{skip\_gates}$ (the starting active gate index) and
 $m = \text{num\_gates}$ (the number of active gates in the stage), we can define
 a specific stage polynomial's masking polynomial as $s_\text{global}$ subtracted
-by
+by a **notch polynomial**
 
 $$
 \sum\limits_{i=0}^{m - 1} \left( (XY)^{g + i} + (XY)^{2n - 1 - g - i} + (XY)^{2n + g + i} + (XY)^{4n - 1 - g - i} \right)
```
TalDerei, line 112: after this equation, i'd add a motivating narrative 

```suggestion
\sum\limits_{i=0}^{m - 1} \left( (XY)^{g + i} + (XY)^{2n - 1 - g - i} + (XY)^{2n + g + i} + (XY)^{4n - 1 - g - i} \right)

Because $s_\text{global}$ is identical across all bonding polynomials in a
registry, Ragu computes it only once per evaluation. Each stage mask
contributes only its $-\text{notch}$ term, and the registry adds back the
shared $s_\text{global}$ scaled by the sum of Lagrange coefficients over the
bonding polynomials. This avoids redundant per-stage work.
```

TalDerei (commented):
left some observations

book/src/protocol/local/overview.md
```diff
@@ -114,12 +114,10 @@ coefficient for $z^0$, where the combined check yields
 $$
 c_0 d_0 + \revdot{\v{r}}{\v{s}} = \dot{\v{k}}{\v{y^{4n}}}.
 $$
-Meanwhile, the $z^{4n - 1}$ coefficient gives
-$$
-\v{d}_0 \v{c}_0 = 0
-$$
-and substituting into the $z^0$ coefficient gives
+Meanwhile, the $z^{4n - 1}$ term independently enforces $d_0 c_0 = 0$.
+Thus, the $z^0$ coefficient equation above simplifies to
```
ebfull, line 118: Technically correct, but imprecise in a way that obscures the argument. This passage is just algebra that shows the two equations can be safely combined. The protocol-level language like "independently enforces" re-attributes meaning to the individual coefficients, when we only need to be mechanical here.

book/src/protocol/local/wiring.md
```diff
@@ -61,35 +61,52 @@ wiring polynomials, they must contain the $\kappa$ constraint.
...
-Thus, we could define the global mask
+$$
+s_{all}(X, Y) = \sum_i X^i Y^{\sigma(i)}
+$$
 
+where $\sigma$ is a permutation. The resulting wiring layout matrix is a
+_permutation matrix_ where exactly one entry of each row and each columns is $1$
+while the rest is all zeroes. Subsequently,
+$\revdot{\v{r}}{\v{s}} = \sum_i \rv{r}_i \cdot y^{\sigma(i)} = 0$
+enforces $\v{r} = \v{0}$.
```
ebfull, line 84: My goal is to keep these parts of the explanation as simple as possible, for personal preference, but also to minimize the context bloat for LLMs that need to quickly parse and reason about the math.

I'm concerned that the permutation stuff isn't buying us much explanation power here. It's useful for understanding how I arrived at the specific masking polynomial, since it generalizes, but beyond that it makes everything seem convoluted.

Explain that the strictest masking polynomial would be one that enforces each element of the trace is zero. Give $\sum_i (XY)^i$ as an _example_ of one that does this with a conveniently simple closed form. Then explain that all masking polynomials can be described as the difference between the strictest one and a (very sparse) polynomial that disables the specific "active" portions of the stage, which tend to be significantly fewer than the size of the traces. This latter polynomial _also_ has a simple closed form solution.

ebfull (changes requested):

alxiong:
gauging from feedback so far, my attempt didn't achieve net positive, thus closing this PR. 

But to clarify for @TalDerei, what I meant by anti-diagonal (at least how i visualize it) is the wiring matrix for these bonding polynomials are as follows:
(i break them up into 4 tables because otherwise it's hard to squeeze in the full-width and still look square-ish) 
(but you can clearly see the anti-diagnoal `1` from $(X^0, Y^0)$ all the way to $(X^{4n-1}, Y^{4n-1})$ sans the rows for SYSTEM gate, and sans the active regions.) 

Screenshot 2026-04-10 at 10 44 35 PM

I didn't include this in my PR because I feel this takes too much space and text description is clear enough. But I conceded that the original description is already quite clear.

PR #639, opened 2026-04-11
chore: loading / copying preliminaries suggestions

TalDerei:
Addressing my own feedback in https://github.com/tachyon-zcash/ragu/pull/632.

### DIFF
--- a/crates/ragu_pcd/src/lib.rs
+++ b/crates/ragu_pcd/src/lib.rs
@@ -34,8 +34,6 @@ mod verify;
 use alloc::collections::BTreeMap;
 use core::{any::TypeId, cell::OnceCell, marker::PhantomData};
 
-use header::Header;
-pub use proof::{Pcd, Proof};
 use ragu_arithmetic::Cycle;
 use ragu_circuits::{
     polynomials::Rank,
@@ -43,7 +41,11 @@ use ragu_circuits::{
 };
 use ragu_core::{Error, Result};
 use rand::CryptoRng;
+
+use header::Header;
 use step::{Step, internal::adapter::Adapter};
+
+pub use proof::{Pcd, Proof};
 
 /// Domain separation tag for Ragu PCD protocol.
 // FIXME: choose a permanent domain separation tag before release.
--- a/crates/ragu_pcd/src/proof/builder.rs
+++ b/crates/ragu_pcd/src/proof/builder.rs
@@ -666,17 +666,6 @@ impl<'params, C: Cycle, R: Rank> ProofBuilder<'params, C, R> {
         macro_rules! cached {
             ($field:ident) => {
                 Cached(
-                    *self
-                        .$field
-                        .get()
-                        .expect(concat!(stringify!($field), " not set")),
-                )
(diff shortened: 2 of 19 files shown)


crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -616,52 +630,21 @@ impl<'params, C: Cycle, R: Rank> ProofBuilder<'params, C, R> {
...
-                            .as_ref()
-                            .expect(concat!(stringify!($poly), " not set"))
-                            .commit_to_affine(host_gen),
-                    concat!(stringify!($cache), " does not match ", stringify!($poly))
-                );
-            };
-        }
-        verify!(native_a_commitment, native_a_poly);
-        verify!(native_b_commitment, native_b_poly);
-        verify!(native_p_commitment, native_p_poly);
```
TalDerei, line 664: references https://github.com/tachyon-zcash/ragu/pull/632#discussion_r3068085377. these could optionally be made into debug assertions, but this macro feels more ceremonial than necessary.

crates/ragu_pcd/src/lib.rs
```diff
@@ -34,17 +34,19 @@ mod verify;
...
     polynomials::Rank,
     registry::{Registry, RegistryBuilder},
 };
 use ragu_core::{Error, Result};
 use rand::CryptoRng;
+
+use header::Header;
 use step::{Step, internal::adapter::Adapter};
 
+pub use proof::{Pcd, Proof};
```
TalDerei, line 48: realized import ordering will be churn on next nightly-rustfmt, so we should have some kind of skill for https://github.com/tachyon-zcash/ragu/issues/218.

crates/ragu_pcd/src/verify.rs
```diff
@@ -78,21 +78,21 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
-                // where `c` is an independently allocated witness element.
-                raw_c: pcd.proof().c(),
                 application_ky,
                 unified_bridge_ky,
                 unified_ky,
             };
 
             native::ky_values(&ky_source)
                 .zip(builder.a.iter().zip(builder.b.iter()))
+                .skip(1)
```
ebfull: This creates a hidden dependency between the claim `build` code that dictates the order of the claims, and this code which picks and chooses which ones to enforce. I thought about it and couldn't think of another option besides leaving it the way it is for now and putting a TODO. I would much rather do that...

ebfull (changes requested):

crates/ragu_pcd/src/verify.rs
TalDerei: yeah can't think of a way around this dependency besides manually enforcing that `raw_c` is always the first claim in the builder; reverted to the inline note in https://github.com/tachyon-zcash/ragu/pull/639/commits/d885236f59f00dd44f4ed06347ca244f0e2bc599.

ebfull (approved):

PR #642, opened 2026-04-13
Clean integration

mitschabaude:


TalDerei:
@mitschabaude assuming the CI pipeline integration will land in a follow-up PR?

TalDerei:
can we migrate this under the `qa/` subdirectory?

fv/crates/lean_extraction/src/codegen.rs
```diff
@@ -0,0 +1,123 @@
...
+            } else {
+                format!("(var {})", i - 1)
+            }
+        }
+        Expr::InputVar(i) => format!("(input_var.get {i})"),
+        Expr::Const(c) => display_coeff(c),
+        Expr::Add(l, r) => format!("({} + {})", display_expr(l), display_expr(r)),
+        Expr::Mul(l, r) => format!("({} * {})", display_expr(l), display_expr(r)),
+    }
+}
```
TalDerei, line 48: Interesting, during the extraction driver lifetime (codegen), this unconditionally strips the one wire. If we eventually extend this to other parts of the code that _already_ have their one wire stripped, would need to think if there's any downstream side-effects. intuitively feels like it needs some kind of runtime gaurd.

fv/lean/Ragu/Instances/Autogen/Point/AddIncomplete.lean
TalDerei, line 1: recalling my holistic understanding from our last sync; a rust [circuit instance](https://github.com/tachyon-zcash/ragu/pull/642/changes#diff-399dd1fa057d5a2e3dbb5f791c6a762de240da8b043cd2b4d6d0b1ce74774769) (thin wrapper that internally calls ragu types / gadgets) goes through the extractor (extraction driver) to produce this autogen (simply prints lean flat op trace) with exported ops and output. 

then the [reimplementation](https://github.com/tachyon-zcash/ragu/pull/642/changes#diff-eabde7396396229234c50c14b8a8179ac8c4da0396050d61fdc950e43d88f838) imports the autogen, deserializes input / serializes output, picks a reimplementation (structured clean circuit in `GeneralCircuitForm` packaged with assumptions / spec / soundness / completeness definitions). it "reimplements" ragu's gadget circuit logic (witness alloc, constraints, output).

then [FormalInstance](https://github.com/tachyon-zcash/ragu/pull/642/changes#diff-8a85f4b26a509f495d681b9de57c540d5f693bd8442f4f63dc7b022e49a8d369)  checks that autogen and reimplementation match.

in terms of the trusted / untrusted objects, the untrusted boundary interface there starts at the reimplementation's soundness proof (and everything below like the `FormalInstance`-level invariants).

fv/crates/lean_extraction/src/instances/point_alloc.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    fn circuit(dr: &mut ExtractionDriver<Fp>) -> ragu_core::Result<Vec<Expr<Fp>>> {
+        // MaybeKind = Empty: the closure is never called.
+        let assignment = ExtractionDriver::<Fp>::just(|| Fp::ZERO);
+        let point = Point::<_, EpAffine>::alloc(dr, assignment)?;
+
+        // NOTE: assumes that the serialization is [x, y].
+        // TODO: This is an assumption we should not make in general, and would be better if we "manually"
+        // serialize the output into a Vector. However, Point wires are private, so this is the only way
+        // for now
```
TalDerei, line 23: recalling there were some open questions around there being no canonical serialization/deserialization of gadget wire orderings, so you implemented a hacky workaround for now. is there a suggested path forward?

fv/lean/Ragu/Circuits/Point/AddIncomplete.lean
```diff
@@ -0,0 +1,144 @@
...
+    simp_all only [id_eq, inv_pow, and_self]
+
+  · simp_all only [id_eq, add_neg_cancel, mul_zero, implies_true, zero_mul, mul_eq_zero, false_or,
+    true_and]
+    rw [add_neg_eq_zero]
+    intro h1 _
+    apply Ne.symm
+    exact h1
+
+theorem completeness (curveParams : Spec.CurveParams p)
```
TalDerei, line 123: last sync, we discussed completeness being tricky since witness gen was hard to extract from rust + the FFI complexity (ie. the honest witness function needs to be written in lean, but real witness gen is in rust and can use arbitrary computation, so there’s a coverage gap there). we then discussed some lean-native witness implementation matched against a Rust impl. even if that's not implemented, and I haven't traced deeply enough yet to know, what's the state of the completeness property? noticing that completeness is proved at the reimplementation layer?

.github/workflows/fv.yml
TalDerei, line 1: ACK – CI pipeline calls lake build.

fv/lean/Ragu/Circuits/Point/AddIncomplete.lean
mitschabaude: Yep our clean circuits define honest witness generators and we prove they are complete = satisfy the constraints

mitschabaude: It doesn't say anything about the Rust witness gen, it's a caveat but IMO good enough

fv/crates/lean_extraction/src/instances/point_alloc.rs
mitschabaude: I'll discuss with @gio54321 again but in my estimation, this is fine as the serialization order is not trusted. Constraints are associated with particular wires. If by accident, we would talk about the output wires in the wrong order on the lean side, we simply wouldn't be able to prove circuit equivalence.

In practice, both Ragu and clean use derivers that will map struct fields to wires in field definition order. This code uses the derived order which seems like the right approach to me.

fv/crates/lean_extraction/src/codegen.rs
mitschabaude: I wonder if it wouldn't be better if we would extract it as is. In Lean, that would probably mean making the one wire an input to gadgets that use it, and adding the assumption that it equals 1. Which seems appropriate since that's exactly what ragu circuits do?

qa/crates/lean_extraction/src/codegen.rs
```diff
@@ -0,0 +1,123 @@
...
+            } else {
+                format!("(var {})", i - 1)
+            }
+        }
+        Expr::InputVar(i) => format!("(input_var.get {i})"),
+        Expr::Const(c) => display_coeff(c),
+        Expr::Add(l, r) => format!("({} + {})", display_expr(l), display_expr(r)),
+        Expr::Mul(l, r) => format!("({} * {})", display_expr(l), display_expr(r)),
+    }
+}
```
mitschabaude: discussed this and decided it's equivalent and simpler to just make `Driver::ONE = Expr::const(1)`

TalDerei (approved):
looks good on passing CI!

@mitschabaude observation; noticed the field arithmetic primitives for element have lean reimplementations, but no rust circuit instance, so they're never run through the extractor? seems fine since they appear to be used as subcircuits for other gadgets. 

update: clarified OOB that's the case; a natural entry point for extending this FV impl.

TalDerei:
> looks good on passing CI!

contingent also on migration under `qa/` subdirectory.

qa/crates/lean_extraction/src/instances/point_alloc.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    fn circuit(dr: &mut ExtractionDriver<Fp>) -> ragu_core::Result<Vec<Expr<Fp>>> {
+        // MaybeKind = Empty: the closure is never called.
+        let assignment = ExtractionDriver::<Fp>::just(|| Fp::ZERO);
+        let point = Point::<_, EpAffine>::alloc(dr, assignment)?;
+
+        // NOTE: assumes that the serialization is [x, y].
+        // TODO: This is an assumption we should not make in general, and would be better if we "manually"
+        // serialize the output into a Vector. However, Point wires are private, so this is the only way
+        // for now
```
mitschabaude: Discussed this. There are two things to mention, cc @TalDerei @ebfull

- we believe this would be somewhat mitigated by having a canonical serialization order that we can rely on on the Ragu side. the to_wires / from_wires PR achieves that: https://github.com/zksecurity/ragu/pull/8
- however, to some degree, this is not something that can be changed, because we are extracting circuits at the level of raw input/output wires, not in a form where the high-level structure of inputs/outputs is preserved. that means there is some irreducible possibility of "semantically incorrect formalization".

for example, consider an `assertLt(x, y)` circuit that checks whether `x < y`. also, assume the circuit makes a mistake and accidentally checks `x > y`. when formalizing this, the mistake could be "shadowed" by specifying the inputs in the wrong order on the lean side, i.e. defining the input struct to be `{ y, x }` and then proving that `x < y`. due to the naming of variables, it would look like the circuit is doing exactly what it was supposed to do, even though it doesn't.  the point is that the input order/naming basically has to be considered part of the spec of a circuit, in other words, it's trusted: to see that the formalization makes sense, we need to check both that the spec makes sense and that the input order makes sense.

Side note: a "bad spec" is likely to be noticed when using a circuit as part of others, because then, the child circuit spec won't give us the statement we actually need in the parent circuit. For the same reason and in the same way, a "bad input order" would be noticed once used in a parent circuit.

TalDerei:
@ebfull see internal nested discussion threads here for supplemental context before ACK.

ebfull (approved):

PR #643, opened 2026-04-13
chore(primitives): sparser boolean gate layout

TalDerei:
https://github.com/tachyon-zcash/ragu/pull/606 modified driver-level allocs (the default `Driver::alloc` still calls `mul`, but synthesis drivers override the gate layout to (0, b, 0, d)). 

This instead modifies the gadget-level alloc for the boolean gadget to (a, b, 0, d), where in practice it's either (0, 1, 0, d) or (1, 0, 0, d) with d currently 0. Specifically, the gate gives us a * b = c with c = 0, and the `enforce_zero` gives us 1 - a - b = 0. This is a sparser gate layout, meaning sparser multiexp. 

This also frees up the d-wire for an extra allocation.

### DIFF
--- a/crates/ragu_primitives/src/boolean.rs
+++ b/crates/ragu_primitives/src/boolean.rs
@@ -38,15 +38,18 @@ pub struct Boolean<'dr, D: Driver<'dr>> {
 impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
     /// Allocates a boolean with the provided witness value.
     ///
-    /// This costs one gate and one constraint. The gate layout is
-    /// $(a, b, 0, d)$ where $a \cdot b = 0$ and the linear constraint
-    /// $1 - a - b = 0$ forces $a$ to be zero or one.
+    /// This costs one gate and two constraints. The gate layout is
+    /// $(a, b, 0, d)$ where $a \cdot b = c$, and the linear constraints
+    /// $c = 0$ and $1 - a - b = 0$ force $a$ to be zero or one.
     pub fn alloc(dr: &mut D, value: DriverValue<D, bool>) -> Result<Self> {
         let complement = value.not();
-        let (a, b, _c) =
+        let (a, b, c) =
             dr.mul(|| Ok((value.coeff().take(), complement.coeff().take(), Coeff::Zero)))?;
 
-        // Enforce a + b = 1. Together with the gate constraint a * b = 0,
+        // Enforce c = 0, so the gate constraint becomes a * b = 0.
+        dr.enforce_zero(|lc| lc.add(&c))?;
+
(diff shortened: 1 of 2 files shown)


crates/ragu_primitives/src/boolean.rs
```diff
@@ -38,22 +38,19 @@ pub struct Boolean<'dr, D: Driver<'dr>> {
 impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
     /// Allocates a boolean with the provided witness value.
     ///
-    /// This costs one gate and two constraints.
+    /// This costs one gate and one constraint. The gate layout is
```
ebfull: You have to `enforce_zero(c)` or this isn't sound, so it's not actually cheaper in terms of constraints.

ebfull (changes requested):

crates/ragu_primitives/src/boolean.rs
TalDerei: good catch, fixed!

crates/ragu_primitives/src/boolean.rs
```diff
@@ -38,22 +38,22 @@ pub struct Boolean<'dr, D: Driver<'dr>> {
 impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
     /// Allocates a boolean with the provided witness value.
     ///
-    /// This costs one gate and two constraints.
+    /// This costs one gate and two constraints. The gate layout is
+    /// $(a, b, 0, d)$ where $a \cdot b = c$, and the linear constraints
+    /// $c = 0$ and $1 - a - b = 0$ force $a$ to be zero or one.
```
ebfull: I think we should actually just omit this documentation, since the caller doesn't care.

alxiong (changes requested):
Good attempt directionally, but I don't see any benefit for this change. 
same number of gates and LC constraints, same sparsity, only alternative way to constrain it.

Frankly the old constraint is more straightforward. prefer sticking with what it was.

TalDerei:
> same number of gates and LC constraints, same sparsity, only alternative way to constrain it.

sparsity argument isn't right. this change was made because it has a sparser gate layout, meaning sparser multiexp.

alxiong (commented):
When a=0,
old: (0,0,0,d); new: (0,1,0,d)

When a=1,
old: (1,1,1,0); new: (1,0,0,d)

I guess free-d is nice and assuming 50/50 of true and false value, then the new is slightly more sparse?

Okay, I'm neutral now.

ebfull (approved):

ebfull:
@alxiong tbh the real win here is the `d`-wire, you're right that the sparsity improvements aren't significant.

PR #644, opened 2026-04-13
Minor nits and improvements

ebfull:


crates/ragu_pcd/src/lib.rs
```diff
@@ -21,7 +21,7 @@
 compile_error!("`ragu_pcd` requires the `alloc` feature to be enabled.");
 extern crate alloc;
 
-#[cfg(feature = "multicore")]
```
TalDerei, line 24: seems like a reasonable change. the original multicore gating emulated halo2's approach. the main distinctions are (1) we have cfg branching in some areas (like circuit trace eval) since we can't get around just having a single impl, and (2) iirc halo2 doesn't have a std feature flag at all, and most crates unconditionally require std.

book/src/protocol/local/overview.md
```diff
@@ -107,12 +107,12 @@ $$
 \revdot{\v{r}}{\v{r} \circ \v{z^{4n}} + \v{t} + \v{s}} = \dot{\v{k}}{\v{y^{4n}}}
 $$
 
-This combined equation is sound because all the terms of
+This combined equation is sound because the expansion of
 $\revdot{\v{r}}{\v{r} \circ \v{z^{4n}} + \v{t}}$
-are linearly independent (in $z$) from $\revdot{\v{r}}{\v{s}}$ except at the
-coefficient for $z^0$, where the combined check yields
+is polynomial in $z$, while $\revdot{\v{r}}{\v{s}}$ is $z$-independent, so
+the two interact only at $z^0$, where the combined check yields
```
TalDerei, line 113: ACK – that's clearer, reference https://github.com/tachyon-zcash/ragu/pull/635#discussion_r3063024293 if we wanted to be more explicit.

book/src/protocol/local/overview.md
```diff
@@ -83,15 +83,15 @@ suffices with overwhelming probability.
 The circuit's constraints, each requiring a fixed linear combination of wires to
 equal an instance value, can be similarly collapsed into a single revdot check.
 Given a (sparse) instance vector $\v{k} \in \F^{4n}$, the $j$-th constraint can
-be written $\revdot{\v{r}}{\v{s_j}} = \v{k}_j$ where the structured vector
-$\v{s_j}$ encodes the constraint's wire weights at the revdot complements of the
+be written $\revdot{\v{r}}{\v{u_j}} = \v{k}_j$ where the structured vector
+$\v{u_j}$ encodes the constraint's wire weights at the revdot complements of the
```
TalDerei, line 87: hmm, is `u_j` for basis vectors clearer notation than `s_j`?  `s = \sum y^j s_j` makes sense.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -274,6 +268,16 @@ struct Counter<F> {
...
 
+    /// Base for the remap geometric sequence used by [`WireMap::convert_wire`].
+    /// Independent from `x0..x3` so remap-minted wire values cannot collide
+    /// with allocated wire values.
+    x_remap: F,
+
+    /// Running value for [`WireMap::convert_wire`]: each call returns the
+    /// current value and multiplies by `x_remap`. Persists across routines and
+    /// across input/output remaps so every minted value is distinct.
+    remap_current: F,
```
ebfull: This shouldn't live in the `Counter`, since we need to reset it in every routine.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -256,12 +256,6 @@ struct Counter<F> {
     num_gates: usize,
     segments: Vec<SegmentRecord>,
 
-    /// When false, `gate` advances geometric sequences but does not increment
-    /// constraint counts.  Used during input and output wire remapping in
-    /// [`routine`](Driver::routine), where only `alloc` (and transitively
-    /// `mul`, which calls `gate`) is reachable via [`WireMap::convert_wire`].
-    counting: bool,
```
TalDerei, line 263: yeah this flag always seemed hacky to me for distinguishing between remap and real allocation within the same routine call.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -316,6 +320,7 @@ impl<F: FromUniformBytes<64>> Counter<F> {
         let y = point(4);
         let h = point(5);
         let one = point(6);
+        let x_remap = point(7);
```
TalDerei, line 334: ACK – smart, removes assumption on alloc. In this approach, the wires derived from this separate geometric sequence are unique.

TalDerei (approved):
LGTM

crates/ragu_pcd/src/lib.rs
ebfull: multicore implies std, so this feature change is algebraically consistent

book/src/protocol/local/overview.md
ebfull: s_j ends up being overloaded notationally later in several spots, and the only purpose of this s_j thing is to implicitly write the weighted sum in this description. It isn't used elsewhere, so I picked a fresh variable.

crates/ragu_pcd/src/lib.rs
TalDerei: sorry should've been more explicit in diff linkage. the change to gating on std here seems like a reasonable change!

crates/ragu_circuits/src/metrics.rs
TalDerei, line 1: these kinds of refactors often carry stale doc references, i'd grep for anything that references the counting bool / input map.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -611,7 +569,9 @@ pub(crate) mod tests {
         type Dst = Counter<F>;
 
         fn convert_wire(&mut self, _: &Src::ImplWire) -> Result<WireEval<F>> {
-            self.counter.alloc(|| unreachable!())
+            let v = self.counter.scope.remap_current;
+            self.counter.scope.remap_current *= self.counter.x_remap;
+            Ok(WireEval::Value(v))
```
TalDerei: nit; optionally, a helper on `Counter` would deduplicate these `convert_wire` impls.

PR #646, opened 2026-04-13
doc: add simulator page and fix routing description in concrete drivers

LucidSamuel:
adds `book/src/guide/drivers/simulator.md` covering the Simulator driver's API, constraint checking, metric collection, and rank budgeting workflow. 

fixes concrete.md to match the actual implementation (prediction is used only for aux data; no consistency check occurs).


ebfull (changes requested):
Similar to the allocator book PR you opened, this should exist in a Primitives subsection rather than being shoehorned into the existing driver section, since it's provided in that crate.

PR #647, opened 2026-04-14
Second class allocation

### LINKED ISSUE #645: Move wire allocation patterns to `ragu_primitives`
Drivers currently provide their own `alloc` method (for standalone wire allocation) due to a confluence of design choices:

1. All wires are created in groups of four (a `gate` call.)
2. All wires are assigned when they are created.
3. Different abstractions in the same circuit/routine will need to perform standalone allocations.
4. Multiple standalone allocations optimally reuse the same gate. (The `b` and `d` wires.)
5. We need to know the wire locations in many situations when we `enforce_zero`, for performance. We cannot just queue up future `gate` calls; we need the `Wire` type.

Design choice (2) is deliberate. There is currently no way to reserve a `d` wire for future assignment at a (undetermined) later part of synthesis, so a paired-allocator pattern has to be baked into the drivers themselves. This pattern is actually implemented multiple times in `ragu_circuits` for each compatible driver.

However, the `d` wire is actually a good exception to (2). It's already a special-case; it can only have a nonzero assignment in certain situations, and it is generally not invoked by circuit developers. As long as we're still forced to provide a `d` assignment before we involve it in any constraints, it seems fine for us to have a dangling token representing the `d` wire in every gate that can be assigned at some point afterwards.

My desired approach is to change the `gate` method: don't ask for an assignment for `d` immediately, instead returning an `Extra` token (associated type on `DriverTypes`) that can be dropped and ignored later, with the assumed assignment of zero. Users can stash this `Extra` somewhere and ask the driver to assign it a value later (via another method, `assign_extra` on `DriverTypes`) returning the actual `D` wire. (The `Extra` type can be used to store metadata needed to construct the wire when the assignment decision is made by the circuit.)

This allows `Driver::alloc` to be deleted and for an allocation pattern to be created which exists entirely within the `ragu_primitives` crate. The standard allocator (default impl of `Driver::alloc`) which wastes a full `gate` just to create a single free wire can be implemented for the stateless allocator `()`. The paired allocator abstraction can be implemented, stashing an `Extra` internally. Both can implement a shared `Allocator` trait.

* As a result, more complicated allocators can be created. For example, if the `Boolean` gadget is [refactored](https://github.com/tachyon-zcash/ragu/pull/643) to produce a spare `d` wire it can be handed to the generic `Allocator`, which can then store many `Extra` for future use by another part of the circuit/routine.
* The drivers no longer do allocation bookkeeping, which is something I've wanted for a while.

ebfull:
Closes #645.

### DIFF
--- a/book/src/guide/drivers/index.md
+++ b/book/src/guide/drivers/index.md
@@ -45,18 +45,6 @@ available everywhere. The [`constant()`] method is a simple helper that returns
 a virtual wire assigned to a constant, which is free because it is a virtual
 wire that scales [`ONE`].
 
-### Allocation
-
-Sometimes only a single wire is needed, but [`mul()`] always allocates three. To
-support single-wire allocation, drivers provide an additional [`alloc()`] method
-that allocates and assigns one wire.
-
-By default, `alloc` calls `mul`, returns the $a$ wire, and sets the
-corresponding $b$ and $c$ wires to zero to satisfy the gate—wasting $b$ and $c$. Drivers may override `alloc` to avoid this
-overhead. For example, synthesis drivers return the $a$ wire from a `mul`
-operation, stash the associated $b$ wire for the next `alloc` call, and fill in
-$c$ later to satisfy the gate.
-
 ### The `'dr` Lifetime {#the-dr-lifetime}
 
 Drivers are parameterized by a lifetime `'dr` that ties [routines] to the
@@ -89,7 +77,7 @@ should always bound on `Driver<'dr>`, not `DriverTypes`.
 
 ### Purity {#purity}
 
-All four closure-accepting `Driver` methods—[`mul()`], [`alloc()`], [`add()`],
+All three closure-accepting `Driver` methods—[`mul()`], [`add()`],
 and [`enforce_zero()`]—require their closures to be [`Fn`], not `FnOnce` or
 `FnMut`. This is a deliberate signal that closures should be side-effect-free:
 synthesis must produce identical constraints regardless of whether a given
@@ -97,10 +85,10 @@ driver invokes the closure. `Fn` prevents accidental `&mut` captures, although
 it does not prevent interior mutability.
 
 The two closure families differ in whether they have additional protection
-beyond `Fn`. For the witness-providing closures on [`mul()`] and [`alloc()`],
+beyond `Fn`. For the witness-providing closure on [`mul()`],
 the [`Maybe`]/[`DriverValue`] system provides a harder compile-time guarantee:
-drivers with `MaybeKind = Empty` never call those closures at all, and the
-closure bodies are dead-code-eliminated. The expression-building closures on
+drivers with `MaybeKind = Empty` never call the closure at all, and the
+closure body is dead-code-eliminated. The expression-building closures on
 [`add()`] and [`enforce_zero()`] have no such backstop; drivers with `MaybeKind
 = Empty` still call these closures when building constraint structure.
 
@@ -115,7 +103,6 @@ to have the same value by calling [`enforce_zero()`] on their difference.
 [`add()`]: ragu_core::drivers::Driver::add
 [`ONE`]: ragu_core::drivers::Driver::ONE
 [`constant()`]: ragu_core::drivers::Driver::constant
-[`alloc()`]: ragu_core::drivers::Driver::alloc
 [`Driver`]: ragu_core::drivers::Driver
 [`Routine`]: ragu_core::routines::Routine
 [`routine()`]: ragu_core::drivers::Driver::routine
--- a/crates/ragu_circuits/benches/criterion/trace.rs
+++ b/crates/ragu_circuits/benches/criterion/trace.rs
@@ -9,7 +9,7 @@ use ragu_core::{
     routines::{Prediction, Routine},
 };
 use ragu_pasta::Fp;
-use ragu_primitives::Element;
+use ragu_primitives::{Element, allocator::StubAllocator};
 use rand::{SeedableRng, rngs::StdRng};
 
 /// A synthetic routine that does `depth` squarings in `execute()` but
@@ -45,6 +45,7 @@ impl Routine<Fp> for HeavyKnownRoutine {
         // Predict the output by computing the repeated squaring on the value.
         let output = Element::alloc(
             dr,
+            &mut (),
             D::try_just(|| {
                 let mut v = *input.value().take();
                 for _ in 0..self.depth {
@@ -74,7 +75,7 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         instance: DriverValue<D, Self::Instance<'instance>>,
     ) -> Result<Bound<'dr, D, Self::Output>> {
-        Element::alloc(dr, instance)
+        Element::alloc(dr, &mut StubAllocator, instance)
     }
 
     fn witness<'dr, 'witness: 'dr, D: Driver<'dr, F = Fp>>(
@@ -82,7 +83,7 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         witness: DriverValue<D, Self::Witness<'witness>>,
     ) -> Result<WithAux<Bound<'dr, D, Self::Output>, DriverValue<D, Self::Aux<'witness>>>> {
-        let input = Element::alloc(dr, witness)?;
+        let input = Element::alloc(dr, &mut StubAllocator, witness)?;
         let routine = HeavyKnownRoutine { depth: self.depth };
 
         let mut result = dr.routine(routine.clone(), input.clone())?;
--- a/crates/ragu_circuits/benches/criterion/trace.rs
+++ b/crates/ragu_circuits/benches/criterion/trace.rs
@@ -9,7 +9,7 @@ use ragu_core::{
     routines::{Prediction, Routine},
 };
 use ragu_pasta::Fp;
-use ragu_primitives::{Element, allocator::StubAllocator};
+use ragu_primitives::{Element, allocator::SimpleAllocator};
 use rand::{SeedableRng, rngs::StdRng};
 
 /// A synthetic routine that does `depth` squarings in `execute()` but
@@ -75,7 +75,8 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         instance: DriverValue<D, Self::Instance<'instance>>,
     ) -> Result<Bound<'dr, D, Self::Output>> {
-        Element::alloc(dr, &mut StubAllocator, instance)
+        let mut allocator = SimpleAllocator::new();
+        Element::alloc(dr, &mut allocator, instance)
     }
 
     fn witness<'dr, 'witness: 'dr, D: Driver<'dr, F = Fp>>(
@@ -83,7 +84,8 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         witness: DriverValue<D, Self::Witness<'witness>>,
     ) -> Result<WithAux<Bound<'dr, D, Self::Output>, DriverValue<D, Self::Aux<'witness>>>> {
-        let input = Element::alloc(dr, &mut StubAllocator, witness)?;
+        let mut allocator = SimpleAllocator::new();
+        let input = Element::alloc(dr, &mut allocator, witness)?;
         let routine = HeavyKnownRoutine { depth: self.depth };
 
         let mut result = dr.routine(routine.clone(), input.clone())?;
--- a/crates/ragu_circuits/benches/criterion/trace.rs
+++ b/crates/ragu_circuits/benches/criterion/trace.rs
@@ -75,8 +75,8 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         instance: DriverValue<D, Self::Instance<'instance>>,
     ) -> Result<Bound<'dr, D, Self::Output>> {
-        let mut allocator = SimpleAllocator::new();
-        Element::alloc(dr, &mut allocator, instance)
+        let allocator = &mut SimpleAllocator::new();
+        Element::alloc(dr, allocator, instance)
     }
 
     fn witness<'dr, 'witness: 'dr, D: Driver<'dr, F = Fp>>(
@@ -84,8 +84,8 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         witness: DriverValue<D, Self::Witness<'witness>>,
     ) -> Result<WithAux<Bound<'dr, D, Self::Output>, DriverValue<D, Self::Aux<'witness>>>> {
-        let mut allocator = SimpleAllocator::new();
-        let input = Element::alloc(dr, &mut allocator, witness)?;
+        let allocator = &mut SimpleAllocator::new();
+        let input = Element::alloc(dr, allocator, witness)?;
         let routine = HeavyKnownRoutine { depth: self.depth };
 
         let mut result = dr.routine(routine.clone(), input.clone())?;
--- a/crates/ragu_circuits/src/polynomials/mod.rs
+++ b/crates/ragu_circuits/src/polynomials/mod.rs
@@ -90,8 +90,9 @@ pub trait Rank:
 
         *Emulator::emulate_wireless((x, z), |dr, xz| {
             let (x, z) = xz.cast();
-            let x = Element::alloc(dr, x)?;
-            let z = Element::alloc(dr, z)?;
+            let allocator = &mut ();
+            let x = Element::alloc(dr, allocator, x)?;
+            let z = Element::alloc(dr, allocator, z)?;
 
             dr.routine(txz::Evaluate::<Self>::new(), (x, z))
         })
--- a/crates/ragu_circuits/src/polynomials/txz.rs
+++ b/crates/ragu_circuits/src/polynomials/txz.rs
@@ -115,6 +115,7 @@ impl<F: Field, R: Rank> Routine<F> for Evaluate<R> {
         // Compute output using the precomputed inversions
         let output = Element::alloc(
             dr,
+            &mut (),
             D::try_just(|| {
                 let x = *input.0.value().take();
                 let z = *input.1.value().take();
@@ -144,7 +145,7 @@ impl<F: Field, R: Rank> Routine<F> for Evaluate<R> {
 #[cfg(test)]
 mod tests {
     use ragu_pasta::Fp;
-    use ragu_primitives::Simulator;
+    use ragu_primitives::{Simulator, allocator::StubAllocator};
 
     use super::*;
     use crate::polynomials::ProductionRank;
@@ -160,8 +161,8 @@ mod tests {
 
         Simulator::simulate((x, z), |dr, witness| {
             let (x, z) = witness.cast();
-            let x = Element::alloc(dr, x)?;
-            let z = Element::alloc(dr, z)?;
+            let x = Element::alloc(dr, &mut StubAllocator, x)?;
+            let z = Element::alloc(dr, &mut StubAllocator, z)?;
 
             dr.reset();
             dr.routine(evaluator.clone(), (x, z))?;
--- a/crates/ragu_circuits/src/polynomials/txz.rs
+++ b/crates/ragu_circuits/src/polynomials/txz.rs
@@ -145,7 +145,7 @@ impl<F: Field, R: Rank> Routine<F> for Evaluate<R> {
 #[cfg(test)]
 mod tests {
     use ragu_pasta::Fp;
-    use ragu_primitives::{Simulator, allocator::StubAllocator};
+    use ragu_primitives::{Simulator, allocator::SimpleAllocator};
 
     use super::*;
     use crate::polynomials::ProductionRank;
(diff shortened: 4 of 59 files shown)


crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// Allocator that forwards to [`Driver::alloc`] unchanged, preserving any
+/// driver-specific override.
+///
+/// Use this at call sites that need to respect the running driver's own
+/// allocation behavior — for instance, the allocation counter tracked by
+/// [`Simulator`](crate::Simulator), or the paired-allocation optimization
+/// used by the production prover drivers.
+pub struct StubAllocator;
```
TalDerei: ACK – this is a temp allocator. assuming once driver allocs are removed, it'll be replaced by what you're referring to as a "batching implementation" that internally calls `dr.gate` and manages d-wire (via that `extra` token). so wire pairing allocations move to gadget level via allocator from globally at the driver level.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// Allocates wires on behalf of a gadget.
+///
+/// Implementations decide how to turn a witness-producing closure into a
+/// driver wire. The simplest implementations forward to [`Driver::alloc`]
+/// (see [`StubAllocator`]) or imitate its default body (see the impl for
+/// `()`). A batching implementation can pair two consecutive allocations
+/// into a single $a \cdot b = c$ gate, reusing the $a$ and $c$ wires that
+/// [`Driver::alloc`]'s default leaves zeroed.
+pub trait Allocator<'dr, D: Driver<'dr>> {
```
TalDerei, line 35: ACK – gadget-level (local) alloc strategy.

TalDerei:
meta observation: isn't `Driver::routine` a similar anti-pattern?  ie. we're baking a policy decision (how to dispatch routines) into a driver trait method? was thinking about a quasi-split where routine is extracted to the gadget level, and the save/restore scope and execution stuff remains inside the driver. this is a vague intuition.

crates/ragu_circuits/benches/criterion/trace.rs
```diff
@@ -74,15 +75,15 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         instance: DriverValue<D, Self::Instance<'instance>>,
     ) -> Result<Bound<'dr, D, Self::Output>> {
-        Element::alloc(dr, instance)
+        Element::alloc(dr, &mut StubAllocator, instance)
```
TalDerei: alternative construction i've been entertaining; wondering if instead of threading allocators everywhere, we have some kind of newtype wrapper around the driver

```ts
  struct WithAllocator<'a, D, A> {                                                                                                                                                                                                                   
      driver: &'a mut D,                                                                                                                                                                                                                             
      allocator: A,                                                                                                                                                                                                                                  
  }                                                                                                              
                                                                                                                                                                                                                                                     
  impl<D: Driver, A: Allocator> Driver for WithAllocator<D, A> {                                                                                                                                                                                     
      // delegate gate, mul, enforce_zero, etc. to self.driver + override alloc to use self.allocator                                                                                                                                                                                                        
  }   
```

and gadget callsite creates it, and scopes it to the method body                                                                                                                                                                                           
                      
```ts
  let mut dr = WithAllocator::new(dr, PairedAllocator::new());                                                                                                                                                                                       
  let x = Element::alloc(&mut dr, value1)?;                                                                                                                                                                                                          
  let y = Element::alloc(&mut dr, value2)?;                                                                                                                                                                                                         
```

so it's like structurally similar to driver-level overrides, but the wrapper is created and scoped at the gadget level, and we don't need these API changes. guess the downside here is delegating the entire driver trait through the wrapper is probably a lot of boilerplate.

ebfull: Yeah! I was hoping we could explore that afterward, as an ergonomic improvement. I wasn't sure what it would look like, but figured it could take better shape when the other moving parts have finished.

ebfull:
Only the specific driver knows how to handle, schedule and process routines, and must be able to do so flexibly (with full knowledge of the wire values, for instance). Thus, routines not only cannot be scheduled by applications or gadgets but the driver must take control flow away from the circuits when they are executed.

crates/ragu_circuits/benches/criterion/trace.rs
alxiong: my first reaction was default generic param `Vec<T, A = Global>` analogous design where `sx::Collector<D, A = PairedAllocator>` (instead of wrapper type). But haven't thought through carefully yet.

book/src/guide/drivers/index.md
```diff
@@ -45,18 +45,6 @@ available everywhere. The [`constant()`] method is a simple helper that returns
 a virtual wire assigned to a constant, which is free because it is a virtual
 wire that scales [`ONE`].
 
-### Allocation
```
ebfull, line 48: We should have a section inside of the primitives part of the guide, which @LucidSamuel has volunteered to write in a separate PR, describing allocation. It is no longer a concept in the core traits.

crates/ragu_circuits/benches/criterion/trace.rs
```diff
@@ -74,15 +75,17 @@ impl Circuit<Fp> for HeavyRoutineCircuit {
         dr: &mut D,
         instance: DriverValue<D, Self::Instance<'instance>>,
     ) -> Result<Bound<'dr, D, Self::Output>> {
-        Element::alloc(dr, instance)
+        let mut allocator = SimpleAllocator::new();
+        Element::alloc(dr, &mut allocator, instance)
```
ebfull: I think all `instance` methods are ran on emulators, and so we should just pass in `&mut ()` here and in the other places.

crates/ragu_circuits/src/polynomials/txz.rs
```diff
@@ -160,13 +161,12 @@ mod tests {
 
         Simulator::simulate((x, z), |dr, witness| {
             let (x, z) = witness.cast();
-            let x = Element::alloc(dr, x)?;
-            let z = Element::alloc(dr, z)?;
+            let mut allocator = SimpleAllocator::new();
+            let x = Element::alloc(dr, &mut allocator, x)?;
+            let z = Element::alloc(dr, &mut allocator, z)?;
```
ebfull, line 167: There's a mixture of conventions in the diff: some do `let mut allocator` and others do `let allocator = &mut`

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -181,10 +183,6 @@ impl<F: Field, R: Rank> DriverScope<SxScope<F>> for Evaluator<'_, F, R> {
 impl<F: Field, R: Rank> Evaluator<'_, F, R> {
     /// Advances the gate counter and running monomials, returning the raw
     /// $(a, b, c)$ monomial evaluations before advancement.
-    ///
-    /// This is the shared core of [`gate`](DriverTypes::gate) and
-    /// [`mul`](Driver::mul). The $d$-wire monomial ($b \cdot \text{b\_to\_d}$)
-    /// is only computed by `gate`, saving one field multiplication per `mul` call.
```
ebfull, line 187: Good, `mul` is a convenience method now; this "shared core" is implicit. Consequence: we should not be overloading `mul` anywhere in the codebase.

crates/ragu_circuits/src/s/sx.rs
```diff
@@ -181,10 +183,6 @@ impl<F: Field, R: Rank> DriverScope<SxScope<F>> for Evaluator<'_, F, R> {
...
     fn advance_gate(&mut self) -> Result<(F, F, F)> {
```
ebfull: This is just used in `gate` now, which means it should be inlined?

crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -44,27 +45,30 @@ where
 {
     /// Builds a [`BondingObject`] from this [`MultiStage`] circuit.
     ///
-    /// The witness must use no gates: [`Driver::alloc`],
-    /// [`Driver::add`], and [`Driver::enforce_zero`] are permitted (without
-    /// referencing the [`Driver::ONE`] wire), but [`Driver::mul`] and
-    /// [`Driver::constant`] are rejected.
+    /// After the [`StageBuilder`] is finalized, only [`Driver::add`] and
+    /// [`Driver::enforce_zero`] are permitted (without referencing the
+    /// [`Driver::ONE`] wire). [`Driver::mul`] and [`Driver::constant`]
```
ebfull: `Driver::gate` is rejected which is why `Driver::mul` is.

crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -44,27 +45,30 @@ where
...
     {
-        // Validate: run synthesis with a driver that rejects mul/gate and ONE usage.
+        // Validate: run synthesis with a driver that rejects ONE usage
+        // and — after the stage builder finalizes — mul/gate.
         let mut validator = BondingValidator::<F>::new();
-        self.witness(&mut validator, Empty)?;
+        self.circuit.witness(
+            StageBuilder::new_with_on_finish(&mut validator, BondingValidator::freeze),
+            Empty,
+        )?;
```
ebfull, line 73: There are so few places where the stage builder is actually created, and it is not a public interface, that forcing there to be a single constructor with a succinct name is preferable.

crates/ragu_circuits/src/raw.rs
```diff
@@ -46,10 +46,11 @@ use crate::WithAux;
...
 ///
-/// Wraps the `(a, b, c, d)` tuple returned by
-/// [`DriverTypes::gate`](ragu_core::drivers::DriverTypes::gate) with named
-/// fields for readability.
-#[allow(dead_code)] // TODO: this is only used by tests...
+/// Wraps the $(A, B, C)$ wires returned by
+/// [`DriverTypes::gate`](ragu_core::drivers::DriverTypes::gate) together with
+/// the $D$ wire obtained via
+/// [`DriverTypes::assign_extra`](ragu_core::drivers::DriverTypes::assign_extra).
+#[allow(dead_code)]
```
ebfull: Explanation of this was removed.

crates/ragu_core/src/drivers.rs
```diff
@@ -116,14 +116,27 @@ pub trait DriverTypes {
...
+
+    /// Allocates the wires $(A, B, C)$ with the constraint $A \cdot B = C$
+    /// and returns an [`Extra`](Self::Extra) token for the auxiliary $D$ wire.
+    /// The $D$ wire is assigned to zero by default. The gate also imposes the
+    /// constraint $C \cdot D = 0$, which makes $D$ useless in the typical
+    /// case: whenever $C$ is nonzero, $D$ is forced to zero. But when $C$ is
+    /// guaranteed to be zero, $D$ becomes unconstrained and available for use.
+    ///
+    /// To override the default $D = 0$ assignment, pass the returned `Extra`
+    /// to [`assign_extra`](Self::assign_extra).
```
ebfull: I think this could be better phrased to reassure the reader that the default $d = 0$ trivially satisfies the c*d=0 constraint.

crates/ragu_pcd/src/header.rs
```diff
@@ -93,8 +93,13 @@ pub trait Header<F: Field>: Send + Sync + Any {
     type Output: Write<F>;
 
     /// Encode some data into a gadget representing this header.
-    fn encode<'dr, D: Driver<'dr, F = F>>(
+    ///
+    /// Implementations should pass `allocator` to
+    /// [`Element::alloc`](ragu_primitives::Element::alloc) calls rather than
```
ebfull: This is too specific to `Element`, the allocation pattern is intended to be more general than that.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,94 @@
...
+//! This module ships two implementations:
+//!
+//! - [`Allocator`] for `()` — a stateless allocator that allocates a
+//!   multiplication gate with zeroed $a$ and $c$ wires via
+//!   [`Driver::mul`], returning the $b$ wire. Simple but wasteful: each
+//!   allocation consumes a full gate.
+//!
+//! - [`SimpleAllocator`] — pairs consecutive allocations into one gate,
+//!   stashing the `Extra` token from the first allocation for use by the
+//!   second. This halves the number of gates consumed by pure allocations.
```
ebfull: Allocation should be _introduced_ here; it doesn't exist in the core traits. The driver isn't responsible or aware of allocation anymore.

The purpose is to resolve a separate concern: wires can currently only be allocated in groups of $4$ by the driver, but we often need standalone wires. It's generally possible to assign two arbitrary values to two of the four wires (leaving the others zero), which means we need an allocator that can repurpose wires from previous gates.

We also need to be generic because allocators have state, and there are different ones that might need to work for different purposes.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,94 @@
...
+/// [`Extra`](ragu_core::drivers::DriverTypes::Extra) token for the $D$ wire.
+/// The second call redeems that token via
+/// [`assign_extra`](ragu_core::drivers::DriverTypes::assign_extra),
+/// returning $D$ without allocating a new gate.
+///
+/// This is the standard paired-allocation strategy. Gadgets that perform
+/// many allocations should use `SimpleAllocator` to halve gate cost.
+///
+/// Dropping a `SimpleAllocator` that still holds a stashed token is safe:
+/// the driver already assigned $D = 0$ for that gate.
```
ebfull, line 71: Kind of an implementation detail but we can leave it.

crates/ragu_circuits/benches/criterion/trace.rs
ebfull: Ah, the `witness` could theoretically call the `instance` method, and I've done this before. I'll leave this, but also the `instance` method should be removed someday.

crates/ragu_circuits/src/s/sxy.rs
```diff
@@ -232,25 +238,15 @@ impl<'dr, F: Field, R: Rank> Driver<'dr> for Evaluator<'_, F, R> {
...
 
-    /// Advances the gate counter and returns $(a, b, c)$ without computing the
-    /// $d$-wire monomial.
-    fn mul(
-        &mut self,
-        _: impl Fn() -> Result<(Coeff<Self::F>, Coeff<Self::F>, Coeff<Self::F>)>,
-    ) -> Result<(Self::Wire, Self::Wire, Self::Wire)> {
-        let (a, b, c) = self.advance_gate()?;
-        Ok((WireEval::Value(a), WireEval::Value(b), WireEval::Value(c)))
-    }
```
TalDerei, line 252: ACK – purged overrides (iirc motivation was to skip the extra multiplication `gate` was doing to compute d).

crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn alloc(&mut self, dr: &mut D, value: impl Fn() -> Result<Coeff<D::F>>) -> Result<D::Wire> {
+        if let Some(extra) = self.stash.take() {
+            dr.assign_extra(extra, value)
+        } else {
+            let (_, b, _, extra) = dr.gate(|| Ok((Coeff::Zero, value()?, Coeff::Zero)))?;
+            self.stash = Some(extra);
+            Ok(b)
+        }
+    }
+}
```
TalDerei, line 99: ACK stashing logic.

book/src/guide/drivers/index.md
TalDerei, line 48: found some other stale book references which claims gate returns four wires, which is now stale.

crates/ragu_circuits/src/staging/builder.rs
```diff
@@ -252,7 +259,11 @@ impl<'a, 'dr, D: Driver<'dr>, R: Rank, Finished: Stage<D::F, R>>
     StageBuilder<'a, 'dr, D, R, Finished, Finished>
 {
     /// Obtains the underlying driver after finishing the last stage.
+    ///
+    /// If the builder was constructed with an `on_finish` hook, the hook
```
TalDerei, line 263: nit; 'if' implies this is optional.

crates/ragu_primitives/src/element.rs
```diff
@@ -60,11 +61,16 @@ pub struct Element<'dr, D: Driver<'dr>> {
 }
 
 impl<'dr, D: Driver<'dr>> Element<'dr, D> {
-    /// Allocates an element with the provided witness assignment.
+    /// Allocates an element with the provided witness assignment, using the
+    /// supplied [`Allocator`] to create the underlying wire.
     ///
     /// This costs one allocation.
```
TalDerei, line 67: "This costs one allocation" is now stale, cost technically depends on allocator passed in.

TalDerei (approved):
this looks great! filed https://github.com/tachyon-zcash/ragu/issues/650 as follow-up work.

book/src/guide/drivers/index.md
ebfull: There may be some high level sense in which it's appropriate to say it returns four wires, since it's not used much and the API clears things up by literally giving `d` a special type and documenting how to use it. But I'm open to clarifications.

crates/ragu_primitives/src/element.rs
ebfull: "One allocation" is accurate; we call the provided allocator once. The caller, if they care at all about this, would only find that information useful.

crates/ragu_circuits/src/raw.rs
```diff
@@ -192,8 +187,9 @@ where
     RC: RawCircuit<F> + 'dr,
 {
     // 1. Allocate the SYSTEM gate (gate 0).
-    let system_gate =
-        GateWires::from(dr.gate(|| Ok((Coeff::Zero, Coeff::Zero, Coeff::Zero, Coeff::Zero)))?);
+    let (a, b, c, extra) = dr.gate(|| Ok((Coeff::Zero, Coeff::Zero, Coeff::Zero)))?;
+    let d = dr.assign_extra(extra, || Ok(Coeff::Zero))?;
```
alxiong:
```suggestion
    let (a, b, c, d) = dr.gate(|| Ok((Coeff::Zero, Coeff::Zero, Coeff::Zero)))?;
```

crates/ragu_primitives/src/allocator.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// it may be called zero or more times, it must be side-effect-free,
+    /// and errors returned from it propagate to the caller.
+    fn alloc(&mut self, dr: &mut D, value: impl Fn() -> Result<Coeff<D::F>>) -> Result<D::Wire>;
+}
+
+/// Stateless allocator that uses one gate per allocation.
+///
+/// Each call produces a multiplication gate $0 \cdot b = 0$ and returns
+/// the $b$ wire, wasting the $a$ and $c$ wires.
+impl<'dr, D: Driver<'dr>> Allocator<'dr, D> for () {
```
alxiong, line 48: I'd wish to explicitly `struct StatelessAllocator` (or `DefaultAllocator` or something), we abuse `()` too much in our code base, `Element::alloc(&mut dr, &mut (), v)` reads slightly confusing.

crates/ragu_circuits/src/polynomials/mod.rs
```diff
@@ -90,8 +90,9 @@ pub trait Rank:
 
         *Emulator::emulate_wireless((x, z), |dr, xz| {
             let (x, z) = xz.cast();
-            let x = Element::alloc(dr, x)?;
-            let z = Element::alloc(dr, z)?;
+            let allocator = &mut ();
+            let x = Element::alloc(dr, allocator, x)?;
+            let z = Element::alloc(dr, allocator, z)?;
```
alxiong: what's wrong with using `SimpleAllocator` here?

crates/ragu_pcd/src/fuse/_05_outer_error.rs
```diff
@@ -74,11 +74,12 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             ),
             |dr, witness| {
                 let (preamble_witness, inner_error_terms, y, mu, nu) = witness.cast();
+                let allocator = &mut ();
```
alxiong: again, given that we know subsequent `y, mu, nu` are unrelated, why can't we use `SimpleAllocator`?

crates/ragu_pcd/src/internal/native/stages/inner_error.rs
```diff
@@ -69,10 +69,10 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize, FP: fold_revdot::Parameters>
     where
         Self: 'dr,
     {
-        // Allocate nested error terms
+        let allocator = &mut ();
```
alxiong, line 72: Why not `SimpleAllocator`

alxiong: when trying to reason about this, i find it tricky because it's a non-local argument.

we use stateless allocator here because the caller, `StageBuilder::configure_stage()`, only use wireless driver to get wire count `stage.witness(Emulator::counter())`, and the stage builder actually use a `SimpleAllocator` for actual allocation:
https://github.com/tachyon-zcash/ragu/blob/f94ffb1551ece95ef9dd1ff8f4f8cb3326eddd4c/crates/ragu_circuits/src/staging/builder.rs#L205-L208

So it doesn't matter what allocator to use here, because they will render exactly the same `num_wires()`. 
This realization is no longer local, but determined in accordance with the caller/usage context.

PR #651, opened 2026-04-14
chore(primitives): pool allocator

### LINKED ISSUE #650: allocator for reclaiming spare wires
The only thing in https://github.com/tachyon-zcash/ragu/issues/645 (first bullet point at the bottom) that https://github.com/tachyon-zcash/ragu/pull/647 doesn't close since it's out of scope.

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/650.

### DIFF
--- a/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
+++ b/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
@@ -60,7 +60,7 @@ use ragu_core::{
     gadgets::Bound,
     maybe::Maybe,
 };
-use ragu_primitives::{Element, Endoscalar, GadgetExt, allocator::SimpleAllocator};
+use ragu_primitives::{Element, Endoscalar, GadgetExt, allocator::PoolAllocator};
 
 use super::super::{
     InternalCircuitIndex, InternalCircuitValues, RxComponent, RxIndex,
@@ -159,7 +159,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
         let query = query.unenforced(dr, witness.as_ref().map(|w| w.query_witness))?;
         let eval = eval.unenforced(dr, witness.as_ref().map(|w| w.eval_witness))?;
 
-        let allocator = &mut SimpleAllocator::new();
+        let allocator = &mut PoolAllocator::new();
         let mut unified_output = OutputBuilder::new(witness.map(|w| w.unified));
 
         // Retrieve Fiat-Shamir challenges from the unified instance.
@@ -225,7 +225,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
             // First extract endoscalar from pre_beta and compute effective beta.
             let computed_v = {
                 let pre_beta = unified_output.pre_beta.read(dr, allocator)?;
-                let beta_endo = Endoscalar::extract(dr, pre_beta)?;
+                let beta_endo = Endoscalar::extract(dr, allocator, pre_beta)?;
                 let effective_beta = beta_endo.lift(dr)?;
                 let mut horner = Horner::new(&effective_beta);
                 fu.write(dr, &mut horner)?;
--- a/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
+++ b/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
@@ -162,7 +162,14 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
         let allocator = &mut PoolAllocator::new();
         let mut unified_output = OutputBuilder::new(witness.map(|w| w.unified));
 
+        // Extract endoscalar early: each of the 128 Boolean gates has a
+        // spare D wire. Donating them to the pool lets subsequent reads
+        // reuse those wires instead of allocating fresh gates.
+        let pre_beta = unified_output.pre_beta.read(dr, allocator)?;
+        let beta_endo = Endoscalar::extract(dr, allocator, pre_beta)?;
+
         // Retrieve Fiat-Shamir challenges from the unified instance.
+        // These reads draw from the pool donated above.
         let w = unified_output.w.read(dr, allocator)?;
         let y = unified_output.y.read(dr, allocator)?;
         let z = unified_output.z.read(dr, allocator)?;
@@ -222,10 +229,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
 
             // Step 3: Compute v = f(u) + beta * eval via Horner accumulation.
             // This combines f(u) with the evaluation component polynomials.
-            // First extract endoscalar from pre_beta and compute effective beta.
             let computed_v = {
-                let pre_beta = unified_output.pre_beta.read(dr, allocator)?;
-                let beta_endo = Endoscalar::extract(dr, allocator, pre_beta)?;
                 let effective_beta = beta_endo.lift(dr)?;
                 let mut horner = Horner::new(&effective_beta);
                 fu.write(dr, &mut horner)?;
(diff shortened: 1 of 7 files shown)


crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
```diff
@@ -159,10 +159,17 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
...
         let mut unified_output = OutputBuilder::new(witness.map(|w| w.unified));
 
+        // Extract endoscalar early: each of the 128 Boolean gates has a
+        // spare D wire. Donating them to the pool lets subsequent reads
+        // reuse those wires instead of allocating fresh gates.
+        let pre_beta = unified_output.pre_beta.read(dr, allocator)?;
+        let beta_endo = Endoscalar::extract(dr, allocator, pre_beta)?;
+
         // Retrieve Fiat-Shamir challenges from the unified instance.
+        // These reads draw from the pool donated above.
```
TalDerei, line 172: `compute_v` change to demonstrate gate savings by sourcing from the pool allocator. in fact, there seems to be 128 free d (per https://github.com/tachyon-zcash/ragu/pull/651#discussion_r3083329281) wires of capacity to exploit, so we can rearrange some internal circuit stuff.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -53,6 +57,9 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
         // this gives a(1 - a) = 0, so a is zero or one.
         dr.enforce_zero(|lc| lc.add(&D::ONE).sub(&a).sub(&b))?;
 
+        // The gate's spare $D$ wire is donated to `allocator`.
+        allocator.donate(extra);
```
TalDerei, line 61: gadgets can donate their spare `extra` tokens back into the pool for `Element::alloc` to consume. specifically, `Boolean::alloc` calls `gate()` directly, gets the spare d wire, and calls `allocator.donate(extra)`, and `PoolAllocator` stores it. and later, `Element::alloc` on the _same_ allocator pops it from the pool and calls `assign_extra` to override a gate's default d wire.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -39,6 +41,14 @@ pub trait Allocator<'dr, D: Driver<'dr>> {
...
     /// and errors returned from it propagate to the caller.
     fn alloc(&mut self, dr: &mut D, value: impl Fn() -> Result<Coeff<D::F>>) -> Result<D::Wire>;
+
+    /// Accepts a spare [`Extra`](ragu_core::drivers::DriverTypes::Extra)
+    /// token from an external gate whose $D$ wire is unconstrained.
+    ///
+    /// Allocators that can pool tokens (like [`PoolAllocator`]) store them
+    /// for future [`alloc`](Self::alloc) calls. The default implementation
+    /// drops the token, keeping the driver's default $D = 0$.
+    fn donate(&mut self, _extra: D::Extra) {}
```
TalDerei, line 58: stored on trait so gadgets don't need to know which allocator it's talking to (for boolean, it inherits the default no-op from the trait).

crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
```diff
@@ -159,10 +159,17 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
...
 
-        let allocator = &mut SimpleAllocator::new();
+        let allocator = &mut PoolAllocator::new();
         let mut unified_output = OutputBuilder::new(witness.map(|w| w.unified));
 
+        // Extract endoscalar early: each of the 128 Boolean gates has a
+        // spare D wire. Donating them to the pool lets subsequent reads
+        // reuse those wires instead of allocating fresh gates.
+        let pre_beta = unified_output.pre_beta.read(dr, allocator)?;
+        let beta_endo = Endoscalar::extract(dr, allocator, pre_beta)?;
```
TalDerei, line 169: `Endoscalar::extract` internally calls a 128 iteration `Boolean::alloc` loop, which internally does this wire donation. then reference https://github.com/tachyon-zcash/ragu/pull/651#discussion_r3082376765.

crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
```diff
@@ -159,10 +159,17 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> MultiStageCircuit<C::CircuitFi
         let query = query.unenforced(dr, witness.as_ref().map(|w| w.query_witness))?;
         let eval = eval.unenforced(dr, witness.as_ref().map(|w| w.eval_witness))?;
 
-        let allocator = &mut SimpleAllocator::new();
+        let allocator = &mut PoolAllocator::new();
```
TalDerei, line 162: as long as this allocator is treated just as a local variable I think (like not sharing `PoolAllocator` across routine boundaries), things are _safe_. things get very messy mentally if you remove that assumption. that's currently how all allocators are processed.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -249,7 +256,7 @@ pub fn multipack<'dr, D: Driver<'dr, F: ff::PrimeField>>(
 
 impl<'dr, D: Driver<'dr>> Consistent<'dr, D> for Boolean<'dr, D> {
     fn enforce_consistent(&self, dr: &mut D) -> Result<()> {
-        Self::alloc(dr, self.value())?.enforce_equal(dr, self)
+        Self::alloc(dr, &mut (), self.value())?.enforce_equal(dr, self)
```
ebfull, line 259: Interesting that this requires an allocator to thread through or else it's wasteful. This is yet another good reason to give the driver the opportunity to hold on to a default or "global" allocator, in a future refactoring, besides the ergonomic improvements.

ebfull (approved):

crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
ebfull: It's not possible to put `Extra` in a gadget without violating the API contract and going nutso with `unsafe` in the process, so it cannot cross routine boundaries.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -137,7 +144,7 @@ impl<E> Default for PoolAllocator<E> {
 
 impl<E> PoolAllocator<E> {
     /// Creates a new `PoolAllocator` with an empty pool.
-    pub fn new() -> Self {
+    pub const fn new() -> Self {
```
TalDerei, line 147: same as `SimpleAllocator`

crates/ragu_primitives/src/boolean.rs
alxiong: I experimented with the "hold a default allocator" idea, but all approaches are unsatisfactory:
- `trait DriverExt: Driver` with `alloc()/donate()` (assuming the driver carries an allocator). Problem: circuit witness functions receive `D: Driver` from `MultiStageCircuit` — the driver type is fixed by the trait hierarchy. We can't add a `D: DriverExt` bound without changing `MultiStageCircuit` and forcing all drivers to implement it, defeating the "optional" intent.
- struct `WithAllocator<'a, D, A>(&'a mut D, A)` implementing Driver. Problem: gadgets created through the wrapper have type `Element<WithAllocator<D,A>>` instead of `Element<D>`. then `Bound<WithAllocator<D,A>, K> != Bound<D, K>`, requiring `CloneWires` conversions at every routine boundary, plus ~80 lines of Driver delegation boilerplate. And since circuit witnesses receive a fixed `D: Driver`, you can only create the wrapper internally, but then your gadget types don't match the function's return type.
-  Add `enforce_consistent_with<A: Allocator>()` as a provided method on `Consistent` (default delegates to `enforce_consistent`), plus `StageGuard::enforced_with()`. This lets enforced stages pool spare d-wires from Boolean enforcement into the circuit's existing allocator. But currently our enforced stages don't have boolean gates, and the endoscaling steps, the Boolean bits are loaded unenforced, and only enforced by the routing circuit. In short, no benefit (i.e. no gate count reduction). 

I suggest we keep as is. (Unless you guys have better idea), just reporting what I tried. 
cc @ebfull @TalDerei

PR #652, opened 2026-04-15
doc: add allocation page to book

LucidSamuel:
- Summary                                                             
  `book/src/guide/gadgets/allocation.md` covering the `Allocator` trait, unit allocator `()`,            
  `SimpleAllocator`, cost accounting with the Simulator, and guidance on choosing an allocator                               
  - Backpatched stale rustdoc link in      
  `simple.md` — `Allocator` now points to the new book page instead of a bare `ragu_primitives::` path that doesn't resolve in mdbook    
  - Added wire-case inconsistency (uppercase vs lowercase gate wires across pages) to `book/FIXME.md` for follow-up                                  
  - Placed after Simple Gadgets in SUMMARY.md per the gadgets section flow


book/src/guide/gadgets/allocation.md
```diff
@@ -0,0 +1,178 @@
+# Allocation
+
+Drivers allocate four wires at a time via [`gate()`]. Circuit code
+frequently needs standalone wires, a single field element, a boolean
+flag, a hash input. The [`Allocator`] trait bridges this gap, living
+in `ragu_primitives` as a separate concern from the core driver.
```
ebfull: > All wires are created in groups whenever new gates are created via [`mul()`] (or, more generally, with [`gate()`].) In some rare cases, [gadgets] need _standalone_ wires that cannot be easily represented as linear combinations of existing wires. The [`Allocator`] trait generalizes over (possibly stateful) strategies for assigning arbitrary data to free wires.

**Why?**
* some readers might not be familiar with `gate` (for good reason) and so mentioning `mul` (and not mentioning the arity of a gate specifically) in the introduction is a good way to ease into the topic, giving us a chance to reintroduce it
* we shouldn't mention booleans as an example of a standalone wire that circuit code needs, because boolean constraints always inline the allocation of their variables! We get them for free; booleans don't need allocation. so while not technically wrong, it's a bad example.
* hash inputs is too vague while also being accurate for only a specific case
* it living in `primitives` should be obvious by the fact this section will be in that subsection of the user guide (see top level review comment)
* this description is more honest, because there are **some** situations where allocation is wasteful if the wire ends up being overdetermined by the constraint system later

book/src/guide/gadgets/allocation.md
```diff
@@ -0,0 +1,178 @@
...
+
+Recall from the [driver types](../drivers/index.md#drivertypes)
+discussion that [`gate()`] produces three wires $(A, B, C)$ and an
+`Extra` token representing $D$, subject to $A \cdot B = C$ and
+$C \cdot D = 0$. When $A = C = 0$, both
+constraints are trivially satisfied: any value of $B$ is compatible
+with $0 \cdot B = 0$, and the auxiliary constraint $0 \cdot D = 0$
+holds for any $D$. A single gate therefore contains **two
+unconstrained slots** — $B$ and $D$ — that an allocator can hand out
+independently.
```
ebfull: These are technical details about how allocators work that could emerge later in the article. My idea for the flow:

1. introduce the trait, show how it's used
2. show the `()` allocator, which shows how it's simple and stateless (but wasteful) to create wires from fake/dummy gates.
3. show the `SimpleAllocator` which shows how `gate` technically gives us a fourth wire we can use, so we can stash it. **this is where you mention gate's four wires and how they work.**
4. show the `PoolAllocator` which shows how gadgets that create constraints like a * b = 0 can then donate the `d` wires to a pool where they can be reused like the `SimpleAllocator` -- the boolean example being the best one.

book/src/guide/gadgets/allocation.md
```diff
@@ -0,0 +1,178 @@
...
+    fn alloc(
+        &mut self,
+        dr: &mut D,
+        value: impl Fn() -> Result<Coeff<D::F>>,
+    ) -> Result<D::Wire>;
+}
+```
+
+[`Element::alloc`] takes an `&mut A` parameter where
+`A: Allocator`, so the caller chooses the allocation strategy:
```
ebfull: This talks about `Element::alloc` too soon. We need some prose first to discuss how the allocator itself is used. It's called with `.alloc()` and provided an assignment function for an individual wire. This can be briefly juxtaposed with the signature of `Driver::mul` for clarity.

`Element::alloc` should be mentioned as an _example_ of what gadgets should do if they need standalone wires like this: they should obtain an allocator from the caller using the generic Allocator trait. I am worried that `Element::alloc` is shown as this instrumental example, it's just the natural (and only) place in our gadget primitives where this kind of thing would first emerge. Anything that would call `Element::alloc` has to thread through an allocator.

ebfull (changes requested):
I don't think this should live in `guide/gadgets` because it's not part of the `Gadget` abstraction. It should live in another part of the user guide. Currently, we have a `guide/standard_library.md` file that should be turned into a subsection (like `Gadgets` or `Drivers` is currently) for documenting `Primitives` more generally. **Allocation is now a primitive.**

The new section should be arranged to accomodate expansion so that it can have individual pages dedicated to subtopics like `Write`/`Buffer` (IO/encoding), the `Simulator`, etc.

book/src/guide/gadgets/allocation.md
LucidSamuel: agreed, I'll add prose after the trait definition walking through `Allocator::alloc()` directly... what the closure does, how it parallels the witness closure in `Driver::mul()`  and then introduce `Element::alloc `later as like a "here's how a real gadget uses this."

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
+# Allocation
+
+All wires are created in groups whenever new gates are created via
+[`mul()`] (or, more generally, with [`gate()`]). In some rare cases,
+gadgets need _standalone_ wires that cannot be easily represented as
+linear combinations of existing wires. The [`Allocator`] trait
+generalizes over (possibly stateful) strategies for obtaining free
+wires to witness arbitrary data.
```
ebfull:
```suggestion
generalizes over (possibly stateful) strategies for assigning
arbitrary data to free wires.
```

ebfull: I can't figure out why I prefer this wording. I think it has something to do with the subject-object distinction; we _witness_ data _by_ assigning _to_ wires -- or something like that.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+```rust,ignore
+pub trait Allocator<'dr, D: Driver<'dr>> {
+    fn alloc(
+        &mut self,
+        dr: &mut D,
+        value: impl Fn() -> Result<Coeff<D::F>>,
+    ) -> Result<D::Wire>;
+
+    fn donate(&mut self, _extra: D::Extra) {}
```
ebfull: I support omitting explanation of this until later in the article, though it doesn't matter that much. I think it's always okay to lie about the full trait description in the book for the purposes of exposition.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+}
+```
+
+`Allocator::alloc` takes a mutable reference to the driver and a
+witness-producing closure, and returns a single wire. The closure
+follows the same purity contract as [`Driver::mul`]: it may be called
+zero or more times and must be side-effect-free. Where `Driver::mul`
```
ebfull: I don't like starting sentences with "Where"

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+produces three wires from a multiplication relationship, `alloc`
+produces one wire carrying an arbitrary value.
+
+The `donate` method accepts a spare
+[`Extra`](ragu_core::drivers::DriverTypes::Extra) token from an
+external gate whose $D$ wire is unconstrained. By default it drops
+the token; advanced allocators override it to pool spare wires for
+reuse.
```
ebfull: If we omit `donate` from the earlier trait description, it can be mentioned later when the `Extra` token actually has use / importance to the reader. It comes out of nowhere here.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+Gadgets that need standalone wires accept a generic `&mut A` where
+`A: Allocator`. For example, [`Element::alloc`] delegates to
+whichever allocator the caller provides:
```
ebfull: We should word this: gadgets that need standalone wires should _prefer_ to accept a generic allocator. It is not always necessary if threading an allocator through the callsite is difficult. (This is subject to change; we may introduce a global allocator that the driver has possession of to simplify these patterns. In any case, the book should document what is currently implemented.)

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+```rust,ignore
```
ebfull: What does `ignore` do in the book? It's needed in doc tests often because those are compiled by default, but the book shouldn't have any CI infra for that yet.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+let x = Element::alloc(dr, &mut allocator, witness)?;
+```
+
+This lets the same gadget code work with any allocation strategy
+the caller chooses.
```
ebfull: more accurately: This allows the caller to determine the optimal allocation strategy for gadgets that do not otherwise care or lack the context to make the best decision themselves.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+        Ok(b)
+    }
+}
+```
+
+Call site:
+
+```rust,ignore
+let x = Element::alloc(dr, &mut (), witness)?;
+```
```
ebfull: The call site varies from one allocator to another in such a trivial, mechanical way that these particular examples pollute the article. Remove.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+One gate per wire, stateless, but wasteful when allocations come in
+pairs.
```
ebfull: overly casual, partial sentence structure. Be precise: "In exchange for being stateless, this allocator creates a gate for every free wire, which is not optimal."

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+## [`SimpleAllocator`] {#simple-allocator}
+
+The unit allocator wastes capacity because each gate allocates four
+wire slots, but [`gate()`] returns three wires plus a spare `Extra`
+token representing the $D$ wire. The gate is subject to
+$A \cdot B = C$ and $C \cdot D = 0$. When $A = C = 0$, both
+constraints are trivially satisfied and both $B$ and $D$ are
+unconstrained. A single gate therefore contains **two free slots**.
```
ebfull: Let the previous section explain the downside of the unit allocator. Defeat it with this paragraph:

this is poor wording, but could inspire a better description:

> Theoretically, an allocator could offer two wires per gate by distributing the `a` and `b` wires of each `mul`, allowing `c` to be a (meaningless) product that is discarded. This is slightly suboptimal, because the trace contains the useless discarded product and is more expensive to manipulate. Instead, the more general [`gate()`] (which creates four wires) can be used, where the `b` and `d` wires are distributed and the `a` and `c` wires are assigned to zero.
>
> [`SimpleAllocator`] exploits this by creating gates, assigning the `b` wire, and then stashing the `d` wire's token for future allocation with the [driver's assistance.]

driver's assistance can link to the `DriverTypes` documentation which describes the `alloc_extra` function (or at least, it _should_ describe it)

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+        } else {
+            let (_, b, _, extra) = dr.gate(
+                || Ok((Coeff::Zero, value()?, Coeff::Zero)),
+            )?;
+            self.stash = Some(extra);
+            Ok(b)
+        }
+    }
+}
+```
```
ebfull: From the user's perspective, all of this code is useless. What it does is useful, to some extent, and the trade-off in terms of performance is relevant. But these details are not. Just say (roughly) "in exchange for being stateful, this is less wasteful."

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+Typical usage:
+
+```rust,ignore
+let allocator = &mut SimpleAllocator::new();
+let a = Element::alloc(dr, allocator, witness_a)?;
+let b = Element::alloc(dr, allocator, witness_b)?;
+// Both a and b share one gate.
+```
```
ebfull, line 89: This is... somewhat okay as an illustration, so my earlier comment about the trivial example might not apply here.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+```admonish info
+Dropping a `SimpleAllocator` that still holds a stashed token is
+safe. The driver already assigned $D = 0$ for that gate, so no
+constraint is violated.
+```
```
ebfull: Let's omit this entirely.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+## [`PoolAllocator`] {#pool-allocator}
+
+Some gadgets allocate a gate whose $D$ wire is unconstrained because
+$C = 0$. Rather than waste that wire, they can _donate_ the `Extra`
```
ebfull: More specifically c is _constrained_ to be zero somehow. Probably via `enforce_zero` but possibly by other methods.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+token to the allocator. [`PoolAllocator`] collects these donated
+tokens and hands them out on future `alloc` calls before falling
+back to new gates.
+
+[`Boolean::alloc`] is the canonical example. The boolean constraint
```
ebfull: LLMs hand out canonical too easily, I like to use it for specific situations like when we're talking about ambiguous representations of ideas or concepts. ("What is canon?") here it's just the only example we have _yet_.

Just say, "As an example, ..."

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+    value: DriverValue<D, bool>,
+) -> Result<Boolean<'dr, D>> {
+    // ... constrains a * (1 - a) = 0 ...
+
+    // The gate's spare D wire is donated to the allocator.
+    allocator.donate(extra);
+
+    Ok(Boolean { value, wire: a })
+}
+```
```
ebfull: hmm, this is 80% signature and 20% useful code.


```suggestion
```rust,ignore
    let (a, b, c, d) = dr.gate(|| /* ... */)?;
    dr.enforce_zero(|lc| lc.add(&c)); // c = 0

    // The gate's spare D wire is donated to the allocator.
    allocator.donate(d);
```
```

File: book/src/guide/primitives/allocation.md [hunk 20]
```diff
@@ -0,0 +1,232 @@
...
+
+Like `SimpleAllocator`, `PoolAllocator` also pairs its own gate
+allocations internally. When the pool is empty and a fresh gate is
+needed, the spare `Extra` from that gate enters the pool for the
+next call.
+
+```admonish info
+Dropping a `PoolAllocator` with tokens still in the pool is safe.
+The driver already assigned $D = 0$ for those gates.
+```
```
ebfull, line 119: I think all of this can be removed, I would hope the reader has gotten the point by now.

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Choosing an Allocator {#choosing}
```
ebfull, line 121: I'm not sure if the recommendations here are complete. There might be different trade-offs. Ultimately the choice of allocator is both a decision a gadget chooses (if they ask to receive one from the caller) and one the circuit developer chooses (if they're asked to provide one) and in each case the decision could be complicated.

ebfull (changes requested):

book/src/SUMMARY.md
```diff
@@ -18,7 +18,8 @@
     - [The Kind! Macro](guide/gadgets/kind.md)
     - [Conversion](guide/gadgets/conversion.md)
   - [Routines](guide/routines.md)
-  - [The Primitives Standard Library](guide/standard_library.md) <!-- todo -->
+  - [The Primitives Standard Library](guide/primitives/index.md)
```
ebfull:
```suggestion
  - [Primitives](guide/primitives/index.md)
```

book/src/guide/primitives/allocation.md
```diff
@@ -0,0 +1,173 @@
+# Allocation
+
+All wires are created in groups whenever new gates are created via
+[`mul()`] (or, more generally, with [`gate()`]). In some rare cases,
+gadgets need _standalone_ wires that cannot be easily represented as
+linear combinations of existing wires. The [`Allocator`] trait
```
ebfull: Ah, even better and more accurate. Perfection is key!

```suggestion
gadgets need _standalone_ wires that cannot be represented as linear
combinations of wires they have access to. The [`Allocator`] trait
```

ebfull (approved):

book/src/guide/gadgets/simple.md
```diff
@@ -25,7 +25,7 @@ serialization using the [`Write`][write-trait] trait.
 
 [`Element::alloc`](ragu_primitives::Element::alloc) can be used to create an
 `Element` which has an assignment based on witness data. It delegates to
-an [`Allocator`](ragu_primitives::Allocator) to obtain the underlying wire.
+an [`Allocator`](../primitives/allocation.md) to obtain the underlying wire.
```
ebfull:
```suggestion
an [allocator](../primitives/allocation.md) to obtain the underlying wire.
```

book/src/SUMMARY.md
```diff
@@ -18,7 +18,8 @@
     - [The Kind! Macro](guide/gadgets/kind.md)
     - [Conversion](guide/gadgets/conversion.md)
   - [Routines](guide/routines.md)
-  - [The Primitives Standard Library](guide/standard_library.md) <!-- todo -->
+  - [Primitives](guide/primitives/index.md)
+    - [Allocation](guide/primitives/allocation.md)
```
ebfull:
```suggestion
  - [Primitives](guide/primitives/index.md) <!-- todo -->
    - [Allocation](guide/primitives/allocation.md) <!-- todo -->
```

ebfull (approved):

ebfull:
Thank you!

ebfull (approved):

PR #654, opened 2026-04-15
chore(primitives): replace `SimpleAllocator` with `Standard`

TalDerei:
`PoolAllocator` is a superior allocator and behaves like `SimpleAllocator` when no wire donations are made.


ebfull:
Can we rename this to `Standard`? (`ragu_primitives::allocator::Standard` seems reasonable.)

ebfull (approved):

TalDerei:
> Can we rename this to `Standard`? (`ragu_primitives::allocator::Standard` seems reasonable.)

addressed

PR #655, opened 2026-04-15
chore(primitives): donate more spare d wires to standard allocator

TalDerei:
Exploiting other gadget candidates where c is structurally zero, similar to the boolean gadget in https://github.com/tachyon-zcash/ragu/pull/643. `is_zero / is_equal` / `conditional_enforce_equal` calls donate an extra token.

### DIFF
--- a/crates/ragu_primitives/src/boolean.rs
+++ b/crates/ragu_primitives/src/boolean.rs
@@ -147,6 +147,7 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
 /// Uses the standard inverse trick for zero checking in arithmetic circuits.
 pub(crate) fn is_zero<'dr, D: Driver<'dr>>(
     dr: &mut D,
+    allocator: &mut impl crate::allocator::Allocator<'dr, D>,
     x: &Element<'dr, D>,
 ) -> Result<Boolean<'dr, D>> {
     // We enforce the constraints:
@@ -164,7 +165,7 @@ pub(crate) fn is_zero<'dr, D: Driver<'dr>>(
     let is_zero = x.value().map(|v| *v == D::F::ZERO);
 
     // Constraint 1: x * is_zero = 0.
-    let (x_wire, is_zero_wire, zero_product) = dr.mul(|| {
+    let (x_wire, is_zero_wire, zero_product, extra) = dr.gate(|| {
         Ok((
             x.value().arbitrary().take(),
             is_zero.coeff().take(),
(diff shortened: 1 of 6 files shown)


crates/ragu_primitives/src/boolean.rs
```diff
@@ -147,6 +163,7 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
 /// Uses the standard inverse trick for zero checking in arithmetic circuits.
 pub(crate) fn is_zero<'dr, D: Driver<'dr>>(
     dr: &mut D,
+    allocator: &mut impl crate::allocator::Allocator<'dr, D>,
```
alxiong: nit: consistently just `use crate::allocator::Allocator` and just `Allocator` here.

alxiong (approved):

crates/ragu_primitives/src/boolean.rs
```diff
@@ -122,8 +123,23 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
...
+        let (cond_wire, diff_wire, zero_wire, extra) = dr.gate(|| {
+            Ok((
+                self.value().coeff().take(),
+                diff.value().arbitrary().take(),
+                Coeff::Zero,
+            ))
+        })?;
+
+        dr.enforce_equal(&cond_wire, self.wire())?;
+        dr.enforce_equal(&diff_wire, diff.wire())?;
```
ebfull: Superior to do `enforce_zero` with the difference lazily computed **in this particular case**, because it's 1) slightly more efficient to avoid creating a virtual wire that is immediately discarded, and 2) doesn't harm readability given how ugly the function is becoming in this PR anyway.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -122,8 +123,23 @@ impl<'dr, D: Driver<'dr>> Boolean<'dr, D> {
...
+        dr.enforce_zero(|lc| lc.add(&zero_wire))?;
+
+        // $C = 0$ makes the $D$ wire unconstrained; donate it.
```
ebfull: LaTeX in comments is pointless since it doesn't render.

ebfull (approved):

crates/ragu_primitives/src/boolean.rs
TalDerei: modified

crates/ragu_primitives/src/boolean.rs
TalDerei: https://github.com/tachyon-zcash/ragu/pull/655/commits/a9334f3b5da42b954cf0e7106674161b35ad0034 mathematically uses the same `s * (a - b) = 0` equation.

Issue #193, opened 2025-12-06
Remove duplicated logic from trivial proof


TalDerei:
The current trivial proof creation _duplicates_ logic from the merge step. As the proof structure becomes more complex, this duplication will grow and require parallel maintenance.

I think trivial proofs should be rather simple:
- Zero polynomials,
- Deterministic blindings (all ones),
- Commitments computed from the above

and we just run merge over them multiple times to produce a trivial random proof (IIRC this matches how [arc-accumulation](https://github.com/arkworks-rs/accumulation) in arkworks handles their dummy / trivial proofs). 

This also, as a side-effect, implies an interesting approach to verification-driven development, where if the trivial proof fails verification after running merge, the failure indicates where a stage or circuit needs bypass logic for the base case. The circuit should allow that specific check to fail _only_ when the header is trivial. Failures become a signal for where to add conditional logic, not where to duplicate more code.

I don't personally think there's a distinction between _trivial_ proofs (zero polynomials with deterministic blinding) and _random_ proofs (random polynomials with random blindings). Our current `random()` method just calls `trivial()`, and the randomness introduced during multiple passes of the `merge()` _should be (?)_ sufficient for zero-knowledge. 

Also, we might want to make `proof` an enum with `Trivial` / `Real` variants, and during merge, trivial cases can substitute random values.

As an aside, maybe the application itself (rather than the library) should be responsible for creating one trivial proof at startup, caching it, and reusing it as the base case for all future PCD proofs?

TalDerei:
closing as completed, but we can reopen if there's anything we need to source from here

PR #656, opened 2026-04-16
pcd: add loading and copying bonding circuits

ebfull:
This will replace #625. It's a vibe-coded alternative used to check my work and changes. I have a more cleaned up branch but it changed significantly and just wanted to get this branch up for some review/analysis/comparison first before I rebased it again.

-----

### Loading and copying circuits

Two new bonding circuits in the nested section work together to constrain every `PointsStage` position.

- **Loading** loads the current step's bridge stages (`preamble`, `s_prime`, `inner_error`, `ab`, `query`, `f`) alongside `PointsStage`, then walks through every `PointsStage` entry and enforces that it equals the corresponding value from somewhere in the bridge stages:
  - Per-child entries match `ChildWitness` fields in the preamble.
  - Current-step entries match their respective bridge stage fields.
  - The `initial` entry matches `BridgeF.native_f`.
  - A `Walker` cursor steps through these inputs sequentially, and its consuming `finish()` method asserts that every input was covered — so if `ChildWitness` or a stage grows a new field without a matching `enforce_equal` call, the circuit fails loudly rather than silently under-constraining.
  - The loading circuit also enforces a relay: `BridgeSPrime.stashed_preamble` must equal `BridgePreamble.native_preamble`. This stashes the current step's native preamble commitment into a different stage type, so that when a parent step's copying circuit later needs to verify it, it can read from `BridgeSPrime` without a wire-position collision on `BridgePreamble`.

- **Copying** handles the other half of the cross-step check. It's parameterized by `Side` (one instance per child proof), and loads the child's bridge stages and `PointsStage` in its own separate trace. It then enforces that each `stashed_*` field in the current step's `ChildWitness` equals the corresponding field from the child's actual bridge stage, with `stashed_p` checked against the last interstitial of the child's `PointsStage`.

- **Why two circuits?** A single trace can't hold both the current and child versions of the same stage type at the same wire positions.

- **Registration and claims.** Both circuits register as bonding objects via `into_bonding_object()` and emit `grouped_bonding_claim`s in `nested::claims::build`, referencing the rx indices of every stage they load.

- **Naming convention.** The `stashed_` prefix uniformly marks fields that are copies placed for cross-circuit verification, distinguishing them from primary introductions like the circuit commitments.

### Proof and claim extensions

- **`ChildWitness`** gains a set of `stashed_*` fields alongside the original fields (application through compute_v) that serve as the primary introduction of child circuit commitments into the nested transcript. The new stashed fields are copies of values that already exist in the child's unified instance or bridge stages, placed here so that loading and copying can enforce them.

- **`BridgeSPrime`** similarly gains a `stashed_preamble` field, populated during fuse from the builder's native preamble commitment — the same convention and concept as the `ChildWitness` stash fields.

- **`ChildStageRx`** is a new struct holding one rx polynomial per stage that the copying circuit loads. It lives in the `Proof` and is populated during fuse via `as_child_stage_rx()`, which clones the relevant polynomials from the child proof so the verifier can later reconstruct the copying circuit's bonding claims.

- **`RxIndex` / plumbing.** Corresponding per-side `Child*` variants are added to `RxIndex` (one per stage in `ChildStageRx`), with matching `Index` impl, `ProofBuilder` storage and setters, and `SingleProofSource` plumbing.

- **`InternalCircuitIndex`.** `Loading` and `Copying(Side)` are registered with their bonding objects wired up in `register_all`.

- **Field order.** `ChildWitness` and `ChildOutput` field order matches the `_10_p` accumulation order so the loading walker can iterate in declaration order.

### Trivial proof

The trivial proof can no longer get away with all-ones polynomials for nested rx, because multi-stage bonding masks require traces that actually satisfy their witness constraints.

- Every bridge rx is now computed via `Stage::rx()` with dummy `host_commitment` witnesses.
- All endoscaling rx (endoscalar stage, points stage, step circuits) are computed by constructing a real `PointsWitness` from the dummy points and tracing each `EndoscalingStep`.
- The computation order matters:
  1. s_prime, inner_error, and f bridges first.
  2. Endoscaling next (Horner-evaluates the dummy points to produce `p_commitment`).
  3. `native_p_poly` (needs the real `p_commitment`).
  4. Preamble bridge last (needs `p_commitment` for `ChildWitness.stashed_p`).
- Both `child_left_stage_rx` and `child_right_stage_rx` are set to match the proof's own stage rx, since a trivial proof is its own child.
- The old `ones_nested` polynomial is eliminated entirely — every nested rx is now properly derived from witness data.

### DIFF
--- a/crates/ragu_pcd/src/fuse/_03_s_prime.rs
+++ b/crates/ragu_pcd/src/fuse/_03_s_prime.rs
@@ -37,6 +37,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             &nested::stages::s_prime::Witness {
                 registry_wx0: native.registry_wx0_commitment,
                 registry_wx1: native.registry_wx1_commitment,
+                stashed_preamble: builder.native_preamble_commitment(),
             },
         )?;
         let bridge_commitment = bridge_rx.commit_to_affine(C::nested_generators(self.params));
--- a/crates/ragu_pcd/src/fuse/mod.rs
+++ b/crates/ragu_pcd/src/fuse/mod.rs
@@ -213,6 +213,10 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
         builder.set_u(*u.value().take());
         builder.set_pre_beta(*pre_beta.value().take());
 
+        // Store children's stage rx polynomials for copying circuit claims.
+        builder.set_child_left_stage_rx(left.as_child_stage_rx());
+        builder.set_child_right_stage_rx(right.as_child_stage_rx());
+
         self.compute_internal_circuits(
             rng,
             &preamble_witness,
--- a/crates/ragu_pcd/src/internal/mod.rs
+++ b/crates/ragu_pcd/src/internal/mod.rs
@@ -30,7 +30,7 @@ pub mod transcript;
 
 /// Identifies which of the two child proofs a component came from.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum Side {
+pub enum Side {
     Left,
     Right,
 }
--- a/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
+++ b/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
@@ -1,0 +1,84 @@
+//! Copying circuit for the nested section.
+//!
+//! Final form relates the current step's preamble to a child proof's
+//! bridge stages, ensuring that child commitments stashed in
+//! [`ChildWitness`](crate::internal::nested::stages::preamble::ChildWitness)
+//! match the actual values committed in the child's bridge stages. One
+//! instance per child proof ([`Side::Left`] and [`Side::Right`]).
+//!
+//! In this initial commit the circuit's stage skeleton is in place but
+//! no `enforce_equal` calls fire — the bonding polynomial folds to zero
+//! so the `grouped_bonding_claim` passes trivially.
+
+use core::marker::PhantomData;
+
+use ragu_arithmetic::CurveAffine;
+use ragu_circuits::{
+    WithAux,
+    polynomials::Rank,
+    staging::{MultiStageCircuit, StageBuilder},
+};
+use ragu_core::{
+    Result,
+    drivers::{Driver, DriverValue},
+    gadgets::Bound,
+};
+
+use crate::internal::{
+    Side,
+    endoscalar::{EndoscalarStage, PointsStage},
+    nested::{NUM_ENDOSCALING_POINTS, stages},
+};
+
+/// Copying circuit that relates the current preamble to a child's stages.
+pub struct Circuit<C: CurveAffine, R: Rank> {
+    side: Side,
+    _marker: PhantomData<(C, R)>,
+}
+
+impl<C: CurveAffine, R: Rank> Circuit<C, R> {
+    pub fn new(side: Side) -> Self {
+        Self {
+            side,
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
+    type Last = stages::eval::Stage<C, R>;
+    type Instance<'source> = ();
+    type Witness<'source> = ();
+    type Output = ();
+    type Aux<'source> = ();
+
+    fn instance<'dr, 'source: 'dr, D: Driver<'dr, F = C::Base>>(
+        &self,
+        _dr: &mut D,
+        _instance: DriverValue<D, ()>,
+    ) -> Result<Bound<'dr, D, ()>> {
+        Ok(())
+    }
+
+    fn witness<'a, 'dr, 'source: 'dr, D: Driver<'dr, F = C::Base>>(
+        &self,
+        dr: StageBuilder<'a, 'dr, D, R, (), Self::Last>,
+        _witness: DriverValue<D, ()>,
+    ) -> Result<WithAux<Bound<'dr, D, ()>, DriverValue<D, ()>>> {
+        // `side` is recorded for future enforce_equal dispatch.
+        let _ = self.side;
+        let dr = dr.skip_stage::<EndoscalarStage>()?;
+        let (_points_guard, dr) = dr.add_stage::<PointsStage<C, NUM_ENDOSCALING_POINTS>>()?;
+        let (_preamble_guard, dr) = dr.add_stage::<stages::preamble::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::s_prime::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::inner_error::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::outer_error::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::ab::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::query::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::f::Stage<C, R>>()?;
+        let dr = dr.skip_stage::<stages::eval::Stage<C, R>>()?;
+        let _dr = dr.finish();
+
+        Ok(WithAux::new((), D::unit()))
+    }
+}
--- a/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
+++ b/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
@@ -21,7 +21,8 @@ use ragu_circuits::{
 use ragu_core::{
     Result,
     drivers::{Driver, DriverValue},
-    gadgets::Bound,
+    gadgets::{Bound, Gadget},
+    maybe::Maybe,
 };
 
 use crate::internal::{
@@ -65,19 +66,37 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
         dr: StageBuilder<'a, 'dr, D, R, (), Self::Last>,
         _witness: DriverValue<D, ()>,
     ) -> Result<WithAux<Bound<'dr, D, ()>, DriverValue<D, ()>>> {
-        // `side` is recorded for future enforce_equal dispatch.
-        let _ = self.side;
         let dr = dr.skip_stage::<EndoscalarStage>()?;
         let (_points_guard, dr) = dr.add_stage::<PointsStage<C, NUM_ENDOSCALING_POINTS>>()?;
-        let (_preamble_guard, dr) = dr.add_stage::<stages::preamble::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::s_prime::Stage<C, R>>()?;
+        let (preamble_guard, dr) = dr.add_stage::<stages::preamble::Stage<C, R>>()?;
+        let (s_prime_guard, dr) = dr.add_stage::<stages::s_prime::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::inner_error::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::outer_error::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::ab::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::query::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::f::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::eval::Stage<C, R>>()?;
-        let _dr = dr.finish();
+        let dr = dr.finish();
+
+        macro_rules! w {
+            () => {
+                _witness.as_ref().map(|_| unreachable!())
+            };
+        }
+        let preamble = preamble_guard.unenforced(dr, w!())?;
+        let s_prime = s_prime_guard.unenforced(dr, w!())?;
+
+        let child = match self.side {
+            Side::Left => &preamble.left,
+            Side::Right => &preamble.right,
+        };
+
+        // stashed_preamble: routes through this step's BridgeSPrime,
+        // which was enforced by loading to equal
+        // preamble.native_preamble at proof-generation time.
+        child
+            .stashed_preamble
+            .enforce_equal(dr, &s_prime.stashed_preamble)?;
 
         Ok(WithAux::new((), D::unit()))
     }
--- a/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
+++ b/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
@@ -1,14 +1,15 @@
 //! Copying circuit for the nested section.
 //!
-//! Final form relates the current step's preamble to a child proof's
-//! bridge stages, ensuring that child commitments stashed in
-//! [`ChildWitness`](crate::internal::nested::stages::preamble::ChildWitness)
-//! match the actual values committed in the child's bridge stages. One
-//! instance per child proof ([`Side::Left`] and [`Side::Right`]).
+//! Relates the current fuse step's preamble to the child proof's stages,
+//! ensuring that child commitments stashed in
+//! [`ChildWitness`](crate::internal::nested::stages::preamble::ChildWitness) match the
+//! actual values committed in the child's bridge stages. One instance per
+//! child proof ([`Side::Left`] and [`Side::Right`]).
 //!
-//! In this initial commit the circuit's stage skeleton is in place but
-//! no `enforce_equal` calls fire — the bonding polynomial folds to zero
-//! so the `grouped_bonding_claim` passes trivially.
+//! The copying circuit loads the child's bridge stages in its own trace
+//! (no wire-position collision with loading, which loads the current
+//! step's stages) and enforces equality between each `ChildWitness`
+//! stash field and the corresponding child bridge stage field.
 
 use core::marker::PhantomData;
 
@@ -78,6 +79,8 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
         let (eval_guard, dr) = dr.add_stage::<stages::eval::Stage<C, R>>()?;
         let dr = dr.finish();
 
+        // Load stage gadgets. Witness values are never accessed — the circuit
+        // only runs during `into_bonding_object` where MaybeKind = Empty.
         macro_rules! w {
             () => {
                 _witness.as_ref().map(|_| unreachable!())
@@ -92,11 +95,15 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
         let query = query_guard.unenforced(dr, w!())?;
         let eval = eval_guard.unenforced(dr, w!())?;
 
+        // Select the child corresponding to this circuit's side.
         let child = match self.side {
             Side::Left => &preamble.left,
             Side::Right => &preamble.right,
         };
 
+        // Enforce that each ChildWitness stash field matches the
+        // corresponding child bridge stage field.
+        //
         // stashed_preamble: routes through this step's BridgeSPrime,
         // which was enforced by loading to equal
         // preamble.native_preamble at proof-generation time.
--- a/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
+++ b/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
@@ -67,7 +67,7 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
         _witness: DriverValue<D, ()>,
     ) -> Result<WithAux<Bound<'dr, D, ()>, DriverValue<D, ()>>> {
         let dr = dr.skip_stage::<EndoscalarStage>()?;
-        let (_points_guard, dr) = dr.add_stage::<PointsStage<C, NUM_ENDOSCALING_POINTS>>()?;
+        let (points_guard, dr) = dr.add_stage::<PointsStage<C, NUM_ENDOSCALING_POINTS>>()?;
         let (preamble_guard, dr) = dr.add_stage::<stages::preamble::Stage<C, R>>()?;
         let (s_prime_guard, dr) = dr.add_stage::<stages::s_prime::Stage<C, R>>()?;
         let (inner_error_guard, dr) = dr.add_stage::<stages::inner_error::Stage<C, R>>()?;
@@ -83,6 +83,7 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
                 _witness.as_ref().map(|_| unreachable!())
             };
         }
+        let points = points_guard.unenforced(dr, w!())?;
         let preamble = preamble_guard.unenforced(dr, w!())?;
         let s_prime = s_prime_guard.unenforced(dr, w!())?;
         let inner_error = inner_error_guard.unenforced(dr, w!())?;
@@ -115,6 +116,14 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
             .stashed_registry_xy
             .enforce_equal(dr, &query.registry_xy)?;
         child.stashed_eval.enforce_equal(dr, &eval.native_eval)?;
+
+        // P: the child's accumulated p commitment is the last interstitial
+        // of the child's PointsStage.
+        let last_interstitial = points
+            .interstitials
+            .last()
+            .expect("NUM_ENDOSCALING_POINTS guarantees >= 1 interstitial");
+        child.stashed_p.enforce_equal(dr, last_interstitial)?;
 
         Ok(WithAux::new((), D::unit()))
     }
--- a/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
+++ b/crates/ragu_pcd/src/internal/nested/circuits/copying.rs
@@ -70,12 +70,12 @@ impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
         let (_points_guard, dr) = dr.add_stage::<PointsStage<C, NUM_ENDOSCALING_POINTS>>()?;
         let (preamble_guard, dr) = dr.add_stage::<stages::preamble::Stage<C, R>>()?;
         let (s_prime_guard, dr) = dr.add_stage::<stages::s_prime::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::inner_error::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::outer_error::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::ab::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::query::Stage<C, R>>()?;
+        let (inner_error_guard, dr) = dr.add_stage::<stages::inner_error::Stage<C, R>>()?;
+        let (outer_error_guard, dr) = dr.add_stage::<stages::outer_error::Stage<C, R>>()?;
+        let (ab_guard, dr) = dr.add_stage::<stages::ab::Stage<C, R>>()?;
+        let (query_guard, dr) = dr.add_stage::<stages::query::Stage<C, R>>()?;
         let dr = dr.skip_stage::<stages::f::Stage<C, R>>()?;
-        let dr = dr.skip_stage::<stages::eval::Stage<C, R>>()?;
+        let (eval_guard, dr) = dr.add_stage::<stages::eval::Stage<C, R>>()?;
         let dr = dr.finish();
 
         macro_rules! w {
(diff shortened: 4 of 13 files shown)


crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -299,6 +299,10 @@ pub(crate) struct ProofBuilder<'params, C: Cycle, R: Rank> {
     bridge_ab_commitment: OnceCell<C::NestedCurve>,
     bridge_query_commitment: OnceCell<C::NestedCurve>,
     bridge_eval_commitment: OnceCell<C::NestedCurve>,
+
+    // Children's stage rx (for copying circuit claims)
+    child_left_stage_rx: Option<super::ChildStageRx<C::ScalarField, R>>,
+    child_right_stage_rx: Option<super::ChildStageRx<C::ScalarField, R>>,
```
ebfull, line 305: Why did these end up `Option`

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,159 @@
...
+            walker.enforce_equal(dr, &child.stashed_p)?;
+        }
+
+        walker.enforce_equal(dr, &s_prime.registry_wx0)?;
+        walker.enforce_equal(dr, &s_prime.registry_wx1)?;
+        walker.enforce_equal(dr, &inner_error.registry_wy)?;
+        walker.enforce_equal(dr, &ab.a)?;
+        walker.enforce_equal(dr, &ab.b)?;
+        walker.enforce_equal(dr, &query.registry_xy)?;
+
```
ebfull, line 156: `walker` should be exhausted, so we should `finish` it and assert `inputs` is empty.

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,159 @@
...
+//! [`ChildWitness`](super::stages::preamble::ChildWitness) stash fields
+//! and current-step bridge stage fields) plus `initial` (matched against
+//! `BridgeF.native_f`). The accumulation walk mirrors `compute_p` in
+//! `_10_p` so that correctness can be verified by visual comparison.
+//!
+//! Also enforces: `BridgeSPrime.stashed_preamble` ==
+//! `BridgePreamble.native_preamble`, stashing the current step's native
+//! preamble so that a future
+//! [`copying`](super::circuits::copying) circuit can verify it from the
+//! child's `BridgeSPrime` without wire-position collision.
```
ebfull: All comments making specific claims about the numbers of things should be adjusted to be more future-proof.

crates/ragu_pcd/src/proof/builder.rs
ebfull: Oh right, they're the rx.

crates/ragu_pcd/src/fuse/_03_s_prime.rs
```diff
@@ -37,6 +37,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             &nested::stages::s_prime::Witness {
                 registry_wx0: native.registry_wx0_commitment,
                 registry_wx1: native.registry_wx1_commitment,
+                stashed_preamble: builder.native_preamble_commitment(),
```
TalDerei, line 40: ACK – moving `stashed_preamble` under `s_prime`, meaning `as_child_stage_rx` will have one extra field.

crates/ragu_pcd/src/fuse/mod.rs
```diff
@@ -213,6 +213,10 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
         builder.set_u(*u.value().take());
         builder.set_pre_beta(*pre_beta.value().take());
 
+        // Store children's stage rx polynomials for copying circuit claims.
+        builder.set_child_left_stage_rx(left.as_child_stage_rx());
+        builder.set_child_right_stage_rx(right.as_child_stage_rx());
+
```
TalDerei, line 219: ACK – parent keeps copies of the child's own rx polynomials, otherwise it gets lost. references https://github.com/tachyon-zcash/ragu/pull/625/changes#diff-6497f64ee602432a1607e0f98b36b7c9a34770b3ed4c207fae87781a116fafcdR44-R61 (now lifted to top-level proof fields, aligns with the flattened proof structure).

crates/ragu_pcd/src/internal/nested/circuits/copying.rs
```diff
@@ -0,0 +1,132 @@
...
+//!
+//! Relates the current fuse step's preamble to the child proof's stages,
+//! ensuring that child commitments stashed in [`ChildWitness`] match the
+//! actual values committed in the child's bridge stages. One instance per
+//! child proof ([`Side::Left`] and [`Side::Right`]).
+//!
+//! The copying circuit loads the child's bridge stages in its own trace
+//! (no wire-position collision with loading, which loads the current
+//! step's stages) and enforces equality between each `ChildWitness`
+//! stash field and the corresponding child bridge stage field.
```
TalDerei, line 12: ACK – references https://github.com/tachyon-zcash/ragu/pull/625/changes#r3012711287.

crates/ragu_pcd/src/internal/nested/circuits/copying.rs
```diff
@@ -0,0 +1,132 @@
...
+        let eval = eval_guard.unenforced(dr, w!())?;
+
+        // Select the child corresponding to this circuit's side.
+        let child = match self.side {
+            Side::Left => &preamble.left,
+            Side::Right => &preamble.right,
+        };
+
+        // Enforce that each ChildWitness stash field matches the
+        // corresponding child bridge stage field.
```
TalDerei, line 105: nit; richer internal docs, https://github.com/tachyon-zcash/ragu/pull/625/changes#r3012598055.

ACK on actual one-to-one mapping copying enforces (despite the asymmetry of 9 stashed fields versus 7 `ChildStageRx` polys).

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl<C: CurveAffine, R: Rank> Circuit<C, R> {
+    pub fn new() -> Self {
+        Self {
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<C: CurveAffine, R: Rank> MultiStageCircuit<C::Base, R> for Circuit<C, R> {
```
TalDerei, line 82: ACK – references https://github.com/tachyon-zcash/ragu/pull/625/changes#r3012708974.

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,159 @@
...
+        let preamble = preamble_guard.unenforced(dr, w!())?;
+        let s_prime = s_prime_guard.unenforced(dr, w!())?;
+        let inner_error = inner_error_guard.unenforced(dr, w!())?;
+        let ab = ab_guard.unenforced(dr, w!())?;
+        let query = query_guard.unenforced(dr, w!())?;
+        let f_stage = f_guard.unenforced(dr, w!())?;
+
+        // Walk through PointsStage inputs, mirroring the accumulation order
+        // in `compute_p` (_10_p.rs).
+        let mut walker = Walker::new(&points);
```
TalDerei, line 131: why aren't we coupling the walker to `RxIndex::ALL`? verified it's the same 37-point enforcement as previous impl though.

crates/ragu_pcd/src/internal/nested/claims.rs
```diff
@@ -139,6 +139,35 @@ where
...
+                    .zip(source.rx(RxIndex::ChildBridgeInnerError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeOuterError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeAB(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeQuery(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeEval(side)))
+                    .map(|(((((((bp, cp), cs), ci), co), ca), cq), ce)| {
+                        [bp, cp, cs, ci, co, ca, cq, ce].into_iter()
+                    });
+                processor.grouped_bonding_claim(id, groups)?;
+            }
```
TalDerei, line 170: ACK – claim wiring, but still wondering about https://github.com/tachyon-zcash/ragu/pull/625/changes#r3012729232.

crates/ragu_pcd/src/internal/nested/mod.rs
```diff
@@ -246,6 +286,14 @@ pub fn register_all<'params, C: Cycle, R: Rank>(
...
                 registry.register_bonding(stages::eval::Stage::<C::HostCurve, R>::mask()?)
             }
+            Loading => {
+                let circuit = circuits::loading::Circuit::<C::HostCurve, R>::new();
+                registry.register_bonding(MultiStage::new(circuit).into_bonding_object()?)
+            }
+            Copying(side) => {
+                let circuit = circuits::copying::Circuit::<C::HostCurve, R>::new(side);
+                registry.register_bonding(MultiStage::new(circuit).into_bonding_object()?)
+            }
```
TalDerei, line 305: ACK – materializing bonding polynomials outside of tests.

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -427,27 +469,206 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> crate::Application<'_, C, R, H
...
-        builder.set_bridge_f_rx(ones_nested.clone(), bridge_commitment);
-        // Cached bridges (outer_error, ab, query, eval) are computed lazily by the builder.
-
-        // Nested endoscaling
-        builder.set_nested_endoscaling_step_rxs(vec![
-            ones_nested.clone();
-            NumStepsLen::<NUM_ENDOSCALING_POINTS>::len()
-        ]);
-        builder.set_nested_endoscalar_rx(ones_nested.clone());
-        builder.set_nested_points_rx(ones_nested);
```
TalDerei, line 450: we replaced this with duplicated logic from fuse, reminding me of https://github.com/tachyon-zcash/ragu/issues/193, which doesn't seem ideal and may diverge.

crates/ragu_pcd/src/internal/nested/circuits/copying.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) -> Result<Bound<'dr, D, ()>> {
+        Ok(())
+    }
+
+    fn witness<'a, 'dr, 'source: 'dr, D: Driver<'dr, F = C::Base>>(
+        &self,
+        dr: StageBuilder<'a, 'dr, D, R, (), Self::Last>,
+        _witness: DriverValue<D, ()>,
+    ) -> Result<WithAux<Bound<'dr, D, ()>, DriverValue<D, ()>>> {
+        let dr = dr.skip_stage::<EndoscalarStage>()?;
```
TalDerei, line 70: ACK – endoscaling gap per https://github.com/tachyon-zcash/ragu/pull/353#discussion_r2702571715 will be follow-up work.

crates/ragu_pcd/src/internal/nested/claims.rs
```diff
@@ -139,6 +139,35 @@ where
...
+                    .zip(source.rx(RxIndex::ChildBridgeSPrime(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeInnerError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeOuterError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeAB(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeQuery(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeEval(side)))
+                    .map(|(((((((bp, cp), cs), ci), co), ca), cq), ce)| {
+                        [bp, cp, cs, ci, co, ca, cq, ce].into_iter()
+                    });
+                processor.grouped_bonding_claim(id, groups)?;
```
TalDerei, line 169: nit; cosmetic, doesn't match the stage-add order, unlike loading.

crates/ragu_pcd/src/internal/mod.rs
TalDerei, line 1: optional enum for child bridge stages per https://github.com/tachyon-zcash/ragu/pull/625#discussion_r3012645945.

TalDerei (commented):
structural consistency pass against https://github.com/tachyon-zcash/ragu/pull/625.

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
ebfull: Fixing.

crates/ragu_pcd/src/internal/nested/stages/preamble.rs
```diff
@@ -83,6 +122,57 @@ pub struct ChildOutput<'dr, D: Driver<'dr>, C: CurveAffine<Base = D::F>> {
...
+            InnerCollapse => &self.inner_collapse,
+            OuterCollapse => &self.outer_collapse,
+            ComputeV => &self.compute_v,
+            Preamble => &self.stashed_preamble,
+            InnerError => &self.stashed_inner_error,
+            OuterError => &self.stashed_outer_error,
+            Query => &self.stashed_query,
+            Eval => &self.stashed_eval,
+        }
+    }
```
TalDerei, line 175: we now couple the walker to `RxIndex::ALL` per https://github.com/tachyon-zcash/ragu/pull/656#discussion_r3095254549 to cover 11 `RxIndex` child fields, but I think we should extend this also to other things which are still enumerated by hand:

- 4 child fields (`stashed_ab_a, stashed_ab_b, stashed_registry_xy, stashed_p`)
- (maybe) 6 current step bridge inputs (`s_prime.registry_wx0/1, inner_error.registry_wy, ab.a/b, query.registry_xy`)

helper on `ChildOutput` impl:

```ts
pub fn points_for_p(&self) -> impl Iterator<Item = &Point<'dr, D, C>> {
    RxIndex::ALL.iter().map(move |&id| &self[id]).chain([
        &self.stashed_ab_a,
        &self.stashed_ab_b,
        &self.stashed_registry_xy,
        &self.stashed_p,
    ])
}
```

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,164 @@
...
+            walker.enforce_equal(dr, &child.stashed_registry_xy)?;
+            walker.enforce_equal(dr, &child.stashed_p)?;
+        }
+
+        walker.enforce_equal(dr, &s_prime.registry_wx0)?;
+        walker.enforce_equal(dr, &s_prime.registry_wx1)?;
+        walker.enforce_equal(dr, &inner_error.registry_wy)?;
+        walker.enforce_equal(dr, &ab.a)?;
+        walker.enforce_equal(dr, &ab.b)?;
+        walker.enforce_equal(dr, &query.registry_xy)?;
```
TalDerei, line 148: per the suggestion [above](https://github.com/tachyon-zcash/ragu/pull/656#discussion_r3118212297).

```suggestion
        for child in [&preamble.left, &preamble.right] {
            for point in child.points_for_p() {
                walker.enforce_equal(dr, point)?;
            }
        }

        for point in [
            &s_prime.registry_wx0,
            &s_prime.registry_wx1,
            &inner_error.registry_wy,
            &ab.a,
            &ab.b,
            &query.registry_xy,
        ] {
            walker.enforce_equal(dr, point)?;
        }
```

crates/ragu_pcd/src/internal/nested/claims.rs
```diff
@@ -139,6 +139,35 @@ where
...
+                    .zip(source.rx(RxIndex::BridgePreamble))
+                    .zip(source.rx(RxIndex::ChildBridgeSPrime(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeInnerError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeOuterError(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeAB(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeQuery(side)))
+                    .zip(source.rx(RxIndex::ChildBridgeEval(side)))
+                    .map(|(((((((cp, bp), cs), ci), co), ca), cq), ce)| {
+                        [cp, bp, cs, ci, co, ca, cq, ce].into_iter()
+                    });
```
TalDerei, line 168: `itertools::multizip` (dep already in zebra) is marginally better by flattening the map arg, but not worth it. a crude manual zip impl is _optional_:

```ts
  pub fn zip_rx<S: Source, const N: usize>(                                                                                                                                                                                                          
      source: &S,                                                                                                                                                                                                                                    
      components: [S::RxComponent; N],                                                                                                                                                                                                               
  ) -> impl Iterator<Item = [S::Rx; N]>                                                                                                                                                                                                              
  where                                                                                                          
      S::Rx: Clone,   
  {                                                                                                                                                                                                                                                  
      // Gather rxs per component, then stream row-by-row (transpose).
      let columns: [Vec<S::Rx>; N] = components.map(|c| source.rx(c).collect());                                                                                                                                                                     
      let rows = columns[0].len();                                                                                                                                                                                                                   
      (0..rows).map(move |r| core::array::from_fn(|c| columns[c][r].clone()))
  }                                                                                                                                                                                                                                                   
```

```ts
let groups = zip_rx(
    source,
    [
        RxIndex::ChildPointsStage(side),
        RxIndex::BridgePreamble,
        RxIndex::ChildBridgeSPrime(side),
        RxIndex::ChildBridgeInnerError(side),
        RxIndex::ChildBridgeOuterError(side),
        RxIndex::ChildBridgeAB(side),
        RxIndex::ChildBridgeQuery(side),
        RxIndex::ChildBridgeEval(side),
    ],
)
.map(<[_; 8]>::into_iter);
```

crates/ragu_pcd/src/internal/nested/mod.rs
```diff
@@ -1,7 +1,17 @@
...
-//! commitment accumulation was computed correctly via Horner's rule.
+//! Contains three groups of circuits:
+//!
+//! - **Endoscaling**: verifies that the commitment accumulation
+//!   in `compute_p` was computed correctly via Horner's rule.
+//! - **Loading**: enforces consistency between [`PointsStage`]
+//!   inputs and the bridge stage commitments for the current step.
+//! - **Copying**: enforces that [`ChildWitness`] stash fields in
+//!   `BridgePreamble` match the corresponding child proof's bridge
+//!   stage contents.
```
TalDerei, line 11: you mentioned wanting a graph / visualization of the stages linked across fuse steps for the exhaustiveness of the last pieces, did you get to that? It'd be nice to include that somewhere.

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -71,6 +76,36 @@ impl<C: Cycle, R: Rank, H: Header<C::CircuitField>> Clone for Pcd<C, R, H> {
...
+    pub bridge_query: sparse::Polynomial<F, R>,
+    pub bridge_eval: sparse::Polynomial<F, R>,
+}
+
+impl<C: Cycle, R: Rank> Proof<C, R> {
+    /// Extract stage rx polynomials from this proof for storage as child
+    /// data in a parent proof.
+    pub(crate) fn as_child_stage_rx(&self) -> ChildStageRx<C::ScalarField, R> {
+        ChildStageRx {
+            points_stage: self.nested_points_rx.clone(),
```
TalDerei, line 97: can we add a todo comment here that we'll eventually wrap these child poly's in `Arc` to avoid cloning.

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -427,27 +470,206 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> crate::Application<'_, C, R, H
...
+                &nested::stages::f::Witness {
+                    native_f: host_commitment,
+                },
+            )
+            .expect("trivial f rx");
+            let commitment = rx.commit_to_affine(nested_gen);
+            builder.set_bridge_f_rx(rx, commitment);
+        }
+
+        // Nested endoscaling: compute properly from dummy commitments so
```
TalDerei, line 527: https://github.com/tachyon-zcash/ragu/pull/656#discussion_r3095387957 is unaddressed; i'd extract the inline endoscaling logic here into a `compute_endoscaling` method, and then refactor `compute_p` (in `_10_p.rs`) to use the same helper.

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
```diff
@@ -0,0 +1,164 @@
...
+        let dr = dr.skip_stage::<stages::eval::Stage<C, R>>()?;
+        let dr = dr.finish();
+
+        // Load stage gadgets. Witness values are never accessed — the circuit
+        // only runs during `into_bonding_object` where MaybeKind = Empty.
+        macro_rules! w {
+            () => {
+                _witness.as_ref().map(|_| unreachable!())
+            };
+        }
```
TalDerei, line 120: duplicate macro across loading / copying, can probably be hoisted as single top-level macro.

TalDerei (approved):
ACK second pass; rebase + more granular commit splitting for reviewability, not primarily content diff changes. left some suggestions

crates/ragu_pcd/src/internal/nested/stages/preamble.rs
ebfull: Is this proposed `points_for_p` being used in both `_10_p` and in the loading circuit? If so, I agree. Otherwise this is slightly premature.

crates/ragu_pcd/src/internal/nested/claims.rs
ebfull: If we bring in `itertools`, which I'm not that opposed to, there are things in `fold_revdot` that can be replaced. But we should do that in another PR.

crates/ragu_pcd/src/internal/nested/mod.rs
ebfull: Work in progress.

crates/ragu_pcd/src/proof/mod.rs
ebfull: I'd love to see if this could be simplified; could you do this in an isolated commit for easier review? I suspect you're right, I just don't know where the ideal location for this helper/method.

crates/ragu_pcd/src/internal/nested/stages/preamble.rs
TalDerei: yeah that was the intention

crates/ragu_pcd/src/internal/nested/circuits/loading.rs
ebfull: Possibly... this is one of those things I'd do in a local closure if I could, but it's not possible to define such things in Rust. Extracting it to a function/macro in another file requires the reader to figure out what it does, which is worse in some ways. I would just say leave it.

crates/ragu_pcd/src/internal/nested/stages/preamble.rs
TalDerei: in hindsight, there wasn't really a good way to collapse the callsites, but parallel helpers seem organizationally nicer. we can revert or play with the location of those helpers.

crates/ragu_pcd/src/internal/nested/claims.rs
TalDerei: tracking in https://github.com/tachyon-zcash/ragu/issues/685.

crates/ragu_pcd/src/proof/mod.rs
TalDerei: added comment

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -586,62 +650,25 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> crate::Application<'_, C, R, H
...
-            assert_eq!(points.len(), NUM_ENDOSCALING_POINTS);
+            points.push(registry_xy_commitment); // native_registry_xy
 
-            let witness =
-                PointsWitness::<C::HostCurve, NUM_ENDOSCALING_POINTS>::new(beta_endo, &points);
-
-            let endoscalar_rx = <EndoscalarStage as StageExt<C::ScalarField, R>>::rx(
-                C::ScalarField::ONE,
+            let mut trivial_rng =
+                <rand::rngs::StdRng as rand::SeedableRng>::from_seed([0u8; 32]);
```
TalDerei: https://github.com/tachyon-zcash/ragu/pull/656/commits/578a35fe6f5dd8d8bf6302c72af158d6664280ed (references https://github.com/tachyon-zcash/ragu/pull/656#discussion_r3138672306); this is now a shared rng (after extracting a shared utility) instead of per-step fresh seed, previously

```ts
let step_rx = self
    .nested_registry
    .assemble(
        &step_trace,
        nested::InternalCircuitIndex::EndoscalingStep(step as u32).circuit_index(),
        &mut <rand::rngs::StdRng as rand::SeedableRng>::from_seed([0u8; 32]),
    )
    .expect("trivial step rx");
```

crates/ragu_pcd/src/internal/nested/mod.rs
```diff
@@ -139,6 +139,40 @@ impl InternalCircuitIndex {
...
+    /// changes the nested registry digest.
+    pub const ALL: [Self; 6] = [
+        Self::SPrime,
+        Self::InnerError,
+        Self::OuterError,
+        Self::AB,
+        Self::Query,
+        Self::Eval,
+    ];
+}
```
TalDerei, line 174: https://github.com/tachyon-zcash/ragu/pull/656/changes/0fac74907f157039c3625779303e9b63b3e767b4, references:
- https://github.com/tachyon-zcash/ragu/pull/625#discussion_r3012645945
- https://github.com/tachyon-zcash/ragu/pull/625#discussion_r3012652329 
- https://github.com/tachyon-zcash/ragu/pull/625#discussion_r3012654823

registry digest and `RxIndex::ALL` ordering is preserved between recent commits.

crates/ragu_pcd/src/internal/nested/stages/preamble.rs
TalDerei: https://github.com/tachyon-zcash/ragu/pull/656/changes/2a241b6805e04ef36e30ae717c5c7cb0b97757c0

TalDerei:
ACK third pass addressing nits; slightly expands scope by introducing helpers and enums to encapsulate shared logic (these are an optional changes).

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -388,6 +388,40 @@ impl<C: Cycle, R: Rank> Proof<C, R> {
...
+                    &self[RxComponent::AbB],
+                    self.native_commitment(RxComponent::AbB),
+                ),
+                (
+                    self.native_registry_xy_poly(),
+                    self.native_registry_xy_commitment(),
+                ),
+                (self.native_p_poly(), self.native_p_commitment()),
+            ])
+    }
```
ebfull: I don't see how this helper has assisted, it just moved the logic from one file to another, and made it harder to read.

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -471,6 +472,69 @@ impl<C: Cycle, R: Rank> Proof<C, R> {
 }
 
 impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> crate::Application<'_, C, R, HEADER_SIZE> {
+    /// Runs endoscaling over the host-curve commitments that feed
+    /// `PointsStage`, in the order `compute_p` (`_10_p.rs`)
+    /// accumulates them. Writes `nested_endoscalar_rx`,
```
ebfull, line 452: This is a bit tautological, since `compute_p` literally calls this.

crates/ragu_pcd/src/internal/nested/mod.rs
```diff
@@ -214,18 +239,15 @@ impl RxIndex {
...
-        push(&mut slots, &mut c, Self::ChildBridgeEval(Side::Left));
-        push(&mut slots, &mut c, Self::ChildBridgeEval(Side::Right));
+        {
+            let mut i = 0;
+            while i < ChildBridgeKind::ALL.len() {
+                let kind = ChildBridgeKind::ALL[i];
+                push(&mut slots, &mut c, Self::ChildBridge(kind, Side::Left));
+                push(&mut slots, &mut c, Self::ChildBridge(kind, Side::Right));
+                i += 1;
+            }
```
ebfull, line 249: Why isn't this a loop over `ChildBridgeKind::ALL`?

crates/ragu_pcd/src/proof/mod.rs
ebfull: I can see that it kind of helps because the helpers are colocated in the same file, so I'm +0 on this.

TalDerei: yeah that's fair, more of a stylistic preference for readability, will revert.

crates/ragu_pcd/src/proof/mod.rs
TalDerei: ditto, https://github.com/tachyon-zcash/ragu/pull/656#discussion_r3139717890.

crates/ragu_pcd/src/internal/nested/mod.rs
TalDerei: tried that originally, but borrow checker was complaining with "for loops on [T; N] aren't supported in const fn on Rust 1.90 stable".

ebfull: Oh sorry! I missed that this was a `const fn`.

crates/ragu_pcd/src/proof/mod.rs
TalDerei: @ebfull ACK on this judgement call would be good.

TalDerei: side-effect is each trace assembly call previously got the same alpha, and now gets a different one, which seems fine.

Issue #665, opened 2026-04-19
Reduce revdot claims to polynomial openings


TalDerei:
Implement a generic utility in the `ragu_arithmetic` crate: decomposition trick for reducing a revdot claim into a polynomial query, described in [section 2.2.3](https://tachyon.z.cash/ragu/protocol/prelim/structured_vectors.html) ("Structured Vectors") of our book. This will become applicable once we incorporate succinctness.

Synthesizing the construction from the book: our univariate PCS (bulletproofs-style IPA) can't commit to the product polynomial `a * b` directly because of the degree bounds, so we split into p, q parts that fit in the PCS. The revdot claim `revdot(a, b) = c` is equivalent to "the (n−1)th coefficient of `a * b` equals c", and the split is constructed so the coefficient lands at `p(0)`. That lets us reduce revdot claims into polynomial queries: a consistency check that p, q come from `a * b` and the claim `p(0) = c`. 

Follow-up work from this issue will involve plumbing this utility into the IPA module (ie. the second half of this issue tracked in https://github.com/tachyon-zcash/ragu/issues/668).

TalDerei:
closed per #667

PR #667, opened 2026-04-20
feat(arithmetic): polynomial decomposition

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/665; Implements arithmetic utilities for polynomial multiplication (`poly_mul`) and decomposition (`decomp_poly`).

### DIFF
--- a/crates/ragu_arithmetic/src/util.rs
+++ b/crates/ragu_arithmetic/src/util.rs
@@ -333,12 +333,7 @@ pub fn poly_mul<F: PrimeField>(a: &[F], b: &[F]) -> Vec<F> {
 /// Writes $c(X) = a(X) \cdot b(X)$ into `out`; `scratch` holds $b$'s evaluation
 /// form and is left unspecified on return. Both are resized as needed — reusing
 /// them across many calls avoids per-call allocations.
-pub fn poly_mul_into<F: PrimeField>(
-    a: &[F],
-    b: &[F],
-    out: &mut Vec<F>,
-    scratch: &mut Vec<F>,
-) {
+pub fn poly_mul_into<F: PrimeField>(a: &[F], b: &[F], out: &mut Vec<F>, scratch: &mut Vec<F>) {
     out.clear();
     scratch.clear();
 
--- a/crates/ragu_arithmetic/src/util.rs
+++ b/crates/ragu_arithmetic/src/util.rs
@@ -320,9 +320,12 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
 
 /// Writes $c(X) = a(X) \cdot b(X)$ into `out` via FFT.
 ///
-/// `out.len()` becomes `a.len() + b.len() - 1` on return, but capacity is
-/// retained at `2 * next_power_of_two(len)` for reuse; one-shot callers that
-/// care about the slack should `shrink_to_fit` at the handoff.
+/// If either input is empty, `out` is cleared and the function returns.
+/// Otherwise `out.len()` becomes `a.len() + b.len() - 1` on return, with
+/// capacity left at `≥ 2 · next_power_of_two(a.len() + b.len() - 1)`
+/// (the function uses the tail as FFT scratch); one-shot callers that
+/// care about the slack should `shrink_to_fit` before handing the buffer
+/// to consumers.
 pub fn poly_mul<F: PrimeField>(a: &[F], b: &[F], out: &mut Vec<F>) {
     out.clear();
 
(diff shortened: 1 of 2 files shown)


crates/ragu_arithmetic/src/util.rs
```diff
@@ -319,6 +319,84 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
...
+    poly_mul_into(a, b, &mut out, &mut scratch);
+    out
+}
+
+/// In-place variant of [`poly_mul`] that reuses caller-provided buffers.
+///
+/// Writes $c(X) = a(X) \cdot b(X)$ into `out`; `scratch` holds $b$'s evaluation
+/// form and is left unspecified on return. Both are resized as needed — reusing
+/// them across many calls avoids per-call allocations.
+pub fn poly_mul_into<F: PrimeField>(a: &[F], b: &[F], out: &mut Vec<F>, scratch: &mut Vec<F>) {
```
TalDerei: inlined impl from `poly_with_roots` was modified and extracted into `poly_mul_into` as a dedicated utility.

crates/ragu_arithmetic/src/util.rs
```diff
@@ -319,6 +319,84 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
...
+
+    domain.ifft(out);
+    out.truncate(result_len);
+}
+
+/// Computes the decomposition $(p, q)$ of $a(X) \cdot b(X)$, with $p$ the
+/// reversed lower half of $c = a \cdot b$ and $q$ its upper half zero-padded
+/// to length $n$. The pair satisfies
+///
+/// $$ a(x) \cdot b(x) = x^{n-1} p(x^{-1}) + x^n q(x) $$
```
alxiong: - add links to https://tachyon.z.cash/ragu/protocol/prelim/structured_vectors.html#reduction-to-polynomial-queries, which offers more details.
- rename `decomp_poly` to `decomp_product_poly` 
- I'd further highlight the intent and use case for this decomposition, the goal is to capture `revdot(a, b)` in $p(0)$. (current language only mentioned this relation, but we should phrase it as the goal, so that it's clear why function contract demands equal length `a` and `b`. Since this decomposition in algebraic form is possible for variable degree multiplicants.

crates/ragu_arithmetic/src/util.rs
```diff
@@ -319,6 +319,84 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
...
+    if a.is_empty() {
+        return (vec![], vec![]);
+    }
+
+    // `c` has length `2n - 1`. Split off the upper `n - 1` coefficients into `q`
+    // and zero-pad to `n`; reverse the remaining lower `n` coefficients in place
+    // to form `p`.
+    let n = a.len();
+    let mut c = poly_mul(a, b);
+    let mut q = c.split_off(n);
```
alxiong:
```suggestion
    let mut c = poly_mul(a, b);
    assert_eq!(c.len(), 2 * n - 1);
    let mut q = c.split_off(n);
```

crates/ragu_arithmetic/src/util.rs
```diff
@@ -319,6 +319,84 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
...
+    q.push(F::ZERO);
```
alxiong: personally, i'm not sure why we need to pad `q`, especially pad its highest degree coefficient.
in our `Polynomial` (and most impl out there), leading zeros on the highest coefficients are trimmed anyways. 

for clarity, we can include this contract in the rustdoc, informing the expected length of decomposed `p, q`.

crates/ragu_arithmetic/src/util.rs
```diff
@@ -334,40 +412,19 @@ pub fn poly_with_roots<F: PrimeField>(roots: &[F]) -> Vec<F> {
...
-            scratch2[poly2_len..n].fill(F::ZERO);
-            domain.fft(&mut scratch2[..n]);
-
-            for j in 0..n {
-                scratch1[j] *= scratch2[j];
-            }
-
-            domain.ifft(&mut scratch1[..n]);
-
+            poly_mul_into(&polys[2 * i], &polys[2 * i + 1], &mut out, &mut scratch);
```
alxiong: imo, we passing in scratch, an internal buffer, in public api feels like an anti-pattern. we can pass in a longer `out` and use that as its internal scratch.

alxiong (approved):
Address all my own comments.

crates/ragu_arithmetic/src/util.rs
TalDerei: well this was meant to be an internal helper (not part of the public API) that carried a minor perf improvement to maintain parity with previous impl, but unifying it into a single method and stripping the scratch seems fine.

crates/ragu_arithmetic/src/util.rs
TalDerei: yeah fair, that was extraneous.

crates/ragu_arithmetic/src/util.rs
TalDerei: good set of suggestions, thanks for fixing!

TalDerei:
> Address all my own comments.

ACK changes 👍

crates/ragu_arithmetic/src/util.rs
```diff
@@ -318,6 +318,121 @@ pub fn geosum<F: Field>(mut r: F, mut m: usize) -> F {
     sum
 }
 
+/// Writes $c(X) = a(X) \cdot b(X)$ into `out` via FFT.
```
ebfull, line 321: The fact an FFT is used is irrelevant to the caller and may even change later. FFTs are not always more efficient.

ebfull (approved):

Issue #613, opened 2026-03-24
Possibly consolidate driver synthesis passes


TalDerei:
References https://github.com/tachyon-zcash/ragu/pull/610#discussion_r2976851185. We may want to consolidate some synthesis passes in certain parts of the code under a higher-level abstraction (maybe some kind of driver combinator). Filing this purely as an informational reminder.

PR #669, opened 2026-04-20
book: routine implementation page

alxiong:
rewrite of #599 (imo, supersedes it since we modify on top of that)

There are some key technical differences:

- in algebraic description, I included system contribution $g_3$
- rename shallow fingerprint to "base fingerprint" to signify what it implies, and stick to "deep fingerprint" (rather than `MemoFingerprint` since both fingerprints implies memoization opportunities)
- deep memoization redefines internal contribution (counting the contribution of output wires of child subroutines), which is something we gloss over in previous attempt, i highlight in admonish box. 
- add algebraic description for inter-circuit aligned segment 
- envisioned that for caching polynomial contribution during actual memoization, we need to expand the `enum Wire` type to have an `::Interface()` variant and update all drivers to handle that case, so that we can drop their contribution.

### DIFF
--- a/book/src/implementation/routines.md
+++ b/book/src/implementation/routines.md
@@ -4,7 +4,7 @@
 that satisfy a simple function-like interface: they take a single [`Input`]
 gadget and a [`Driver`] handle, and return a single [`Output`] gadget. They are
 permitted to do anything that normal circuit code can do with a
-[driver](../guide/drivers.md), such as making gates ([`mul`], [`alloc`]) and
+[driver](../guide/drivers.md), such as making gates ([`gate`], [`mul`], [`alloc`]) and
 constraining their wires ([`enforce_zero`]).
 
 It is possible to invoke them by manually calling their [`Routine::execute`]
@@ -32,7 +32,7 @@ algebraic properties at routine boundaries.
 Routines exploit the fact that contiguous sections of code tend to have
 algebraically convenient structure in the resulting wiring polynomial $s(X, Y)$:
 
-* `mul` advances an $i$ counter and returns $(X^{2n - 1 - i}, X^{2n + i}, X^{4n - 1 - i})$ wires.
+* [`gate`] advances an $i$ counter and returns $(X^{2n - 1 - i}, X^{2n + i}, X^{4n - 1 - i}, X^i)$ wires.
 * `enforce_zero` advances a counter $j$ and adds a fresh linear combination of previous wires multiplied by $Y^j$.
 
 Most of the linear constraints created within a routine consist purely of wires
@@ -160,6 +160,7 @@ out-of-line.**
 [`Output`]: ragu_core::routines::Routine::Output
 [`Routine::execute`]: ragu_core::routines::Routine::execute
 [`enforce_zero`]: ragu_core::drivers::Driver::enforce_zero
+[`gate`]: ragu_core::drivers::DriverTypes::gate
 [`mul`]: ragu_core::drivers::Driver::mul
 [`alloc`]: ragu_core::drivers::Driver::alloc
 [`Any`]: core::any::Any
--- a/book/src/implementation/routines.md
+++ b/book/src/implementation/routines.md
@@ -277,6 +277,24 @@ cached subtree cannot be reused). Consolidating both levels into a single
 identity type loses this distinction and forces the floor planner to
 treat segments as different when only their subtrees differ.
 
+The two-level design avoids four conflation risks that arise when a single
+fingerprint type serves both roles. First, [`TypeId`] pairs do not affect
+what a segment contributes to $s(X, Y)$, so including them in the
+polynomial-equivalence key would prevent the floor planner from grouping
+type-distinct routines that contribute identically. Second, tests that
+assert "different TypeIds ⇒ different fingerprints" are testing memoization
+safety, not polynomial correctness; naming them as if TypeIds were
+necessary for fingerprint correctness obscures which invariant is actually
+being verified. Third, the `Counter` and the wiring polynomial evaluators
+compute fundamentally different things — a context-independent structural
+hash versus an actual polynomial contribution — so bundling both concerns
+into one type hides the boundary between them. Fourth, without an
+independent reference evaluation to compare against, a memoization cache
+that silently returns wrong results could go undetected; the separation
+makes it possible to test polynomial equivalence (via
+[`RoutineFingerprint`]) and memoization substitutability (via
+[`MemoFingerprint`]) with distinct, targeted assertions.
+
 ### Repositioning {#repositioning}
 
 Each segment occupies a **non-overlapping** range of gates and constraints in
--- a/book/src/implementation/routines.md
+++ b/book/src/implementation/routines.md
@@ -24,13 +24,6 @@ drivers use manual `mem::replace` on the scoped state instead when they need to
 inspect the child scope's result before restoring.
 We refer our invocation-depth-independent drivers as being _context insensitive_.
 
-Drivers cannot distinguish routines themselves but merely their **invocations**;
-in each `routine` call, they learn the [`TypeId`] of their [`Input`] and
-[`Output`] gadgets, and thanks to
-[fungibility](../guide/gadgets/index.md#fungibility) this is very
-useful for their structural analysis (via [conversions]) and induces stable
-algebraic properties at routine boundaries.
-
 ## Algebraic Description {#algebraic-description}
 
 Routines exploit the fact that contiguous sections of code tend to have
@@ -98,7 +91,7 @@ Similarly, we can extract the common factor $X^{2n}Y^j$ for the starting
 constraint index $j$ and the remaining univariate $g_3$ is invariant to the
 routine invocation.
 
-## Example: Synthesis Trace
+## Example: Synthesis Trace {#example}
 
 The following circuit serves as a concrete example for all upcoming discussion:
 
@@ -167,17 +160,43 @@ Let's start with some naive but incorrect ways to identify a routine:
   even the same `ScaledTxz` routine type (or `RoutineB` in our example)
   can synthesize to different traces.
   
-In short, drivers cannot distinguish routines themselves but merely their
-**invocations**. Fingerprinting is the process by which a routine invocation is
-succinctly encapsulated so that _matching fingerprints implies opportunities for
-routine memoization_. There could be multiple fingerprints for one invocation,
-each fingerprint signals a different memoization strategy or aggressive level.
+In short, drivers cannot distinguish routines statically alone, but must also 
+consider their **invocations**. Fingerprinting is the process by which a routine
+invocation is succinctly encapsulated so that _matching fingerprints implies
+opportunities for routine memoization_. There could be multiple fingerprints for
+an invocation, each signals a different memoization strategy or aggressive level.
 Ragu produces two fingerprints: a **base fingerprint** that captures _segment-level
 equivalence_ and signals memoization of only the base segment of an invocation;
-and a **deep fingerprint** that captures _subtree-level equivalence_ and signals
-memoization of the entire routine, recursively include all nested sub-routines.
+and a **deep fingerprint** that captures _tree-level equivalence_ and signals
+memoization of the entire routine, recursively including all nested sub-routines.
 
-Conceptually, the simplest approach is to pick a random point $(x_r, y_r)$
+### Base fingerprint {#base-fingerprint}
+
+Ragu tracks the gates and constraint counts of the local/base segment of a
+routine invocation and its polynomial contribution for `BaseFingerprint`.
+In the [example above](#example), although the two invocations of `RoutineB`
+are obviously different, it's possible that the internal contribution of their
+base segments are identical thus memoizable (i.e., `b0 + b1 = b2`).
+
+Note that the polynomial contribution only include invocation's internal and
+system contributions, those invariant to the invocation's context and location,
+and exclude all interface contributions. Further note that polynomial evaluation
+alone is insufficient because it only captures `enforce_zero` calls, not gate
+count which is crucial for floor planner later. Technically, the constraint
+count, already bounded in the contribution scalar, is purely defense-in-depth.
+
+```rust
+struct BaseFingerprint {
+    /// shape of the **base segment** of a routine
+    num_gates: usize,
+    num_constraints: usize,
+    /// polynomial contribution of the base segment
+    /// in practice, we use `low_u64<F>(eval) -> u64`
+    eval: F,
+}
+```
+
+Conceptually, the simplest approach to `eval` is to pick a random $(x_r, y_r)$
 and use the internal contribution of each routine invocation to $s(x_r, y_r)$
 as its fingerprint. By Schwartz-Zippel lemma, two different invocations will
 produce different polynomial contribution with overwhelming probability.
@@ -186,9 +205,88 @@ forward direction ($X^i$ terms) and backward direction ($X^{4n-1-i}$ terms); and
 involves scalar multiplications (for $X^{2n}$ and $X^{4n}$) and expensive
 inversion (for $X^{-1}$) during initialization.
 
-In contrast, Ragu takes a more efficient approach where fingerprints 
-TODO
+Instead, Ragu takes a more efficient approach: use 4 _independent_ geometric
+sequences for wire values (one for each of `a/b/c/d`-wires) and Horner-accumulate
+contribution of `enforce_zero` constraints with $Y$ powers of a random $Y=y_r$.
+Wires from the $i$-th gate corresponds to $(x_0^i, x_1^i, x_2^i, x_3^i)$ monomial
+evaluations for a random $(x_0, x_1, x_2, x_3)$ picked at initialization.
+This simplification is possible thanks to the relaxed requirement on the
+fingerprint: characterize the constraints in any binding way, not necessarily in
+the same way as the wiring polynomial $s(X, Y)$.
 
+Handling interface contributions is the most subtle logic. In our example, the
+first invocation of `RoutineB` accepts an input gadget from the parent routine
+on enter and receives an output gadget from the child `RoutineA` in the middle.
+Both gadgets are allocated outside the scope of the base segment of `RoutineB`,
+thus any constraint on them belongs to interface contributions and should not
+meaningfully affect the `eval` computation. Concretely, relocating wires in these
+interface gadgets should not affect the `eval` fingerprint.
+
+There are two approaches to dealing with wires in these interface gadgets:
+
+1. discriminate them as a separate wire type:
+   (e.g. `enum WireEval { Value, One, Interface }`) so that routines can flag
+   them as `WireEval::Interface(F)` and drivers can drop their contribution.
+2. assign context-insensitive values to them so that their contributions are
+   cancelled out among otherwise equivalent invocations, thus cannot impact
+   equivalence meaningfully.
+
+Ragu takes the second approach, implemented via a wire remapping [`WireMap`]:
+all wires in interface gadgets are remapped to hold values from an independent
+geometric sequence with another random base.[^x-remap] Importantly, every
+routine, no matter at which invocation depth, starts with the same geometric
+sequence re-initialized so that the input gadget wires holds a positional value
+that is context-insensitive. Furthermore, before exit, every routine remap
+the output gadget and appended to its parent's context.
+
+[^x-remap]: For code reference, the variable is named `x_remap` and remapped 
+    wires take on values `x_remap^i` which is independent of normal sequences
+    $(x_0^i, x_1^i, x_2^i, x_3^i)$. Intuitively, every routine starts with a
+    blank page for interface gadget wires written sequentially from the page;
+    and their output gadget wires are appended to the parent's page.
+
+### Deep fingerprint {#deep-fingerprint}
+
+The most aggressive memoization is possible when two routine invocations are
+tree-level equivalent with the structurally identical base and subroutines.
+Intuitively, two invocations with identical deep fingerprints means that their
+shapes (gate and constraint counts) and their polynomial contributions are
+identical if we _flatten and inline all their nested subroutines_.
+
+The main gap from [base fingerprint](#base-fingerprint) is how to handle wires
+of output gadgets from the child subroutines. In base fingerprint, output wires
+are remapped in their parent's context. This remapping erases their positions
+relative to the start of the base/parent routine, which is necessary to test
+equivalence of the inlined routine.[^swapped] Therefore, we must capture the
+positional values of these output wires in the parent's deep fingerprint.
+
+```rust
+struct DeepFingerprint {
+    base: BaseFingerprint,
+    /// deep hash: 
+    /// H(self.base, output_gadget, child_deep_hashes, input_type, output_type)
+    deep: u64,
+}
+```
+
+Ragu augments the base fingerprint with the wire values of the output gadget,
+the deep fingerprint of all its child invocations (one level deeper), and the
+`TypeId` of the input and output gadget types (optional, only defense-in-depth).
+The monomial evaluation of output wires (of a subroutine) already encodes their
+positions relative to the start of the subroutine (thus transitively relative
+to the start of the parent routine). Since the deep fingerprint of the parent
+routine recursively hash in the deep hashes of all child subroutines, the final
+hash binds the entire invocation tree.
+
+
+[^swapped]: As an example, two `RoutineB` nested-call `RoutineA`. The first
+    `RoutineA` invocation: `alloc(w1); alloc(w2); enforce_zero(w1 + w2); return (w1, w2)`;
+    the second invocation: `alloc(w2); alloc(w1); enforce_zero(w1 + w2); return (w1, w2)`.
+    In `BaseFingerprint` computation, the output wires `(w1, w2)` from both
+    invocations are mapped to the same values in their parent `RoutineB` context.
+    Two `RoutineB` invocations gets identical base fingerprints.
+    However, after inlining, the two invocations differ because of swapped
+    allocation of `w1` and `w2`.
 
 ## Pipeline {#pipeline}
 
@@ -457,6 +555,7 @@ fingerprinting model at all, this comparison catches errors in the model
 itself — a routine fingerprint that over-groups, or a memo fingerprint
 that omits a necessary field — not only bugs in the memoization code.
 
+[`WireMap`]: ragu_core::convert::WireMap
 [`RoutineFingerprint`]: ragu_circuits::metrics::RoutineFingerprint
 [`MemoFingerprint`]: ragu_circuits::metrics::MemoFingerprint
 [`TypeId`]: core::any::TypeId
--- a/book/src/implementation/routines.md
+++ b/book/src/implementation/routines.md
@@ -35,7 +35,7 @@ algebraically convenient structure in the resulting wiring polynomial $s(X, Y)$:
 * [`gate`] advances an $i$ counter and returns $(X^{2n - 1 - i}, X^{2n + i}, X^{4n - 1 - i}, X^i)$ wires.
 * `enforce_zero` advances a counter $j$ and adds a fresh linear combination of previous wires multiplied by $Y^j$.
 
-Most of the linear constraints created within a routine consist purely of wires
+Most of the linear constraints created within a routine consist purely of wires also
 created within that routine, meaning that the **internal** contribution of a
 routine can be written as polynomials $g_x, g_{x^{-1}}$ and memoized if the
 routine is invoked multiple times. Given a cache hit, we need only scale $g_x$
@@ -70,11 +70,11 @@ graph LR
         FP[floor plan]
     end
     subgraph tp[trace path]
-        rx --> trace
+        TE[trace eval] --> trace
     end
     C --> MP
-    C --> rx
-    W --> rx
+    C --> TE
+    W --> TE
     metrics --> FP
     FP --> ED
     trace --> TP["r(X)"]
@@ -121,7 +121,7 @@ describes where the $i$-th segment is placed, regardless of whether a
 future floor planner reorders their positions.
 
 The [`Counter`] uses each invocation to build a [`SegmentRecord`] and a
-[`RoutineFingerprint`]. It resets its geometric sequences and Horner
+[`MemoFingerprint`]. It resets its geometric sequences and Horner
 accumulator to a fixed initial state — independent of the caller — so that
 the fingerprint captures only the routine's internal constraint structure.
 Input wires are remapped into the fresh scope without incrementing
@@ -170,7 +170,7 @@ Execution traces are divided into **segments**. All wires allocated outside of
 routine invocations belong to a single **root segment**.[^root-segment] Each
 routine invocation creates a new segment containing only the wires allocated
 directly within it; nested calls produce their own segments in turn. The
-`CircuitExt::rx` method produces a `Trace` that contains these segments in DFS
+`CircuitExt::trace` method produces a `Trace` that contains these segments in DFS
 order, but their actual arrangement in the trace polynomial depends on the floor
 plan's repositioning values.
 
@@ -228,14 +228,14 @@ per-invocation.
 ### Fingerprinting
 
 Two invocations are structurally equivalent when they share a
-[**fingerprint**][`RoutineFingerprint`]: the pair `(TypeId(Input),
-TypeId(Output))`, the Schwartz–Zippel evaluation scalar, and the local gate and
-constraint counts. The [metrics pass](#pipeline) computes a fingerprint for each
-invocation by executing its constraint logic on a lightweight [`Counter`] driver
-that substitutes independent geometric sequences for wire values and
-accumulates constraint contributions via Horner's rule. If the resulting scalar,
-type pairs, and counts all match, the two invocations are structurally
-equivalent with overwhelming probability.
+[**routine fingerprint**][`RoutineFingerprint`]: the Schwartz–Zippel evaluation
+scalar and the local gate and constraint counts. The
+[metrics pass](#pipeline) computes a fingerprint for each invocation by
+executing its constraint logic on a lightweight [`Counter`] driver that
+substitutes independent geometric sequences for wire values and accumulates
+constraint contributions via Horner's rule. If the resulting scalar and counts
+match, the two invocations are structurally equivalent with overwhelming
+probability.
 
 Fingerprints capture only the **local** constraint structure. When entering a
 routine, the `Counter` resets its geometric sequences and Horner accumulator to
@@ -243,7 +243,7 @@ a fixed initial state, and output wires from nested routine calls are remapped
 to fresh positions in the caller's sequence space. This ensures the fingerprint
 is independent of calling context and of any sub-routines the routine invokes.
 
-### Shallow vs. Deep Fingerprints
+### Routine vs. Memo Fingerprints
 
 The [`TypeId`] pairs `(Input, Output)` do not affect what a segment
 contributes to $s(X, Y)$. Two routines with different Rust types but the
@@ -252,28 +252,29 @@ gate and constraint counts — produce identical polynomial contributions.
 This can be verified directly: wrap each routine in a circuit and assert
 that their $s(x, y)$ evaluations agree at random points.
 
-For floor planning, only the polynomial contribution matters. A **shallow
-fingerprint** `(eval, num_gates, num_constraints)` suffices to group
-segments with the same polynomial shape; including [`TypeId`] pairs would
-prevent the floor planner from grouping type-distinct routines that
-contribute identically. For memoization, the constraints are stricter: a
-cached routine must be safely substitutable for a fresh execution, which
-requires matching output wire mappings and recursive subtree structure. A
-**deep fingerprint** extends the shallow fingerprint with `output_eval`
-(a scalar encoding the output wire mapping) and a recursive hash that
-folds in [`TypeId`] pairs, all shallow fields, the child count, and each
-child's deep hash. The floor planner keys on the shallow fingerprint; the
-memo cache keys on the deep fingerprint.
+For floor planning, only the polynomial contribution matters. A
+[`RoutineFingerprint`] `(eval, local_num_gates, local_num_constraints)`
+suffices to group segments with the same polynomial shape; [`TypeId`]
+pairs are deliberately excluded so that the floor planner can group
+type-distinct routines that contribute identically. For memoization, the
+constraints are stricter: a cached routine must be safely substitutable
+for a fresh execution, which requires matching output wire mappings and
+recursive subtree structure. A [`MemoFingerprint`] extends the routine
+fingerprint with `output_eval` (a scalar encoding the output wire
+mapping) and a recursive `deep` hash that folds in [`TypeId`] pairs, all
+routine fingerprint fields, the child count, and each child's deep hash.
+The floor planner keys on the routine fingerprint; the memo cache keys on
+the memo fingerprint.
 
-The distinction is semantic, not just organizational. A shallow
+The distinction is semantic, not just organizational. A routine
 fingerprint captures **segment-level equivalence**: whether two individual
 segments impose the same local constraints and could be aligned in the
 polynomial layout regardless of their subtrees — for instance, padding
-one circuit's segment against another circuit's offset. A deep
+one circuit's segment against another circuit's offset. A memo
 fingerprint captures **subtree-level equivalence**: whether entire routine
 invocations are interchangeable and can be substituted via memoization.
-A segment can be shallow-equivalent to another (same local constraints,
-worth aligning) but deep-inequivalent (different children, so the
+A segment can be routine-equivalent to another (same local constraints,
+worth aligning) but memo-inequivalent (different children, so the
 cached subtree cannot be reused). Consolidating both levels into a single
 identity type loses this distinction and forces the floor planner to
 treat segments as different when only their subtrees differ.
@@ -318,10 +319,11 @@ in a single check. The non-overlapping invariant
 ensures that any disagreement is a real bug rather than a masking
 coincidence. Because the native path does not depend on the
 fingerprinting model at all, this comparison catches errors in the model
-itself — a shallow fingerprint that over-groups, or a deep fingerprint
+itself — a routine fingerprint that over-groups, or a memo fingerprint
 that omits a necessary field — not only bugs in the memoization code.
 
 [`RoutineFingerprint`]: ragu_circuits::metrics::RoutineFingerprint
+[`MemoFingerprint`]: ragu_circuits::metrics::MemoFingerprint
 [`Counter`]: ragu_circuits::metrics::Counter
 [`Trace::assemble`]: ragu_circuits::trace::Trace::assemble
 [`TypeId`]: core::any::TypeId
--- a/book/src/implementation/routines.md
+++ b/book/src/implementation/routines.md
@@ -24,14 +24,6 @@ drivers use manual `mem::replace` on the scoped state instead when they need to
 inspect the child scope's result before restoring.
 We refer our invocation-depth-independent drivers as being _context insensitive_.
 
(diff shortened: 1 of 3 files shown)


book/src/implementation/routines.md
```diff
@@ -0,0 +1,497 @@
...
+Routines can further invoke other routines, creating **nested routines**.
+To avoid complicated call-stack management, drivers keep only _routine-local_
+state. When circuit code calls [`Driver::routine`], the driver saves its
+scoped state (e.g., running wire monomial evaluations for `sx::Evaluator`,
+constraint counters for `metrics::Counter`), and initializes a fresh scope
+for the child routine. On return, the parent scope is restored and the child's
+contribution is folded into the parent's accumulated result. The `DriverScope`
+trait provides a `with_scope` helper that automates this save-and-restore; some
+drivers use manual `mem::replace` instead when they need to inspect the child
+scope before restoring the parent.
```
TalDerei: nit; i'd be more explicit on the actual isolation boundaries and behavior (partly motivated by the vague "folded into the parent's accumulated result" phrasing)

```suggestion
Routines can further invoke other routines, creating **nested routines**.
To avoid complicated call-stack management, drivers keep only _routine-local_
state. When circuit code calls [`Driver::routine`], the driver saves its
scoped state (e.g., running wire monomial evaluations for `sx::Evaluator`,
constraint counters for `metrics::Counter`), and initializes a fresh scope
for the child routine. The parent and child scopes are isolated from each
other: a child scope runs against _no_ inherited parent state, so its
execution depends only on its body and inputs. On return, the parent scope
is restored unchanged, and the child's contribution is folded into
driver-level output rather than the parent's scope. The `DriverScope` trait
provides a `with_scope` helper that automates this save-and-restore; some
drivers use manual `mem::replace` instead when they need to inspect the
child scope before restoring the parent.

File: book/src/implementation/routines.md [hunk 2]
```diff
@@ -0,0 +1,497 @@
...
+
+Because driver behavior does not depend on invocation depth, we call them
+_context insensitive_.
```
TalDerei: nit; "context-insensitive" drivers feels overloaded ("context" shifts meaning across this entry, and strictly the driver is context-aware since it manages scope state, floor plan lookup). think we should tighten this up as a property/invariant that applies to each scope, not the driver itself. or maybe call drivers "depth-insensitive," since they don't branch on invocation depth?

book/src/implementation/routines.md
```diff
@@ -0,0 +1,497 @@
...
+internal contribution, an **interface contribution** from constraints that
+reference wires created outside the routine (e.g., input gadget wires), and a
+**system contribution** from constraints that reference system wires (i.e.,
+`ONE`) at fixed locations.
+
+The interface contribution cannot be factored or memoized the same way:
+the input gadget is allocated by the caller at an unknown position, both
+absolute and relative to the start of the routine, so there are no a priori
+$X^i$ factors to extract. Interface contributions must be computed
+per-invocation.
```
TalDerei: nit ordering; we should discuss the interface contribution first, before the overall contribution in the previous section. trying to collapse Sean's version and yours:

```suggestion
The interface contribution to $s(X, Y)$ involves wires created outside                                                                                                                                                                           
the routine, such as the input gadget wires and `ONE` wire. Unlike the internal 
and system contributions, it cannot be factored or memoized the same way: these wires                                                                                                                                                                       
are allocated by the caller at positions unknown to the routine, both                                        
absolute and relative to its start, so there are no a priori $X^i$ factors                                                                                                                                                                       
to extract. We write the interface contribution as $\sum_\ell h_\ell(X, Y)$,                                                                                                                                                                     
with one $h_\ell$ per interface wire, and compute it per-invocation.

File: book/src/implementation/routines.md [hunk 4]
```diff
@@ -0,0 +1,497 @@
...
+    multiplied by $Y^{j+\ell}$ for some $\ell \geq 0$.
+    The internal contribution — the sum of all $Y$-power-scaled linear
+    combinations of $X$ monomials — therefore shares a common factor $X^i Y^j$
+    (or $X^{-i}Y^j$). Extracting these factors yields $g_1$ and $g_2$.
+
+The overall contribution of a routine invocation to $s(X, Y)$ is the sum of this
+internal contribution, an **interface contribution** from constraints that
+reference wires created outside the routine (e.g., input gadget wires), and a
+**system contribution** from constraints that reference system wires (i.e.,
+`ONE`) at fixed locations.
```
TalDerei: referencing https://github.com/tachyon-zcash/ragu/pull/599#discussion_r2969868717, I think we avoid granularity at the system contribution level? also, it'd make sense to restore the equation showing the contribution of a routine invocation to s(X, Y), which was removed.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,497 @@
...
+requires that output gadgets of child routines are also treated as interface
+contributions of the parent routine.
+It is always safe to classify them this way and recompute on every invocation
+of the parent.
+However, if two parent routines are known to have identical sub-routines, the
+output gadgets can be remapped to matching positions in the parent's context,
+allowing further constraints on them to be memoized as internal contributions.
+We elaborate on this [below](#memoization).
+
+System contributions are memoizable by similar reasoning:
```
TalDerei: I'd honestly remove this section and just black box it, lumping it with the interface contribution. very little memoization benefit afaict, and probably not worth the mental complexity.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,497 @@
...
+
+A complication arises from nested routines. The context-insensitive design
+requires that output gadgets of child routines are also treated as interface
+contributions of the parent routine.
+It is always safe to classify them this way and recompute on every invocation
+of the parent.
+However, if two parent routines are known to have identical sub-routines, the
+output gadgets can be remapped to matching positions in the parent's context,
+allowing further constraints on them to be memoized as internal contributions.
+We elaborate on this [below](#memoization).
```
TalDerei: suggestion to consider; we discussed OOB the different partial memoization axes: (1) the **input wire axis** (G/H poly split) which is described in the previous paragraph, and (2) the **nesting axis** (base / nested subtree split) which is expanded in more detail later. I think we should frame this subsection through those two lenses, since they're the two primary memoization axes we're operating under _algebraically_. maybe these should be subsections under the algebraic section?

```suggestion
Nested routines engage both memoization axes: the input-wire axis                                                                                                                                                                                
(internal $g_1, g_2$ vs. interface $h$) and the nesting axis
(base segment vs. full nested subtree).                                                                                                                                                                                                          
                                                                                                              
Along the **input-wire axis**, output gadgets of child routines are external to                                                                                                                                                                      
the parent — allocated at the child's offset, not the parent's — so
constraints referencing them fall into the interface bucket $h$. This                                                                                                                                                                            
classification is always safe and requires per-invocation computaton.    
                   
Along the **nesting axis**, this classification can be tightened. When two                                                                                                                                                                           
parent invocations share the full nested subtree — not just the base                                         
segment — child outputs sit at matching positions relative to the parent's                                                                                                                                                                       
start and can be promoted from interface to internal. We elaborate on
when this promotion is permitted #memoization.

File: book/src/implementation/routines.md [hunk 7]
```diff
@@ -0,0 +1,497 @@
...
+## Segments {#segments}
+
+Given an execution trace, we need a global indexing of all sub-sections so
+that different drivers can refer to them consistently. Ragu divides the trace
+into **segments** and orders them in a canonical **DFS order** determined by
+routine invocations. In particular, circuit sections at the same invocation
+depth are merged into a single segment; wires allocated outside of any routine
+belong to a special **root segment**.[^root-segment] Each routine invocation
+creates a new **base segment** containing only the wires allocated directly
+within it; nested calls produce their own segments in turn. The DFS-ordered
```
TalDerei:
```suggestion
creates a new **base segment** containing only the wires allocated directly
within in its own body, excluding anything its nested sub-routines allocate; 
nested calls produce their own segments in turn. The DFS-ordered

File: book/src/implementation/routines.md [hunk 8]
```diff
@@ -0,0 +1,497 @@
...
+Drivers cannot distinguish routines statically but must also consider
+their **invocations**. Fingerprinting is the process by which a routine
+invocation is succinctly encapsulated so that _matching fingerprints imply
+opportunities for routine memoization_. An invocation can carry multiple
+fingerprints, each indicating a different memoization strategy or level of
+aggressiveness. Ragu produces two: a **base fingerprint** that captures
+_segment-level equivalence_ and enables memoization of only the base segment,
+and a **deep fingerprint** that captures _tree-level equivalence_ and enables
+memoization of the entire routine, recursively including all nested
+sub-routines.
```
TalDerei: the distinction between "segment-level" versus "tree-level equivalence" is more nuanced than that imo. can we flush out some details lifted from https://github.com/tachyon-zcash/ragu/pull/557#discussion_r2942082544?

TalDerei (commented):
some first-pass observations — haven't reviewed the memoization section and below yet.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,497 @@
...
+ │   └─ Routine C
+ │       └─ c0
+ └─ r2
+```
+
+Each label `r0, a0, b0, …, r2` denotes a contiguous section of circuit trace
+allocated via [`gate`]. At various points the circuit logic invokes a routine,
+which may in turn invoke nested routines.
+
+## Segments {#segments}
```
TalDerei, line 88: prior "segments" section included more details we can probably fold in https://github.com/tachyon-zcash/ragu/pull/599/changes#diff-ef0bcda416ffd5f82de7e097ed971b636a6889d5fb519e1635e7341e05aacccaR165-R195.

TalDerei:
opened some suggestions addressing my comments in https://github.com/tachyon-zcash/ragu/pull/675 rather than pushing to this branch, since it warrants discussion.

book/src/implementation/routines.md
TalDerei: another relevant follow-up in https://github.com/tachyon-zcash/ragu/pull/675#discussion_r3126779138.

book/src/implementation/routines.md
alxiong: sounds good! "context" is indeed a bit too broadly implicating here. (fyi, i originally call it by borrowing the term from program analysis "[context-sensitive procedural analysis](https://www.cs.cmu.edu/~clegoues/courses/15-819O-16sp/notes/notes06-interprocedural.pdf)")
Agree "depth-insensitive" is more appropriate here.

book/src/implementation/routines.md
alxiong: right, originally we have the overall contribution, but i find it more confusing to list out the interface contribution. the $h_\ell$ is under-defined (unexpanded and vague) there. More importantly, we are recomputing it per-invocation anyway, no point to squeeze in the expression that matters here, which is the form we want to demonstrate repositioning effect on.

I still vote for dropping equations involving interface contribution (transitively the overall contribution).

book/src/implementation/routines.md
alxiong: tend to agree. when i first rewrite, i want to make sure i didn't miss anything. 
But a single monomial is indeed not worth the extra complexity. but i would make a footnote about it (in case our SYSTEM gate becomes **gates** and more to memoize there. 

(altho I still wonder if there's routine that heavily refer to `ONE`, such as `Boolean` related routines, but maybe not now)

alxiong:
I forgot to flag something. During rewrite, I wonder "how much interface contribution does input gadget to a routine make"? or in other words, how much work do we have to recompute since we cannot memoize them. 

then I asked Claude to count the constraints, 
- `poseidon::Permutation`: **15/576** (only 15 constraints that involves any one of the input wire in linear constraints, because we quickly operate on input wires and create new wires which are part of internal contribution again)
- `txz::Evaluate`: **7/152**

Basically, a very small fraction, so we are not missing out much with our internal-contribution-only memoization.

An important note is that almost half of all constraints in `Permutation::execute` touches `ONE` wire: https://github.com/tachyon-zcash/ragu/pull/669#discussion_r3128174550

book/src/implementation/routines.md
alxiong: i'm hesitate with this because I personally don't think of "wire allocation" as "a memoization axis". I only treat them as definition/scope of internal contribution v.s. interface contribution. Just by algebraic properties, only the former is invariant to routine invocation, so the latter is _never_ memoizable to begin with. When I hear "optimization axis" i think of them as different direction/strategies of memoization opportunities. But in our case, the only axis is base v.s. deep/nested. 

I'm open to second opinion by @ebfull, but keeping it as-is for now.

book/src/implementation/routines.md
alxiong: I change my mind. I vote for **memoizing `ONE` contribution (don't count them as interface)**.

I asked Claude to write a test to calculate, here's the breakdown:

```
Counting only enforce_zero constraints now (288 mul gate checks no longer tallied):

  Permutation::execute() enforce_zero breakdown (Pasta Poseidon)
  - Total enforce_zero: 576 (all from enforce_equal in Element::mul)
  - Involves input wire: 15 (2.60%)
  - Involves ONE wire: 288 (50.00%)
  - Only internal wires: 288 (50.00%)
  - Overlap (input ∧ ONE): 15 — so sum of the three categories exceeds total by 15, as you predicted.

  The 50/50 ONE vs internal split comes from each sbox's third mul: x⁴ · x = x⁵ produces enforce_equal(a, x⁴) (internal-only) and enforce_equal(b, x) (touches ONE via the
  round-constant path). The first two squarings contribute 4 internal-only constraints per sbox, and 96 sboxes × (2 ONE + 4 internal) = 192 + 384, plus the first-squaring pair
  (2 ONE per sbox) = 96·2 = 192 more ONE constraints → 288 each.

The 50/50 split is because within each sbox (3 muls → 6 enforce_equals), half the constraints link a fresh wire to x (which carries ONE-provenance from
  the round-constant step), and the other half link fresh wires to x² or x⁴ (which are c outputs of a previous mul — pure internal).

  Tracing one sbox x.square().square().mul(x):

  ┌───────┬──────────┬──────────────────────┬──────────────────────┐
  │  mul  │ produces │ enforce_equal(a, …)  │ enforce_equal(b, …)  │
  ├───────┼──────────┼──────────────────────┼──────────────────────┤
  │ x·x   │ x²       │ a ≡ x (ONE+internal) │ b ≡ x (ONE+internal) │
  ├───────┼──────────┼──────────────────────┼──────────────────────┤
  │ x²·x² │ x⁴       │ a ≡ x² (internal)    │ b ≡ x² (internal)    │
  ├───────┼──────────┼──────────────────────┼──────────────────────┤
  │ x⁴·x  │ x⁵       │ a ≡ x⁴ (internal)    │ b ≡ x (ONE+internal) │
  └───────┴──────────┴──────────────────────┴──────────────────────┘

  That's 3 ONE-touching + 3 internal-only per sbox, and 96 sboxes × 3 each = 288 / 288.
```

Almost half of all constraints involve `ONE` wires, missing out on them is not ideal. 
(still doable though, afterall when a driver encounter a constraint that involve both internal wires and `ONE`, it can memoize the contribution from internal wires only, and dropping the term on `ONE`)

The main problem/cost is that: during recomputation (if we categorize it as "interface contribution"), then we have to re-compute 288 `Y`-powers, and not all of them are consecutive, there are small jumps. 

cc @ebfull 


<details>
<summary>If you want to reproduce, here's the test</summary>

```rust
    /// Throwaway test: measure what fraction of `Permutation::execute()`
    /// constraints involve the ONE wire, involve any input-gadget wire, or
    /// live purely among internally-allocated wires. Uses a custom driver
    /// whose `Wire` type is a provenance bitset that unions across linear
    /// combinations.
    #[test]
    fn test_permutation_constraint_breakdown() -> Result<()> {
        extern crate std;
        use alloc::vec::Vec;
        use core::marker::PhantomData;

        use ragu_arithmetic::PoseidonPermutation;
        use ragu_core::{
            Result as CoreResult,
            drivers::{Driver, DriverTypes, LinearExpression},
            maybe::Always,
        };

        // Provenance tag: which "kinds" of wires fed into this wire via
        // linear combinations (or direct allocation).
        #[derive(Clone, Copy, Default)]
        struct Tag {
            uses_one: bool,
            uses_input: bool,
            uses_internal: bool,
        }

        // Linear expression that only unions provenance tags across terms.
        #[derive(Default)]
        struct Lc {
            tag: Tag,
        }

        impl<F: Field> LinearExpression<Tag, F> for Lc {
            fn add_term(mut self, wire: &Tag, _: Coeff<F>) -> Self {
                self.tag.uses_one |= wire.uses_one;
                self.tag.uses_input |= wire.uses_input;
                self.tag.uses_internal |= wire.uses_internal;
                self
            }
            fn gain(self, _: Coeff<F>) -> Self {
                self
            }
        }

        struct Counter<F: Field> {
            total: usize,
            with_input: usize,
            with_one: usize,
            internal_only: usize,
            _marker: PhantomData<F>,
        }

        impl<F: Field> Counter<F> {
            fn new() -> Self {
                Self {
                    total: 0,
                    with_input: 0,
                    with_one: 0,
                    internal_only: 0,
                    _marker: PhantomData,
                }
            }

            fn record(&mut self, tag: Tag) {
                self.total += 1;
                if tag.uses_input {
                    self.with_input += 1;
                }
                if tag.uses_one {
                    self.with_one += 1;
                }
                if !tag.uses_input && !tag.uses_one {
                    self.internal_only += 1;
                }
            }
        }

        const INTERNAL: Tag = Tag {
            uses_one: false,
            uses_input: false,
            uses_internal: true,
        };

        impl<F: Field> DriverTypes for Counter<F> {
            type ImplField = F;
            type ImplWire = Tag;
            type MaybeKind = Always<()>;
            type LCadd = Lc;
            type LCenforce = Lc;
            type Extra = ();

            fn gate(
                &mut self,
                _: impl Fn() -> CoreResult<(Coeff<F>, Coeff<F>, Coeff<F>)>,
            ) -> CoreResult<(Tag, Tag, Tag, ())> {
                Ok((INTERNAL, INTERNAL, INTERNAL, ()))
            }

            fn assign_extra(
                &mut self,
                _: (),
                _: impl Fn() -> CoreResult<Coeff<F>>,
            ) -> CoreResult<Tag> {
                Ok(INTERNAL)
            }
        }

        impl<'dr, F: Field> Driver<'dr> for Counter<F> {
            type F = F;
            type Wire = Tag;
            const ONE: Self::Wire = Tag {
                uses_one: true,
                uses_input: false,
                uses_internal: false,
            };

            fn add(&mut self, lc: impl Fn(Self::LCadd) -> Self::LCadd) -> Self::Wire {
                lc(Lc::default()).tag
            }

            fn enforce_zero(
                &mut self,
            ) -> CoreResult<()> {
                self.record(lc(Lc::default()).tag);
                Ok(())
            }
        }

        // Build a SpongeState whose every Element carries an input-tagged
        // wire, so any constraint that mentions one of these wires (directly
        // or through a linear-combination-derived virtual wire) gets counted
        // in `with_input`.
        type P = <Pasta as Cycle>::CircuitPoseidon;
        let input_tag = Tag {
            uses_one: false,
            uses_input: true,
            uses_internal: false,
        };
        let elements: Vec<Element<'_, Counter<Fp>>> = (0..<P as PoseidonPermutation<Fp>>::T)
            .map(|_| Element::promote(input_tag, <Counter<Fp> as Driver>::just(|| Fp::ZERO)))
            .collect();
        let state =
            SpongeState::<'_, Counter<Fp>, P>::from_elements(elements.try_into().expect("T elems"));

        let mut dr = Counter::<Fp>::new();
        let params = Pasta::baked();
        dr.routine(Permutation::from(Pasta::circuit_poseidon(params)), state)?;

        let total = dr.total as f64;
        let pct = |n: usize| 100.0 * n as f64 / total;
        std::println!("\nPermutation::execute constraint breakdown:");
        std::println!("  total                : {}", dr.total);
        std::println!(
            "  involves input wire  : {:>4} ({:.2}%)",
            dr.with_input,
            pct(dr.with_input)
        );
        std::println!(
            "  involves ONE wire    : {:>4} ({:.2}%)",
            dr.with_one,
            pct(dr.with_one)
        );
        std::println!(
            "  only internal wires  : {:>4} ({:.2}%)",
            dr.internal_only,
            pct(dr.internal_only)
        );
        std::println!(
            "  sum of categories    : {} (overlap = {})",
            dr.with_input + dr.with_one + dr.internal_only,
            (dr.with_input + dr.with_one + dr.internal_only) - dr.total
        );

        Ok(())
    }
```
</details>

book/src/implementation/routines.md
alxiong: i have read them and thought i cover what's important. 
Do you have anything in particular that think I missed?

book/src/implementation/routines.md
alxiong: sure, i will adopt this: https://github.com/tachyon-zcash/ragu/pull/675/changes/e9a694321d5d9b51502f007090809d6f1bd94880#diff-ef0bcda416ffd5f82de7e097ed971b636a6889d5fb519e1635e7341e05aacccaR210-R229

alxiong:
here's my summary when comparing with @ebfull 's [local writeup on routines](https://github.com/ebfull/ragu/commit/5493b0720f71deb0f6af54243feab3dfa11fc960):

### Algebraic Description and Shifting

At its core, our algebraic decomposition of the polynomial contribution of routine invocation is exactly the same as Sean's. But the way Sean presents it lead to claims about memoization opportunities that conflict with ours. 

Repeating the internal contribution expression in my writeup: 

$$
Y^\ell \Big( X^m \cdot g_1(X, Y) + X^{-m} \cdot g_2(X^{-1}, Y) \Big)
$$

and the system contribution as $Y^\ell \cdot g_3(Y)$.
In Sean's note, $g_1$ is $B(y)$, $g_2$ is $A(y)$, and $g_3$ is $C(y)$. The subtle difference is that Sean internalized it as:

>  Each of $A(y)$, $B(y)$, $C(y)$ is a polynomial in $y$ at the fixed evaluation point $x$, with no dependence on the gate offset — the $m$-dependence is isolated into the $x^{\pm m}$ factors.

In other words, $A$ is viewed as $A_x$ restricted at a particular $x$, whereas my expression remains general and in bivariate form $g_1(X, Y)$ which is the quotient after extracting out the common factor $X^m Y^\ell$. In theory, both description is equivalent. But in mental model, this leads to Sean's following statement:

> This asymmetry — gate offset baked in, constraint offset deferred — is why the scalar path can share work across placements differing only in ℓ , while different gate offsets produce different Horner results even for the same routine type.

and eventually a wrong verdict: (imo, it's wrong, please challenge if I'm mistaken)

> The gate-offset asymmetry extends to a per-driver memoization split. The `sx` driver writes coefficients at absolute polynomial indices tied to the gate offset; the `sy` driver initializes its $Y$-power from the absolute constraint offset. Neither produces a position-independent result, **so neither can memoize across placements**. Only the `sxy` driver — whose local Horner accumulator depends on the routine type and input wire evaluations alone — produces a position-independent result that can be cached and repositioned with a single $y^\ell$ multiplier. 

I internalize Sean's verdict as follows: if we think " $x$ is fixed first, which determines $A, B, C$ polys, then we evaluate at $A/B/C(y)$", then for `sx` driver, when it encounter different invocations (even with matching fingerprints), the $A_x$ will be different because they bake in the absolute wire position (? I could put words in his mouth here). Therefore, memoization is not possible for `sx` driver. 

Anyhow, my claim here is the bivariate $g_1(X, Y)$ is invariant and can be memoized for all of `sx/sy/sxy` drivers. The first two drivers memoize a univariate polys that later gets scaled ($O(|routine|)$ amount of shifting) whereas the `sxy` driver only need to scale a single cached scalar ($O(1)$ shifting).

### Child Routine Output

An important agreement that both of our writeup highlight is: 

we should classify output of child routines to be interface when only shallow fingerprint matches (or during "horizontal alignment");
and memoize it as internal contribution when deep fingerprint matches (during "vertical memoization").

### Changes/Languages that I cherrypick from Sean

- explicit about 4 classes of wires: internal, input, child output, ONE
- on hermetic routine: internal wires won't be referenced from outside and interface only contains input and ONE
- the $(m, \ell)$ symbol for starting index/position of a segment 
- overall narrative order. 
- dedicated "shifting/repositioning" section and give explicit shifting algebra

### Bugs/Mistakes I corrected thanks to Sean

- Fingerprint computation needs 1-based rather than 0-based, so that the first wire in each routine is not contributing to $x_{0/1/2/3}^0$, but at least $x_{0/1/2/3}^1$. this is reflected in my `[^onebase]` footnote
- need to hash child output gadget length into deep fingerprint. We already did this in implementation, but I forgot to write in doc
- independent challenge $x_{one}$ for `ONE` wire, so that `ONE` wire's contribution won't be confused with the first `d`-wire local to any routine. since they are all $X^0$ after factoring out the $X^m$

### Points that I favor mine over Sean

- explicit struct `BaseFingerprint` and `DeepFingerprint` and design rationale for them
  - naive 3 ways to identify a routine as strawman
  - 2 approaches to handle interface wires (discriminate wire type or context-insensitive input wire monomial)
- Dropping the explicit interface polynomial (same reply here: https://github.com/tachyon-zcash/ragu/pull/669#discussion_r3127830224) since it's less relevant, not used in memoization or fingerprinting anyway. Just mention they are one of the components and need to recomputed per-invocation
- **both shallow and deep fingerprint matching implies memoization opportunity**, Sean's language sounds like shallow only do alignment, deep implies subtree memoization. 
  - Sean used "vertical memoization v.s. horizontal alignment" which sounds contrastive, but is misleading imho because we also memoize when only shallow fingerprint matches. 
- Base/Deep naming over Shallow/Deep. I find "base" to be explicit and semantically meaningful.
- Our WireMap explanation seems more readable

ebfull:
The writeup explores how fingerprinting identifies routine invocations that contribute identically to $s(X, Y)$, and correctly argues that input/output types are irrelevant for that decision, calling them a "defense-in-depth" measure at best. But memoization during evaluation has a second job the writeup leaves unexamined: every routine still gets invoked at least once, and on subsequent invocations the driver must hand the caller a typed output gadget without re-running the body.

The plausible mechanism is to obtain an empty output gadget and inject the predicted output wire values into it based on the new placement, using the first invocation's results as a reference. The fact that two invocations contribute identically to $s(X, Y)$ does not by itself justify reusing one's predicted output as the other's, since they may expect output gadgets of different shapes or types.

Whether such an injection is safe depends on further analysis -- perhaps it's fine, but there's no separate discussion of this aspect of it, and it seems to rest on different properties than what the fingerprinting currently explores. I would prefer if they could be unified, even at some expressive cost, since simpler is better, which may justify incorporating the types. (I also feel like the "defense-in-depth" nature of including the types anywhere might mask these kinds of things and prevent us from systematically analyzing it from first principles.)

I'm also trying to understand the role the gadget types play. There was a deliberate decision to make `GadgetKind: Any + 'static` that relied on the ability to distinguish the types being important for soundness. If this ends up not being important (or only useful as a speculative "just in case") that justification then dissolves, and a lot of other things might as well.

alxiong:
This is a quite incisive point. Some OOB discussion with @ebfull also enlighten me about the original motivation:

> the [fungibility](https://tachyon.z.cash/ragu/guide/gadgets/index.html#fungibility) guarantee—that a gadget's synthesis behavior is determined entirely by its type—is what gives the driver an identity system for routines in the first place.
>
> Gadgets must be _fungible_: a gadget's behavior during circuit synthesis must be fully determined by its type, not by any particular instance's state.

iiuc, gadgets are required to be fungible because we hope to use them to deduce routine identity. 

Fungibility (in the broadest sense) requires that type (`GadgetKind`) determines:
1. wire layout: wire counts, allocation pattern
2. constraint semantic: constraints over allocated wires within the gadget

In plain, when I see `Boolean`, I know how many wires inside and how they are allocated, as well as the guaranteed that they are constrained to be `0/1`, regardless of the driver context, or witness value of that particular gadget instance.

#### Detour: comments on status quo of fungible gadget

In practice, lots of these are API contract, not guaranteed by type system. The avoidance of `enum` or dynamic array only ensures "type -> wire count", but doesn't enforce anything about how those wires are allocated (e.g. in what order or using what allocator). Nor does the type system enforce that `GadgetKind` determines the same constraints are applied to those wires. 
Empirically, our existing gadgets uphold _layout determinism_ and _constraint semantic equivalence_ via a centralized, single `::alloc()` method. We can only allocate `Boolean` in one way which is `Boolean::alloc()` which ensures a consistent allocation pattern as well as consistent 0/1 constraint on them. But again, this is not statistically enforceable. We expect all gadget implementors to respect the contract.

(A tangential comment is: some current rustdoc _overstate_ the guarantee provided by the proc-macro and type system. comments like "The derive macro enforces this (fungibility)" is inaccurate since the macro at most rejects some basic foot-gun and obvious violation of the layout determinism, but never enforce "types determine allocation pattern and constraint semantic")

#### Back from detour

The original thesis is "fungible gadgets help _statically_ identify routine". 
The original vision, as I hypothesized retroactively, is that `(input_type, output_type, routine_type)` determines the routine contribution. 
But as we analyzed and wrote in the book, `TypeId`s are insufficient for routine id -- both `Square` and `Double` routine have the same I/O gadget type, but clearly different routines and $s(X,Y)$ contribution; configurable routines `DoubleOrSquare { flag: bool }` would behave differently given a runtime state `flag`, thus cannot statically determine a routine invocation.
Our analysis has clearly show that a run-time fingerprinting is necessary to characterize an invocation.

(side note: it's historical incidence that `trait Routine` is never `Any` while assumed to be parameterizable from the start, which should imply this original thesis should be ruled out much earlier (?). My only next guess is that for simple configuration, via const generics, the type triple _might_ still be used as identifier, like `DoubleOrSquare<const flag: bool>` or `ChainedPermute<Input = FixedVec<Element, 5>, Output = FixedVec<Element, 5>, const times: usize>`. But clearly, this doesn't generalize to complex stateful/configurable routines.
Safely ignore this minor comment, it's just me thinking out loud)

So now we focus on our analysis on deep fingerprinting. I currently claimed that: 

> deep_hash = 
> H(base_fingerprint, output_gadget, output_gadget.len(), child_deep_hashes, input_type, output_type)
> ... 
> and the TypeId of the input and output gadget types (**optional, defense-in-depth only**).

Is it sound to remove `input_type, output_type` from deep fingerprint? 
Yes! because from polynomial contribution perspective, these contribution is type-oblivious. matching deep fingerprints implies we can memoize and later shift the cached result for the next invocation. This is true mathematically. 

Then Sean asked (paraphrasing here):
if gadget types doesn't help with routine fingerprinting at all, can we **relax the original fungibility requirement** on the `GadgetKind`? such as dropping `Any` (since no more `TypeId::of::<G>()`) or even allowing dynamic array (since gadgets with different length will be captured by the fingerprint scalar, thus won't affect soundness).

### Relaxing layout determinism to allocation determinism

In response, I realize that I was only focusing on the math invariants for memoization, but forget the actual memoization mechanism. Concretely, 
For memoization to work, we first need routine identification during metric pass; then during actual cache hit, we shift memoized contribution and skip actual wire-walking by the expensive evaluator driver (such as `sx/sy/sxy::Evaluator`) and replace with native-cost wireless `Emulator` to execute and map its output gadget back to the evaluator driver.
 
Safe _output gadget reconstruction_ only needs **a relaxed version of the _wire-layout determinism_ property** from the fungibility guarantee -- let's call it **allocation determinism** (dropping the wire-count determinism).

Concretely, allocation determinism asks that given a `GadgetKind` and the wire stream produced by `Gadget::map()` at the routine boundary, there is a deterministic, context-insensitive algorithm that reconstructs the gadget instance — with the stream's length as the only dynamic input, and no out-of-band metadata.

Under this property, `MyCollectionGadget { wires: Vec<Element> }` is fine: the routine boundary streams `N` wires across, and the reconstruction algorithm wraps each into an `Element` and packs them into a Vec of length `N.` Different `N` produces different fingerprints, so memoization soundness is preserved by the fingerprint, not by type-determined wire counts. 

Enums fail this property: variant choice cannot in general be recovered from the wire stream alone (two variants may have equal wire counts, or the variant set may evolve such that they collide). Permitting enums would require a separate discriminant channel, which the property explicitly disallows.

### Other comments

An orthogonal, related question is: can we relax the "gadget type determines constraint semantic" requirement? 
Technically, this relaxation won't affect routine memoization, but practically it makes circuit code error prone and violate the expectation of dev when they see gadgets of the same type. 
Imagine a world where `Boolean` gadget has both `::alloc()` and `::alloc_unconstrained()` or `::alloc_partial_constrained()`. Then it's confusing to any boolean gadget holder without the constraint contract. 

Finally, given that we _could_ remove `TypeId` from deep hash, _should we_?
I argue that we should! Consider two routine invocations with the same subtree and deep structural equivalence. Both routine returns a single wire, the first routine returns an `Element` gadget, the second returns a `MyElement` gadget. 
Clearly, we should allow memoization, but if hash in `TypeId` then they will be categorized as diverging deep fingerprints. 

## Conclusion

- we could relax fungibility requirement on gadget: wire-layout determinism relaxed to _allocation determinism_ (allowing dynamic sized array, gadget types don't have to determine wire count); and constraint semantic equivalence still required to avoid surprises for circuit dev.
  - should also update the rustdoc about the sub-contract within the fungibility property, clarify what is provided by the proc-macro, and what needs to upheld as API contract by the gadget implementor
- we should remove `TypeId` from deep hash computation
- we could (?) remove `Any + 'static` on `GadgetKind` since there's no need to know their type id anywhere in our code now

TalDerei:
> This is a quite incisive point. Some OOB discussion with @ebfull also enlighten me about the original motivation:

there’s a few discussion threads here; i’ve narrowed focus on what i think the crux is: (1) whether fingerprinting type erasure (pruning `TypeId`s) carries safe output wire injection. we know that at the `BaseFingerprint` level, output wires aren't bound (they don’t capture output wires, so routine invocations with the same s(X, Y) contributions are treated as equivalent even if their output wires differ). that seems fine because it's currently for floor-planner alignment; output wire injection happens at the deep level, where output wires are bound (they're hashed into `DeepFingerprint`). so the question i’m going back to: are two routines with the same s(X, Y) contribution but different output gadget shapes / types equivalent under our model, assuming we proceed with this type erasure? possibly relaxing fungibility requirements / removing trait bounds is downstream of this. my _intuition_ suggests that removing these `TypeId`s is safe under a possibly relaxed (or even unchanged) fungibility contract, but that needs to be verified.

another framing: from first principles, what’s the minimum property / invariant a fingerprint must capture for safe output-wire injection? and then work backwards from there.

(2) sub-question: can we remove extra fields in the deep hash, like output wire length, or is it actually load bearing and prevents vectors we’ve discussed OOB like length-extension collisions?

(3) sub-question: if we are masking a more foundational analysis by including these extra fields, what's the worst thing we're conceding? a nicer internal symmetry (the "unification" mentioned above, and is that even necessary)? hashing more fields doesn't seem inherently bad. 

Investigating these claims, cc @ebfull @alxiong.

alxiong:
That's a very good summary of my long-winded chain-of-thoughts. @TalDerei 
imo, i did provide analysis to answer these questions you extracted, but pls also try to see if you arrive the same conclusion or find gaps. Below contains pointers back to my long comment https://github.com/tachyon-zcash/ragu/pull/669#issuecomment-4359700543.

> are two routines with the same s(X, Y) contribution but different output gadget shapes / types equivalent under our model, assuming we proceed with this type erasure?

yes. it's safe to memoize in those case. also why we should drop `TypeId` from deep hash.
(I argue above via the example of two exact routine, but the output wire has different types, `Element` v.s. `MyElement`.)

> what’s the minimum property / invariant a fingerprint must capture for safe output-wire injection? and then work backwards from there.

what I posited is that output-wire injection safety requires _allocation determinism_ property. 
a short summary of that subsection is: parent routine receives some type-erased stream of output wires, and it knows the output gadget type, as long as the fungibility property of `GadgetKind` ensures a consistent allocation pattern, then any `Gadget` of that kind will be correctly reconstructed.

I did work backward from this simple reasoning to define "allocation determinism" property.

> (2) sub-question: can we remove extra fields in the deep hash, like output wire length,

yes we do; this also reminds me that we should add the length for number of children routines as well.
The preimage of this hash contains `(base_fingerprint, output_len, output_gadget, children_len, child_deep_hashes)`, 
- in the current definition of "fungibility" where gadget type also determine wire count, we only need `children_len` since parent routines has no info about that. 
- in the relaxed definition (with allocation determinism), we need both `output_len` and `children_len`

> (3) sub-question: if we are masking a more foundational analysis by including these extra fields, what's the worst thing we're conceding?

we are not including any more "extra fields" here. Imo the analysis is straightforward: 
- output length + the actual output gadget + allocation determinism => safe output injection
- base_fingerprint + deep fingerprints of child routines => recursive deep fingerprinting

that's all the preimages we are shoveling into deep hash. these fields are both sufficient and necessary imo.

TalDerei:
> yes. it's safe to memoize in those case. also why we should drop `TypeId` from deep hash.

https://github.com/tachyon-zcash/ragu/pull/694/changes/6a27f904a6d6243a05d88a108da3f3565d33a64e branches off https://github.com/tachyon-zcash/ragu/pull/503 (not #557, since it only has the fingerprint refactor and not the memoization layer that actually exercises substitution), and adds two differential type-erasure tests (using the strategy we discussed OOB where you keep the `TypeId` fields explicitly, but you set them as empty):
- (1) same s(X,Y) contribution, _same_ output gadget shape, _different_ input type and body,
- (2) same s(X,Y) contribution, _different_ output gadget shape, _same_ input type and body,

It seems to support our type erasure claims.

LucidSamuel (commented):
I cross-checked this against the current evaluators. the main implementation gap I see is around sx/sy memoization... algebraically the invariant routine pieces look reusable, but currently `sxy` is the only driver that is architecturally close to cache/replay. `sx` writes absolute coefficient-vector positions, and `sy` carries virtual-wire/floor-plan state, so supporting memoization there would require some refactoring rather than just turning on the same cache shape.

the Poseidon interface ratio claim also checks out: `15/576` constraints directly touch input/interface wires.

also while reading this path I noticed a small current-code optimization in `sx`/`sxy` and split it out separately: #745 

+ One design question this raised for me, for #613, should the next experiment be fused driver traversal, or routine-level `sxy` caching first?

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+Routines exploit the fact that contiguous sections of code tend to have
+algebraically convenient structure in the resulting wiring polynomial
+$s(X, Y)$:
+
+* [`gate`] advances a counter $i$ and returns
+  $(X^{2n - 1 - i}, X^{2n + i}, X^{4n - 1 - i}, X^i)$ wires.
+* `enforce_zero` advances a counter $j$ and adds a fresh linear
+  combination of previous wires multiplied by $Y^j$.
+
+Most constraints within a routine involve only wires created within that
```
LucidSamuel, line 148: This might be easier to absorb with one concrete local constraint example before introducing $g_1$ and $g_2$. For example, a routine at offset $(m,\ell)$ that emits a local constraint at relative offset $s$ over local wires at relative gate positions could contribute schematically:

$$
Y^{\ell+s}\left(\alpha X^{m+i_a} + \beta X^{-(m+i_b)} + \gamma
X^{-(m+i_c)}\right)
$$

Then the next paragraph can show that the common $X^m$, $X^{-m}$, and $Y^\ell$ placement factors are what get factored out into the reusable $g_1$/$g_2$ pieces.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+  constraint structure.
+- **Segment**: the polynomial footprint of an invocation — the contiguous
+  range of gate indices and $Y$-power slots it occupies in $s(X, Y)$. Each
+  invocation gets its own segment, and nested calls produce nested segments.
+
+It is possible to invoke a routine by manually calling [`Routine::execute`],
+but this defeats the abstraction. Routines are meant to be invoked through
+[`Driver::routine`], which preserves the call boundaries the driver needs.
+
+Routines can further invoke other routines, creating **nested routines**.
```
LucidSamuel, line 33: this paragraph is accurate, but it front-loads a lot of driver mechanics (`DriverScope`, `with_scope`, `mem::replace`) before the reader has seen why routine-local isolation matters algebraically. maybe we can consider moving the implementation details later, after the algebraic decomposition establishes why scope isolation enables relocation/memoization.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+Each label `r0, a0, b0, …, r2` denotes a contiguous section of circuit trace
+allocated via [`gate`]. At various points the circuit logic invokes a routine,
+which may in turn invoke nested routines.
+
+## Segments {#segments}
+
+Given an execution trace, we need a global indexing of all sub-sections so
+that different drivers can refer to them consistently. Ragu divides the trace
+into **segments** and orders them in a canonical **DFS order** determined by
+routine invocations. Circuit sections at the same invocation depth are merged
```
LucidSamuel, line 93: Given an execution trace, we need a global indexing of all sub-sections so that different drivers can refer to them consistently. Ragu divides the trace into **segments** and orders them by deterministic **DFS order** determined by routine invocations. Circuit sections at the same invocation depth are merged

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+
+The system contribution is memoizable by similar reasoning to the internal
+contribution. The `ONE` wire occupies $d_0$ at $X^0 = 1$, so its $X$-monomial
+is the constant 1, and the system contribution simplifies to:
+
+$$
+\sum_{s \in S} c_s \cdot Y^{\ell + s} = Y^\ell \cdot g_3(Y)
+$$
+
+where $S$ is the set of relative constraint offsets within the routine
```
LucidSamuel, line 206: The set $S$ contains the relative constraint offsets within the routine where ONE is referenced, and $c_s$ is its coefficient at offset $s$.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+subtree replays this folding step from the cached value directly.
+
+```admonish important title="Deep memoization redefines the internal contribution"
+Under deep-fingerprint equivalence, the output wires of nested sub-routines are
+no longer interface contributions — the deep fingerprint guarantees that child
+routines produce structurally identical output at identical relative positions.
+The parent can treat these wires as internal, collapsing the entire subtree into
+a single memoizable unit. This is the key advantage of the deep fingerprint: it
+enables full-tree memoization rather than segment-by-segment caching.
+```
```
LucidSamuel, line 565: This admonition restates the preceding paragraph closely. I think the paragraph already makes the key point: deep equivalence turns child output wires from interface data into internal subtree data. maybe consider cutting the box or shrinking it to one sentence.

book/src/implementation/routines.md
```diff
@@ -0,0 +1,697 @@
...
+segments, since collapsing an entire subtree yields the greatest savings.
+Alignment introduces implicit padding between segments in the global layout, but
+this padding does not materialize in memory: all downstream drivers process
+segments in a streaming fashion using absolute offsets from the floor plan, so
+padded regions are either skipped or processed at negligible cost. The trace
+polynomial $r(X)$ remains sparse, and padding does not increase the cost of the
+polynomial commitment. Base-fingerprint alignment is also applied when the base
+segment is large enough to justify the padding overhead.
+
+### Testability
```
LucidSamuel, line 649: just a thought: this section is useful, but it reads more like a design/test note than the rest of the book page. I think the book only needs the core claim: cached and uncached evaluation should agree at random $(w,x,y)$ points, and disagreement catches floor-planning, repositioning, and fingerprint over-grouping bugs.

LucidSamuel: Suggestion: Cut to 3-4 sentences: "Memoization bugs are detectable because they produce different s(x,y) values than uncached evaluation. The test strategy compares cached and uncached paths at random evaluation points."

LucidSamuel (commented):
Also left a few inline readability comments. The main structural thing I noticed is that the page sometimes introduces implementation mechanics before the reader has the algebraic motivation for them.

The most useful addition, IMO, would be one concrete local-constraint example before the $g_1/g_2$ decomposition. That would make the later relocation/memoization discussion easier to follow. The other comments are mostly about trimming or moving details that are already correct but show up a bit early.

GG.

ebfull (approved):

PR #670, opened 2026-04-20
feat(fv): extend coverage for existing `element / point` gadgets

TalDerei:
Builds on https://github.com/tachyon-zcash/ragu/pull/642 and references https://github.com/tachyon-zcash/ragu/pull/642#pullrequestreview-4140712859. The `Point` gadget implemented the [full stack of artifacts](https://github.com/tachyon-zcash/ragu/pull/642#discussion_r3102920311) for formal verification (rust instance / autogen / clean reimplementation / formal instance), but the field arithmetic primitives for `Element` only had the reimplementation (no standalone extraction). They were transitively verified through the `Point` gadget (ie. `Element`'s reimpl was composed as a subcircuit inside `Point`'s reimpl). This PR extends the FV surface area to _also_ cover that coverage gap (`Core::AllocMul, Element::Mul, Element::Square, Element::AllocSquare, Element::DivNonzero, Point::Endo`). cc @mitschabaude @gio54321.

I ran `lake build` locally to verify the extracted autogen matches the reimplementation at the lean kernel level. The extractor is trusted and  reimplementation is untrusted, and if the reimplementation is buggy, the equivalence theorem / proof will fail, which it doesn't. After the reimpl was verified (soundness / completeness closed), it was checked against the autogen with `~/.elan/bin/lake build -- --wfail`.

Follow-up PRs can cover the rest of `Point` / `Element` gadget components and other gadgets (`Boolean` / `Poseidon` / `Endoscalar`, etc.) that aren't yet implemented in _any_ capacity. 

cc @ebfull for visibility.

### DIFF
--- a/qa/crates/lean_extraction/src/instances/core_alloc_mul.rs
+++ b/qa/crates/lean_extraction/src/instances/core_alloc_mul.rs
@@ -1,0 +1,21 @@
+use ragu_arithmetic::Coeff;
+use ragu_core::drivers::Driver;
+use ragu_pasta::Fp;
+
+use crate::{
+    driver::ExtractionDriver,
+    expr::Expr,
+    instance::CircuitInstance,
+};
+
+pub struct CoreAllocMulInstance;
+
+impl CircuitInstance for CoreAllocMulInstance {
+    type Field = Fp;
+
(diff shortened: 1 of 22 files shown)


ebfull (approved):

mitschabaude (approved):
lgtm!

the main thing that bothers me is how much boilerplate this integration currently has. would love to reduce this to a minimum, will make reviews and everything else easier

PR #672, opened 2026-04-21
feat(fv): supplemental `element / point` gadget coverage

TalDerei:
Extends https://github.com/tachyon-zcash/ragu/pull/670, fleshing out the remaining element and point gadget components, with a few exceptions that either (1) depend on the boolean clean reimpl in the FV tree (`Element::is_zero/is_equal, Point::conditional_endo/conditional_negate`), or (2) are complex compositional gadgets (`double_and_add_incomplete`). That's considered follow-up work. I ran `lake build` after each commit and audited the specs to ensure correctness, cc @mitschabaude @gio54321.

The gadget components implemented here are **Element** (`negate, add, sub, double, scale, add_coeff, enforce_zero, invert_with, invert, alloc, sum, multiadd, fold, enforce_root_of_unity`) and **Point** (`constant`).


ebfull (approved):

qa/lean/Ragu/Circuits/Element/Add.lean
mitschabaude, line 1: this circuit has no operations (addition is just an expression, doesn't add a constraint). same as in ragu. you sure you need this?

qa/lean/Ragu/Circuits/Element/AddCoeff.lean
mitschabaude, line 1: empty circuit like Add

qa/lean/Ragu/Circuits/Element/Alloc.lean
```diff
@@ -0,0 +1,48 @@
...
+Zero))` and returns the middle wire — emitting a full 3-wire mul gate with
+two wires discarded. The returned wire is unconstrained (there always exist
+`a = 0`, `c = 0` satisfying `a · b = c` for any `b`). -/
+def main (hintReader : ProverHint (F p) → Core.AllocMul.Row (F p)) (_ : Unit)
+    : Circuit (F p) (Expression (F p)) := do
+  let ⟨_, b, _⟩ ← Core.AllocMul.circuit hintReader ()
+  return b
+
+def GeneralAssumptions (_hintReader : ProverHint (F p) → Core.AllocMul.Row (F p))
+    (_input : Unit) (_data : ProverData (F p)) (_hint : ProverHint (F p)) := True
```
mitschabaude: if it doesn't use the template argument hintReader, you should not make it an argument!

```suggestion
def GeneralAssumptions (_input : Unit) (_data : ProverData (F p)) (_hint : ProverHint (F p)) := True
```

qa/lean/Ragu/Circuits/Element/Alloc.lean
```diff
@@ -0,0 +1,48 @@
...
+    Core.AllocMul.circuit, Core.AllocMul.Assumptions
+  ]
+
+def generalCircuit (hintReader : ProverHint (F p) → Core.AllocMul.Row (F p))
+    : GeneralFormalCircuit (F p) unit field :=
+  { elaborated hintReader with
+    Assumptions := GeneralAssumptions hintReader,
+    Spec := GeneralSpec,
+    soundness := generalSoundness hintReader,
+    completeness := generalCompleteness hintReader }
```
mitschabaude: I don't know where the pattern to prefix everything with "general" comes from, but it's ugly. Just `circuit`, `Spec`, `soundness` etc is cleaner

```suggestion
def circuit (hintReader : ProverHint (F p) → Core.AllocMul.Row (F p))
    : GeneralFormalCircuit (F p) unit field :=
  { elaborated hintReader with
    Assumptions := Assumptions hintReader,
    Spec,
    soundness := soundness hintReader,
    completeness := completeness hintReader }
```

qa/lean/Ragu/Circuits/Element/Double.lean
mitschabaude, line 1: empty circuit

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
```diff
@@ -0,0 +1,54 @@
+import Clean.Circuit
+import Ragu.Circuits.Element.Mul
+
+namespace Ragu.Circuits.Element.EnforceRootOfUnity
+variable {p : ℕ} [Fact p.Prime]
+
+/-- `Element::enforce_root_of_unity(k)` enforces `self^(2^k) = 1` by
+squaring `k` times. This extraction instance fixes k = 2 (enforces
+`self^4 = 1`). Other values of k require analogous instances with the
+appropriate number of square subcircuits.
```
mitschabaude: ah that's a limitation of our extraction I guess. on the lean side you could still verify the general case though, by making k an argument and using a Circuit.foldl loop. maybe not worth it :D

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
```diff
@@ -0,0 +1,54 @@
...
+
+theorem completeness : GeneralFormalCircuit.Completeness (F p) elaborated Assumptions := by
+  circuit_proof_start [Mul.circuit, Mul.Assumptions, Mul.Spec]
+  obtain ⟨h_sq1, h_sq2⟩ := h_env
+  rw [add_neg_eq_zero]
+  rw [h_sq2, h_sq1]
+  exact h_assumptions
+
+def circuit : GeneralFormalCircuit (F p) field unit :=
+  { elaborated with Assumptions, Spec, soundness, completeness }
```
mitschabaude: in idiomatic `clean`, most circuits are either `FormalCircuit` for function-like circuits (completeness and soundness have the same assumptions, you can think of it as "constraints work for all valid inputs") and `FormalAssertion` for assertion-like circuits (constraints intentionally narrow down the allowed input, because the circuit adds constraints on the input)

this one would be a `FormalAssertion`.

`GeneralFormalCircuit` is the most flexible one but it'll cause you to duplicate stuff if the generality is not needed. Here you duplicate Assumptions = Spec. in FormalAssertion, you wouldn't.

maybe out of scope if that is not documented well for LLMs

qa/lean/Ragu/Circuits/Element/Fold.lean
```diff
@@ -0,0 +1,49 @@
...
+structure Input (F : Type) where
+  x0 : F
+  x1 : F
+  x2 : F
+  s : F  -- scale factor
+deriving ProvableStruct
+
+/-- Horner fold: `((x0 · s) + x1) · s + x2 = x0·s² + x1·s + x2`. -/
+def main (input : Var Input (F p)) : Circuit (F p) (Expression (F p)) := do
+  let acc0_s ← subcircuit Mul.circuit ⟨input.x0, input.s⟩
```
mitschabaude: you don't need `subcircuit`, a circuit is interpreted as a function already

```suggestion
  let acc0_s ← Mul.circuit ⟨input.x0, input.s⟩
```

qa/lean/Ragu/Circuits/Element/Invert.lean
```diff
@@ -0,0 +1,66 @@
...
+/-- `Element::invert(input)` computes `1/input` internally and delegates to
+`invert_with`. Under `MaybeKind = Empty`, the inverse-computation closure is
+erased, so the emitted trace is identical to invert_with's: one mul gate
+with `a = input`, `c = 1`, and `b = 1/input` as the output wire. -/
+def main (hintReader : ProverHint (F p) → Core.AllocMul.Row (F p)) (input : Expression (F p))
+    : Circuit (F p) (Expression (F p)) := do
+  let ⟨a, b, c⟩ ← Core.AllocMul.circuit hintReader ()
+  assertZero (a - input)
+  assertZero (c - 1)
+  return b
```
mitschabaude: if the Rust circuit delegates to another one (here: invert_with), the clean circuit should do the same, right? Then you can use the spec of that child circuit in your proof instead of reproving

qa/lean/Ragu/Circuits/Element/Multiadd.lean
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Ragu.Circuits.Element.Multiadd
+variable {p : ℕ} [Fact p.Prime]
+
+/-- `multiadd` is variadic (`&[Element]` + `&[F]`). This extraction
+instance fixes length 3 as a representative choice; a length-polymorphic
+reimpl would require dependent-type machinery that Clean doesn't
+currently expose. Parameterized over the three coefficients — the
+extraction pins specific values, but the reimpl's soundness/completeness
+are universal over any coefficient choice. -/
```
mitschabaude: > a length-polymorphic reimpl would require dependent-type machinery that Clean doesn't currently expose

this is a lie, we have plenty of length-polymorphic circuits in clean

IMO, the clean formalizations should represent the Rust circuits in their full generality even if the extraction bridge can't represent that. you could make both this circuit and the lean "instance" depend on the length.
then you could extract and check, like, 5 different concrete versions for extra confidence

qa/lean/Ragu/Circuits/Element/Negate.lean
mitschabaude, line 1: empty circuit

qa/lean/Ragu/Circuits/Element/Scale.lean
mitschabaude, line 1: empty circuit

qa/lean/Ragu/Circuits/Element/Sub.lean
mitschabaude, line 1: empty circuit

qa/lean/Ragu/Circuits/Element/Sum.lean
mitschabaude, line 1: empty circuit

qa/lean/Ragu/Circuits/Element/Sum.lean
```diff
@@ -0,0 +1,42 @@
+import Clean.Circuit
+
+namespace Ragu.Circuits.Element.Sum
+variable {p : ℕ} [Fact p.Prime]
+
+/-- `Element::sum` is variadic (`IntoIterator<Item = Element>`). This
+extraction instance fixes length 3 as a representative choice; a
+length-polymorphic reimpl would require dependent-type machinery that
+Clean doesn't currently expose. For other lengths, add analogous
```
mitschabaude: lie

qa/lean/Ragu/Circuits/Point/Constant.lean
mitschabaude, line 1: empty circuit.

do we need every method/helper function to become an empty, trivial circuit?

mitschabaude (commented):
some comments from me. I focused on the Lean part, I trust the extraction since it's very mechanical

qa/lean/Ragu/Circuits/Element/Add.lean
TalDerei: lifting verbatim from OOB discussion: "the empty circuits you noted were intentional for gadget completeness, but afaict aren't really a necessary addition". I'll get rid of all these empty circuits.

`Element.{Add, AddCoeff, Double, Multiadd, Negate, Scale, Sub, Sum}` + `Point.{Constant, Endo, Negate})` are zero-op gadgets (including some from https://github.com/tachyon-zcash/ragu/pull/670).

qa/lean/Ragu/Circuits/Element/Alloc.lean
TalDerei: this was just a stylistic choice early on while I was learning the syntax. but agreed, renaming.

qa/lean/Ragu/Circuits/Element/Fold.lean
TalDerei: oh didn't know that, fixing. slowly picking up these inference rules.

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
TalDerei: distinction makes sense, fixing these to narrow scope.

qa/lean/Ragu/Circuits/Element/Invert.lean
TalDerei: ah that's a good point, simplifying so it's consistent. the boolean impl in #674 also has some of those same patterns.

qa/lean/Ragu/Circuits/Element/Sum.lean
TalDerei: addressing per https://github.com/tachyon-zcash/ragu/pull/672#discussion_r3138786691.

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
TalDerei: yeah this was an extraction limit, was satisfied with this approach for now.

qa/lean/Ragu/Circuits/Element/Multiadd.lean
TalDerei: that's fair, so basically like lean formalizations should be fully general even when extraction isn't (referencing https://github.com/tachyon-zcash/ragu/pull/672#discussion_r3138707787). investigated a polymorphic refactor, but wasn't able to get `lake build` passing, so fixed the doc comment and filed an issue as interstitial measures.

TalDerei:
@mitschabaude pushed some commits that address your comments, and filed https://github.com/tachyon-zcash/ragu/issues/686 as follow-up work.

PR #673, opened 2026-04-21
book: improve consistency in NARK section

S1nus:
The example in the book for transforming revdot claims into polynomial products felt inconsistent, starting with a simplified example, then showing the real protocol's expanded version in the overview of the IOP.

This PR changes it to consistently use the example from the real IOP, complete with gate and wiring polynomials.


TalDerei:
@ebfull low-hanging fruit on queue.

TalDerei:
technically i think the hats on r(x) aren't correct, fixing.

TalDerei:
ACK this makes the vector form (in the arithmetization section) and polynomial form (in the nark section) consistent now. supply chain is failing, but that'll be fixed in another PR. cc @ebfull

TalDerei (approved):

PR #679, opened 2026-04-23
book: add S to instance

S1nus:
Seems like S should be considered part of the NARK instance.


TalDerei (approved):

TalDerei:
@ebfull low-hanging fruit on queue.

ebfull (approved):

Issue #3, opened 2025-10-08
Create accumulator and recursive proof structures


TalDerei:
https://github.com/zcash/halo2/issues/246 proposed using separate proof structures for IVC and PCD, which, for reasons to be delineated, are likely not justifiable against the additional engineering complexity such separation would inject into the internal layers. Instead, Ragu treats all recursion steps as _PCD-based_, even when only IVC semantics are required for a given step. This allows the use of a dummy second input to maintain a uniform structure. Although conjectural, it’s _likely_ the performance cost of two-path PCD over IVC is negligible, which optimistically motivates this design choice. Visually, this corresponds to an arity-2 PCD tree, where IVC emerges as the degenerate case with dummy accumulator inputs, forming a lopsided binary tree structure. 

In Ragu, we aim to expose a single proof structure capable of operating in two modes:

- **Uncompressed mode:** BCLMS21-style split-accumulation form that is non-succinct (not sublinear in the circuit size), with a large witness but inexpensive to generate.
- **Compressed mode:** a succinct form (logarithmic in the circuit size) using an IPA polynomial commitment scheme, with a more expensive verifier (outer decision procedure) that's dominated by a linear-time MSM. 

The recursion operates in “uncompressed” mode, and then we perform a “compression” step at certain boundary conditions for bandwidth reasons. For instance, in shielded transaction aggregation, broadcasting transaction data in the compressed mode optimizes for bandwidth. Naturally, we’ll also need some notion of “uncompressing” the compressed proof. 

This issue simply describes the shape of the proposed proof structures, while the actual in-circuit accumulation implementation will be described in a separate associated issue.   

-----------

```ts
pub struct AccumulatorWitness<C: CurveAffine, R: Rank> {
    pub s_poly: unstructured::Polynomial<C::Scalar, R>,
    pub s_blind: C::Scalar,

    pub a_poly: structured::Polynomial<C::Scalar, R>,
    pub a_blind: C::Scalar,

    pub b_poly: structured::Polynomial<C::Scalar, R>,
    pub b_blind: C::Scalar,

    pub p_poly: unstructured::Polynomial<C::Scalar, R>,
    pub p_blind: C::Scalar,
}

pub struct AccumulatorInstance<C: CurveAffine> {
    pub s_commitment: C,
    pub x: C::Scalar,
    pub y: C::Scalar,
    pub a_commitment: C,
    pub b_commitment: C,
    pub c: C::Scalar,
    pub p_commitment: C,
[...]

PR #682, opened 2026-04-23
pcd: add ggm example

turbocrime:
add an example GGM app, with tests and benches.

simple gadgets with methods are used to abstract handling of data within steps.

1. **GgmMasterSeed** ((), ()) -> GgmMasterHeader: a seed step produces the root of the ggm.
2. **GgmMasterStep** (GgmMasterHeader, ()) -> GgmPrivateHeader: the first descent into the ggm produces an unblinded private header.
3. **GgmNodeStep** (GgmPrivateHeader, ()) -> GgmPrivateHeader: the private header can be walked down the ggm.
4. **GgmBlindStep** (GgmPrivateHeader, ()) -> GgmDelegateHeader: any private header can be blinded into an equivalent delegate header.
5. **GgmDelegateStep** (GgmDelegateHeader, ()) -> GgmDelegateHeader: the delegate header can keep walking down the ggm.
6. **GgmNullifierStep** (GgmDelegateHeader, ()) -> GgmNullifierHeader: at the bottom of the ggm, a dedicated step conducts the output derivation.

tree parameters: arity 4, depth 6 (max leaf 4095).

tree node values and nullifiers are derived via tagged Poseidon hashes.

tests verify epoch distinctness, note distinctness, blind distinctness, and that blinding at the leaf vs. mid-path produces identical nullifiers.

also exposes `Application::seeded_trivial_pcd()` as pub, needed by step impls.


crates/ragu_testing/tests/ggm.rs
TalDerei, line 1: organizational nit; maybe we should place these kinds of tests / benches under an `applications` subdirectory?

turbocrime: moved impl to an apps directory in ragu_testing

moved all new benches and tests also to ragu_testing

turbocrime:
it seems like ggm bench exceeds the CI time limit? should we skip or change CI config?

crates/ragu_pcd/src/lib.rs
```diff
@@ -232,7 +232,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
     ///
     /// The proof is lazily created on first use and cached; subsequent calls
     /// return the same (non-random) proof.
-    fn seeded_trivial_pcd<RNG: CryptoRng>(&self, rng: &mut RNG) -> Pcd<C, R, ()> {
+    pub fn seeded_trivial_pcd<RNG: CryptoRng>(&self, rng: &mut RNG) -> Pcd<C, R, ()> {
```
turbocrime, line 235: only change outside of `ragu_testing` is the visibility of `Application::seeded_trivial_pcd`

crates/ragu_testing/src/apps/ggm.rs
```diff
@@ -0,0 +1,1264 @@
...
+    const SUFFIX: Suffix = Suffix::new(11);
+    type Data = (
+        PrefixKeyWitness<F>,
+        MasterKeyWitness<F>,
+        NoteCommitmentWitness<F>,
+    );
+    // Right-nested 2-tuple: only 2-tuples have a `GadgetKind` blanket impl.
+    type Output = Kind![F;
+        (PrefixKeyGadget<'_, _>, (MasterKeyGadget<'_, _>, NoteCommitmentGadget<'_, _>))
+    ];
```
turbocrime: most annoying part of handling data in headers.

if we want typed data both inside and outside the steps, we need to basically define it twice and do manual encode/decode.

it would be really cool if we had some way to use the same type in both contexts.

crates/ragu_pcd/src/lib.rs
TalDerei: appending OOB discussion for context on this change:

"some historical context https://github.com/tachyon-zcash/ragu/issues/3. we didn’t want a separate IVC concept (we wanted to make everything PCD, so even if you only need IVC in a step, you can just have a dummy second step). so we'll probably eventually expose something like `seeded_trivial_pcd` in the public API. 

halo2 was [considering](https://github.com/zcash/halo2/issues?q=is%3Aissue%20state%3Aopen%20IVC) supporting IVC, but we determined it wasn't worth the internal plumbing complexity or needing to be cognizant of a distinction."

crates/ragu_testing/tests/ggm.rs
```diff
@@ -0,0 +1,358 @@
...
+        poseidon_params,
+        NullifierKeyWitness(nk),
+        note.to_witness(),
+        DelegationTrapdoorWitness(trap),
+        u32::from(epoch),
+    )
+}
+
+#[test]
+fn ggm_epoch_distinctness() -> Result<()> {
```
TalDerei: I'd convert the separate distinctness tests into a unified proptest.

crates/ragu_testing/src/apps/ggm.rs
```diff
@@ -0,0 +1,1264 @@
...
+pub type GgmIndex = u16;
+pub const GGM_MAX: GgmIndex = (GGM_ARITY as u16).pow(GGM_DEPTH as u32) - 1;
+
+#[derive(Gadget, Write)]
+pub struct ChunkGadget<'dr, D: Driver<'dr>> {
+    pub inner: Element<'dr, D>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ChunkWitness([bool; GGM_CHUNK_SIZE as usize]);
```
TalDerei: nit; unused.

crates/ragu_testing/src/apps/ggm.rs
TalDerei, line 1: this doesn't exercise routines at this stage which is fine, but that would be an even more complex consumer of our API, but may also not useful here anyways.

also wanted to mention that your pattern of wrapping gadgets is actually idiomatic. our gadget API calls these 'compositional gadgets': "Gadgets can be composed of other gadgets by definition. Gadgets can even be polymorphic over gadgets, and some gadgets are even composed of gadgets that are instantiated with different drivers."

overall impression is header layout is good and the step trait usage is correct.

crates/ragu_testing/src/apps/ggm.rs
turbocrime: used now (and corrected derivation to absorb chunks per-bit)

crates/ragu_testing/src/apps/ggm.rs
turbocrime: working on adding routines now

crates/ragu_testing/tests/ggm.rs
turbocrime: a unified proptest actually seemed worse after trying it out, so i didn't do this.

Issue #185, opened 2025-12-04
Grouping stages optimization


TalDerei:
Stages currently consume _separate_ mesh slots, leading to redundant revdot claims. We may be able to optimize this by grouping stages using monomial-basis "shifts" (scaling by powers of X), reducing the number of staging polynomials and by extension revdot claims that need to be folded.

- [ ] `StageGroup` trait representing a group of stages with shared constraint structure,
- [ ] `register_stage_group` method on mesh for batch slot registration
- [ ] `shift_by` method on `structured::Polynomial` to multiply by `X^k`
- [ ]  `CircuitObject` extension (`sxy_grouped`) to handle multiple X-scaled stages in `s(X, Y)`

_(This ultimately might not be viable with our current staging model)._

alxiong:
TL;DR: my initial assessment is that this optimization is purely theoretical, but unrealizable in practice. 

iiuc, the idea is: say both stage A and B have a stage mask of size $m$, we can only register one and use shifting to deduce the other. This shifting is transparent to the verifier. Generalizing to multiple stages with varying `num_gates`, we take the GCD of them, and every stage mask can be decompose into a linear combination of this "gcd unit mask". So we could imagine registering only this "gcd mask", and _assemble_ the actual mask at revdot check time.
This theoretically saves the number of masking polynomial registered; smaller registry leads to smaller domain and faster `s(x,y)` eval etc. 

However, current stage mask size:

```
  Native bonding masks:

  ┌────────────┬───────────────────┬───────────┐
  │   Stage    │     values()      │ num_gates │
  ├────────────┼───────────────────┼───────────┤
  │ Preamble   │ 2*(3*4+1+29) = 84 │ 42        │
  ├────────────┼───────────────────┼───────────┤
  │ OuterError │ 342+19+6+5 = 372  │ 186       │
  ├────────────┼───────────────────┼───────────┤
  │ InnerError │ 19*42 = 798       │ 399       │
  ├────────────┼───────────────────┼───────────┤
  │ Query      │ 13+1+32 = 46      │ 23        │
  ├────────────┼───────────────────┼───────────┤
  │ Eval       │ 30+6 = 36         │ 18        │
  └────────────┴───────────────────┴───────────┘

  Nested bonding masks (chain: PointsStage → Preamble → s_prime → inner_error → outer_error → ab → query → f → eval):

  ┌─────────────┬───────────────────┐
  │    Stage    │     num_gates     │
[...]

PR #683, opened 2026-04-24
Relocate `ONE` wire

### LINKED ISSUE #678: SYSTEM gate arrangement and monomial structure
Structured polynomials take the following form by convention:

$$
p(X) = \sum_{i=0}^{n-1} \left(
  \gamma_i\, X^{i} +
  \beta_i\, X^{2n-1-i} +
  \alpha_i\, X^{2n+i} +
  \delta_i\, X^{4n-1-i}
\right)
$$

The four positions split into two direction classes: $\{\alpha, \gamma\}$ are *forward* (index $i$ ascending in monomial degree within their blocks) and $\{\beta, \delta\}$ are *reverse*.

### Degrees of freedom

The gate equations $ab = c$ and $cd = 0$ are fixed, but their assignment to $(\alpha, \beta, \gamma, \delta)$ is not. The mul inputs $a/b$ pair with each other, and $c/d$ pair with each other, so $a/b$ can go to either the $\{\alpha, \beta\}$ or $\{\gamma, \delta\}$ pair, with $c/d$ taking the other.

The `SYSTEM` gate reserves one wire for `ONE = 1` and another for a blinding factor $\zeta$. Given the gate equations, the viable configurations (labeled by $(a_0, b_0, c_0, d_0)$ ) are:

1. $(\text{ONE}, 0, 0, \zeta)$
2. $(0, \text{ONE}, 0, \zeta)$ — **current**
3. $(\zeta, 0, 0, \text{ONE})$
4. $(0, \zeta, 0, \text{ONE})$
5. $(\zeta, \zeta^{-1}, \text{ONE}, 0)$
6. $(\zeta^{-1}, \zeta, \text{ONE}, 0)$

Separately, each gate supports two-wire allocation — two wires that can be assigned arbitrary values. The viable patterns are:

1. $(a, b, ab, 0)$
2. $(a, 0, 0, d)$
3. $(0, b, 0, d)$ — **current**

### Preferences

1. **Simple wiring monomial for `ONE`.** Ideally $X^0 = 1$, since ONE is referenced pervasively throughout circuit code and a trivial monomial simplifies evaluation.
2. **Same-direction allocation**, to support shifting (see #185). The two allocated slots should be in the same direction class.
3. **Symmetric allocation pattern.** $(a, 0, 0, d)$ is symmetric; $(0, b, 0,d)$ is not.
4. **Consistent allocation pattern with the `SYSTEM` gate layout.**

### Current scheme vs. preferences

The current layout is $(0, \text{ONE}, 0, \zeta)$, with allocation $(0, b, 0, d)$:

- (1) ✗ — ONE sits at $\beta$, wiring monomial $X^{2n}$.
- (2) ✓ — allocation uses $\{\beta, \delta\}$, both reverse.
- (3) ✗ — allocation is $(0, b, 0, d)$, asymmetric.
- (4) ✓ — allocation is $(0, b, 0, d)$, and `SYSTEM` gate is $(0, \text{ONE}, 0, \zeta)$

### Proposed adjustment

Three changes, together:

1. Rename $a \leftrightarrow b$, so $a$ takes the reverse slot and $b$ takes the forward slot.
2. Place `ONE` at $d_0$.
3. Place $\zeta$ at $a_0$.

Under the rename, the physical positions become: ONE at $\delta$, $\zeta$ at $\beta$, with the free-wire pair being $\{\alpha, \delta\}$ in letter terms but $\{\beta, \delta\}$ in physical position — both reverse.

Checking preferences:

- (1) ✓ — ONE is at $\delta$, wiring monomial $X^0$.
- (2) ✓ — allocation is at $\{\beta, \delta\}$, both reverse.
- (3) ✓ — in the new labels, allocation reads $(a, 0, 0, d)$.
- (4) ✓ — they're consistent

All four preferences satisfied. In physical terms this is $(0, \zeta, 0, \text{ONE})$, with the $a \leftrightarrow b$ rename applied on top for cosmetic (3).

### Conclusion

Adopt the renaming $a \leftrightarrow b$. Concretely:

$$
  p(X) = \sum_{i=0}^{n-1} \left(
    c_i\, X^{i} +
    a_i\, X^{2n-1-i} +
    b_i\, X^{2n+i} +
    d_i\, X^{4n-1-i}
  \right)
$$

Use the allocator pattern $(a, 0, 0, d)$.

Use the `SYSTEM` gate $a_0 = \zeta$, $d_0 = \text{ONE} = 1$. The others are forced to $0$.

ebfull:
Closes #678.

`ONE` moves to the last wire of the trace, so that its wiring monomial becomes $X^0 = 1$. This simplifies several of the drivers. We also swap the allocation pattern so that it becomes `(a, 0, 0, d)`, and swap $a$ and $b$ in structured polynomials to maintain the "direction class" of the two allocated wire groups (used in stages) to support #185 optimizations if they ever land.

My favorite part of this PR is https://github.com/tachyon-zcash/ragu/commit/afe6e24f75bfc4e9e91eb8750c8e0005d53bfc62.

I explored some other modifications as well, but none of them could satisfy every nice-to-have that I had. Even giving up $r(0) = 0$ wasn't ultimately worth it in any of the alternative arrangements. I'm hoping we can call this arrangement solved.


crates/ragu_circuits/src/tests/identity.rs
```diff
@@ -1732,7 +1732,7 @@ fn test_vanishing_leading_trivial_via_eval() {
 /// DelegatePadEnforceOutput enforces the remapped child output wire;
 /// DelegateAllocEnforceFirst enforces a subsequent local alloc.  The
 /// output wire carries a value from the parent's `x_remap` sequence,
-/// while the local alloc carries a value from the `x1` (b-wire)
+/// while the local alloc carries a value from the `x1` (a-wire)
```
ebfull:
```suggestion
/// while the local alloc carries a value from the `x0` (a-wire)
```

crates/ragu_circuits/src/metrics.rs
```diff
@@ -321,13 +310,15 @@ impl<F: FromUniformBytes<64>> Counter<F> {
...
+        // PRF index 6 is unused: the ONE wire evaluates to F::ONE directly
+        // (see `Driver::ONE` below), which cannot collide with any
+        // geometric-sequence wire value when x0..x3 are pseudorandom.
         let x0 = point(0);
         let x1 = point(1);
         let x2 = point(2);
         let x3 = point(3);
         let y = point(4);
         let h = point(5);
-        let one = point(6);
```
ebfull: These should just be renumbered.

ebfull:
Force pushing to rebase after #656...

crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -353,42 +359,26 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
             });
         }
 
-        assert!(values.len() <= Self::values());
-
-        let mut values = values.into_iter();
-        let mut view = sparse::View::trace();
+        let mut dr = RxDriver::<F, R>::with_capacity(Self::skip_gates() + Self::num_gates());
```
TalDerei, line 362: ACK - `RxDriver` is layout agnostic (which is an nice encapsulation!) and instead that decision lives in `Standard::alloc`.

crates/ragu_circuits/src/trace.rs
```diff
@@ -138,10 +138,10 @@ impl<F: Field> Trace<F> {
...
 
         // Overwrite segment 0's zeroed SYSTEM gate placeholder:
-        // a[0] = c[0] = 0 (already zero), b[0] = 1 (ONE wire),
-        // d[0] = alpha (prevents point-at-infinity commitments).
-        view.b[0] = F::ONE;
-        view.d[0] = alpha;
+        // a[0] = c[0] = 0 (already zero), d[0] = 1 (ONE wire),
+        // b[0] = alpha (prevents point-at-infinity commitments).
+        view.b[0] = alpha;
+        view.d[0] = F::ONE;
```
TalDerei: setting `view.b[0] = alpha` is confusing here because we swap a / b monomial positions in https://github.com/tachyon-zcash/ragu/pull/683/changes/12b9fc3ddf663951c0841fec63647d778332d606, but the the trace didn't change. there's a inconsistency here between the code and the closing issue / book around what gets placed in reverse / forward blocks in the trace.

i'd set

```ts
  // b[0] = c[0] = 0 (already zero), d[0] = 1 (ONE wire),
  // a[0] = alpha (prevents point-at-infinity commitments).
  view.a[0] = alpha;  
  view.d[0] = F::ONE; 
```

and change `staging/mod.rs` (along with comment updates in staging / registry files) to 

```ts
  let (_, _, _, extra) =
      dr.gate(|| Ok((Coeff::Arbitrary(alpha), Coeff::Zero, Coeff::Zero)))?;                                                                                                                                                                          
  dr.assign_extra(extra, || Ok(Coeff::Zero))?;  
``` 

and then obviously repin the registry digests.

crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -353,42 +359,27 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
             });
         }
 
-        assert!(values.len() <= Self::values());
```
TalDerei, line 356: why was this assertion removed? always weary of these kinds of deletions unless it's truly extraneous in this context.

crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -353,42 +359,27 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
...
+        let (_, _, _, extra) =
+            dr.gate(|| Ok((Coeff::Zero, Coeff::Arbitrary(alpha), Coeff::Zero)))?;
+        dr.assign_extra(extra, || Ok(Coeff::Zero))?;
 
         for _ in 0..(Self::skip_gates() - 1) {
-            view.a.push(F::ZERO);
-            view.b.push(F::ZERO);
-            view.c.push(F::ZERO);
-            view.d.push(F::ZERO);
+            // d[0] = 0 is default
```
TalDerei:
```suggestion
            // padding gate: d-wire defaults to F::ZERO (no assign_extra follows)

File: crates/ragu_circuits/src/staging/mod.rs [hunk 7]
```diff
@@ -353,42 +359,27 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
...
 
-        // Layout: (0, b, 0, d) per gate.
-        for _ in 0..Self::num_gates() {
-            let b = values.next().unwrap_or(F::ZERO);
-            let d = values.next().unwrap_or(F::ZERO);
-            view.a.push(F::ZERO);
-            view.b.push(b);
-            view.c.push(F::ZERO);
-            view.d.push(d);
+        let mut allocator = Standard::<usize>::new();
```
TalDerei: nit; turbofish

```suggestion
        let mut allocator = Standard::default();

**Review by TalDerei — COMMENTED** [review]
left some suggestions that align more closely with the closing issue.

File: crates/ragu_circuits/src/trace.rs (hunk 4 above)
Comment (ebfull) [reply crates/ragu_circuits/src/trace.rs]: Nice catch. This will not affects the digest though. As the book mentions, it is "convention" that the `alpha` blinding value be placed in the `a` wires; technically it can appear in the `b` wires. This should be consistent with the allocation process though.

File: crates/ragu_circuits/src/staging/mod.rs (hunk 5 above)
Comment (ebfull) [reply crates/ragu_circuits/src/staging/mod.rs:356]: claude:

> The early `if values.len() > Self::values() { return Err(...) }` exists unchanged a few lines above, so the removed `assert!` was a duplicated belt-and-suspenders check that the new code drops as a cleanup. Functionally, the bound is still enforced; the error path is identical. So it's a benign deletion, not a regression.

**Review by TalDerei — APPROVED** [review]
ACK second pass after cleanup.

File: book/src/protocol/local/wiring.md [hunk 8]
```diff
@@ -1,108 +1,129 @@
...
+equal $1$ through the final constraint, via $k(0) = s(X, 0) = X^0 = 1$.
+
+The $\color{#7e22ce}{\v{c}_0 = 0}$ wire assignment ensures $r(0) = 0$ for every
+trace, and this is forced by the gate equations when $\color{blue}{\v{d}_0 =
+1}$. Conventionally, we allow $\color{#dc2626}{\v{a}_0}$ to carry an arbitrary
+(blinding) value $\color{#dc2626}{\alpha}$ for all traces, and thus $\v{b}_0 =
+0$ is assigned so that the gate equations are always satisfied.
+
+The constraint for $Y^{4n - 1}$ is reserved for the
+[registry](../extensions/registry.md), which injects a fixed "key" value
```
alxiong: explain that we didn't squeeze this into the system constraint because we want it in a fixed position AND extractable for ALL wiring polynomials.

PR #684, opened 2026-04-24
ci: avoid jq panic when no PR_NUMBER

alxiong:
fixing a CI failure notification: https://github.com/tachyon-zcash/ragu/actions/runs/24870248395/job/72814994477

So return a fallback `[]` which is a valid json. our `scripts/find-pr` already assign `pr_number = ""` empty string if not found, and rest of the workflow will cancel -- this is the correct behavior.

### DIFF
--- a/.github/workflows/bench-alert.yml
+++ b/.github/workflows/bench-alert.yml
@@ -26,14 +26,19 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          PR_NUMBER=$(gh pr list \
-              --repo ${{ github.repository }} \
-              --state open \
-              --json number,headRefName,headRefOid,headRepositoryOwner \
+          PR_NUMBER=$(
+            (
+              gh pr list \
+                --repo ${{ github.repository }} \
+                --state open \
+                --json number,headRefName,headRefOid,headRepositoryOwner \
+              || echo '[]'
+            ) \
             | jq --raw-output --from-file .github/scripts/find-pr.jq \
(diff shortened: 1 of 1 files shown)


.github/workflows/bench-alert.yml
```diff
@@ -26,14 +26,15 @@ jobs:
...
           GH_TOKEN: ${{ github.token }}
         run: |
-          PR_NUMBER=$(gh pr list \
+          PR_NUMBER=$( (gh pr list \
               --repo ${{ github.repository }} \
               --state open \
               --json number,headRefName,headRefOid,headRepositoryOwner \
-            | jq --raw-output --from-file .github/scripts/find-pr.jq \
+              || echo '[]') \
+            | jq -r --from-file .github/scripts/find-pr.jq \
```
turbocrime:
```suggestion
            | jq --raw-output --from-file .github/scripts/find-pr.jq \
```

nit: restore long option name.

i prefer to use long option names in scripts because they're more readable

.github/workflows/bench-alert.yml
```diff
@@ -26,14 +26,15 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          PR_NUMBER=$(gh pr list \
+          PR_NUMBER=$( (gh pr list \
```
turbocrime: it might be a good idea to indent the new parens scope to make it clear what the grouping is

turbocrime (approved):

TalDerei (approved):

TalDerei:
closing as outdated per stricter CI workflow changes in https://github.com/tachyon-zcash/ragu/pull/729.

PR #688, opened 2026-04-27
feat(circuits): group stage masks

### LINKED ISSUE #185: Grouping stages optimization

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/185 (the issue was opened prematurely and actual impl diverges, see design decisions in https://github.com/tachyon-zcash/ragu/issues/185#issuecomment-4380341389).

The precondition for this kind of stage mask grouping optimization was "same-direction allocation" [support](https://github.com/tachyon-zcash/ragu/issues/678), which enables an implicit shifting property (a-block and d-block shift in lockstep). We can now group `OuterError / InnerError` and `Query / Eval` stage mask registrations into a single slot separately, along with their associated bonding claims. The side-effect of this slot reshuffling is (1) registry digest changes, and (2) constraint reduction (proportional to the quadratic asymptotics from the `NumGroups` parameter). This only touches the native-side. Fuse savings here are ~5%, with verifier savings ~10%.

### DIFF
--- a/crates/ragu_circuits/src/staging/mask.rs
+++ b/crates/ragu_circuits/src/staging/mask.rs
@@ -1,3 +1,5 @@
+use alloc::{vec, vec::Vec};
+
 use ff::Field;
 use ragu_arithmetic::geosum;
 use ragu_core::Result;
@@ -85,10 +87,14 @@ pub(crate) fn global_project<F: Field, R: Rank>(p: F) -> sparse::Polynomial<F, R
     view.build()
 }
 
+/// Bonding polynomial that masks the data wires outside one or more
+/// stage windows. A single-stage mask is constructed via [`new`](Self::new);
+/// multiple stages can be bundled into one registry slot by chaining
+/// [`with_notch`](Self::with_notch).
 #[derive(Clone)]
 pub struct StageMask<R: Rank> {
-    skip_gates: usize,
-    num_gates: usize,
+    /// One `(skip_gates, num_gates)` per masked stage window.
+    notches: Vec<(usize, usize)>,
     _marker: core::marker::PhantomData<R>,
 }
 
@@ -102,13 +108,9 @@ impl<R: Rank> StageMask<R> {
     /// `d[0]` may or may not be set to 1; `b[0]` and `c[0]` are
     /// zero in all cases.
     pub fn new(skip_gates: usize, num_gates: usize) -> Result<Self> {
-        assert!(skip_gates > 0, "skip_gates must include the SYSTEM gate");
-        if skip_gates + num_gates > R::n() {
-            return Err(ragu_core::Error::GateBoundExceeded { limit: R::n() });
-        }
+        Self::check_notch(skip_gates, num_gates)?;
         Ok(Self {
-            skip_gates,
-            num_gates,
+            notches: vec![(skip_gates, num_gates)],
             _marker: core::marker::PhantomData,
         })
     }
@@ -123,50 +125,77 @@ impl<R: Rank> StageMask<R> {
         if skip_gates > R::n() {
             return Err(ragu_core::Error::GateBoundExceeded { limit: R::n() });
         }
-
         let num_gates = R::n() - skip_gates;
-
         Ok(Self {
-            skip_gates,
-            num_gates,
+            notches: vec![(skip_gates, num_gates)],
             _marker: core::marker::PhantomData,
         })
     }
 
+    /// Bundles an additional stage window into this mask, occupying the
+    /// same registry slot as the original. Caller must ensure the
+    /// bundled stages occupy disjoint gate windows.
+    #[allow(dead_code, reason = "exposed for future stage-grouping migrations")]
+    pub fn with_notch(mut self, skip_gates: usize, num_gates: usize) -> Result<Self> {
+        Self::check_notch(skip_gates, num_gates)?;
+        self.notches.push((skip_gates, num_gates));
+        Ok(self)
+    }
+
+    fn check_notch(skip_gates: usize, num_gates: usize) -> Result<()> {
+        assert!(skip_gates > 0, "skip_gates must include the SYSTEM gate");
+        if skip_gates + num_gates > R::n() {
+            return Err(ragu_core::Error::GateBoundExceeded { limit: R::n() });
+        }
+        Ok(())
+    }
+
     /// Returns a sparse polynomial containing the negated notch entries for
-    /// the active-stage gate wires at positions `skip_gates..skip_gates + num_gates`.
-    /// The SYSTEM gate is excluded (it is already zero in the global mask).
+    /// the active-stage gate wires at positions `skip_gates..skip_gates + num_gates`,
+    /// summed across every notch in the mask. The SYSTEM gate is excluded
+    /// (it is already zero in the global mask).
     ///
-    /// The result has `4 × num_gates` non-zero entries
-    /// (vs `4(n - 1)` for the full mask projection).
+    /// For a single-stage mask the result has `4 × num_gates` non-zero
+    /// entries; bundled masks accumulate per-notch contributions into the
+    /// same view (notches are expected to occupy disjoint windows, so no
+    /// real accumulation occurs in well-formed cases).
     fn notch_project<F: Field>(&self, p: F) -> sparse::Polynomial<F, R> {
-        if self.num_gates == 0 || p == F::ZERO {
+        if p == F::ZERO {
             return sparse::Polynomial::default();
         }
+
+        let max_extent = self.notches.iter().map(|(g, m)| g + m).max().unwrap_or(0);
+        if max_extent == 0 {
+            return sparse::Polynomial::default();
+        }
+
         let n = R::n();
-        let g = self.skip_gates;
-        let m = self.num_gates;
         let mut view = sparse::View::<F, R, _>::wiring();
-        view.d.resize(g + m, F::ZERO);
-        view.a.resize(g + m, F::ZERO);
-        view.b.resize(g + m, F::ZERO);
-        view.c.resize(g + m, F::ZERO);
+        view.d.resize(max_extent, F::ZERO);
+        view.a.resize(max_extent, F::ZERO);
+        view.b.resize(max_extent, F::ZERO);
+        view.c.resize(max_extent, F::ZERO);
 
         let p_inv = p.invert().expect("p is not zero");
-        let mut d = -p.pow_vartime([g as u64]);
-        let mut a = -p.pow_vartime([(2 * n + g) as u64]);
-        let mut b = -p.pow_vartime([(2 * n - 1 - g) as u64]);
-        let mut c = -p.pow_vartime([(4 * n - 1 - g) as u64]);
+        for &(g, m) in &self.notches {
+            if m == 0 {
+                continue;
+            }
+            let mut d = -p.pow_vartime([g as u64]);
+            let mut a = -p.pow_vartime([(2 * n + g) as u64]);
+            let mut b = -p.pow_vartime([(2 * n - 1 - g) as u64]);
+            let mut c = -p.pow_vartime([(4 * n - 1 - g) as u64]);
 
-        for j in g..g + m {
-            view.d[j] = d;
-            view.a[j] = a;
-            view.b[j] = b;
-            view.c[j] = c;
-            d *= p;
-            a *= p;
-            b *= p_inv;
-            c *= p_inv;
+            for j in g..g + m {
+                view.d[j] += d;
+                view.a[j] += a;
+                view.b[j] += b;
+                view.c[j] += c;
+                d *= p;
+                a *= p;
+                b *= p_inv;
+                c *= p_inv;
+            }
         }
 
         view.build()
@@ -174,17 +203,19 @@ impl<R: Rank> StageMask<R> {
 }
 
 impl<F: Field, R: Rank> WiringObject<F, R> for StageMask<R> {
-    /// Evaluates the per-instance notch term $-\text{notch}(x, y)$.
+    /// Evaluates the per-instance notch term $-\text{notch}(x, y)$,
+    /// summed across every notch in the mask.
     ///
     /// The full stage mask is $S\_{\text{global}} - \text{notch}$, but
     /// the invariant $S\_{\text{global}}$ term is factored out and applied
-    /// once by the [`Registry`](crate::registry::Registry). This method
-    /// returns only $-\text{notch}$, where
+    /// once by the [`Registry`](crate::registry::Registry). For each
+    /// `(skip_gates, num_gates)` pair this method contributes
     ///
     /// $$\text{notch}(x, y) = (1 + (xy)^{2n})\bigl((xy)^g + (xy)^{2n-g-m}\bigr)
     ///     \cdot \sum_{i=0}^{m-1} (xy)^i$$
     ///
-    /// with $g = \text{skip\_gates}$ and $m = \text{num\_gates}$.
+    /// with $g = \text{skip\_gates}$ and $m = \text{num\_gates}$. The
+    /// $(1 + (xy)^{2n})$ factor is shared across notches and applied once.
     fn sxy(&self, x: F, y: F, _floor_plan: &[crate::floor_planner::ConstraintSegment]) -> F {
         if x == F::ZERO || y == F::ZERO {
             return F::ZERO;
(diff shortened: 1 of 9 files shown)


crates/ragu_pcd/src/internal/tests.rs
```diff
@@ -106,10 +106,10 @@ fn test_internal_stage_parameters() {
...
 
     check_stage!(Preamble, skip =   1, num = 225);
-    check_stage!(OuterError,  skip = 226, num = 186);
-    check_stage!(InnerError,  skip = 412, num = 399);
-    check_stage!(Query,   skip = 226, num =  23);
-    check_stage!(Eval,    skip = 249, num =  18);
+    check_stage!(OuterError,  skip = 226, num = 134);
+    check_stage!(InnerError,  skip = 360, num = 336);
+    check_stage!(Query,   skip = 226, num =  22);
+    check_stage!(Eval,    skip = 248, num =  18);
```
TalDerei, line 157: what wasn't intuitive until impl time: i'd been assuming you could collapse native stage masks (`Preamble, OuterError, InnerError, Query, Eval`) into a single stage mask. the obvious issue: the windows overlap because the parent chain is a tree (so `OuterError` and `Query` overlap), so we can only collapse groups within sibling chains that are non-overlapping.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -193,11 +223,44 @@ impl<F: Field, R: Rank> WiringObject<F, R> for StageMask<R> {
         let xy = x * y;
         let xy_2n = xy.pow_vartime([2 * R::n() as u64]);
 
-        let gsum = geosum(xy, self.num_gates);
-        let skip = xy.pow_vartime([self.skip_gates as u64]);
-        let tail = xy.pow_vartime([(2 * R::n() - self.skip_gates - self.num_gates) as u64]);
+        // If every active notch shares the same `num_gates`, factor
+        // `geosum(xy, m)` out of the sum and apply it once to the sum of
+        // offsets. Otherwise, evaluate `geosum` per notch.
```
TalDerei: _optional_ optimization vector (production path isn't wired up to exploit this yet): if we make stage masks uniform width, we can pull the geosum calculation out of being computed per notch loop and only compute it once per group. padding the stages like this shouldn't actually affect cost since we can exploit the sparsity (padding adds zeros, and sparse storage drops them).

for more context, this is the algebraic X^k _shift_ from the closing issue which again can only happen when the stages and their notches share the same `num_gates`, ie. the same stage width.

crates/ragu_pcd/src/internal/native/claims.rs
```diff
@@ -221,17 +221,19 @@ where
...
+                processor.grouped_bonding_claim(id, groups)?;
             }
-            EvalStage => {
-                processor.bonding_claim(id, source.rx(Rx(Eval)))?;
+            QueryEvalGroup => {
+                let groups = source
+                    .rx(Rx(Query))
+                    .zip(source.rx(Rx(Eval)))
+                    .map(|(q, e)| [q, e].into_iter());
+                processor.grouped_bonding_claim(id, groups)?;
```
TalDerei, line 236: `OuterInnerErrorGroup` and `QueryEvalGroup` can now exploit the linearity in `grouped_bonding_claim`, summing per-stage rx contributions. instead of N separate revdot checks (for well-formedness), we can sum per-stage rxs and the per-stage masks, then do one combined check for each group.

crates/ragu_pcd/src/internal/tests.rs
```diff
@@ -154,6 +193,35 @@ fn print_internal_circuit_constraint_counts() {
     }
 }
 
+/// Helper test to print current nested constraint counts in copy-pasteable format.
+/// Run with: `cargo test -p ragu_pcd --release print_nested_internal_circuit -- --nocapture`
+#[test]
+fn print_nested_internal_circuit_constraint_counts() {
```
TalDerei, line 205: mimics the native variant to catch drift.

crates/ragu_circuits/src/staging/mask.rs
TalDerei: update, this doesn't seem to currently work because the extra padding in  `OuterError` / `InnerError` exceeds our recursion threshold (`Hashes1` circuit would exceed `R::n() = 2048`). a _optional_ follow-up would be to reduce `Hashes1` body to free up room. 

dropped the commit during rebase that attempted to introduce this change, since this was a minor optimization anyways. 

**experimented with this, but it hasn't fully been implemented yet, locally tracking.**

crates/ragu_pcd/src/internal/native/mod.rs
TalDerei, line 378: are native `InnerErrorFinalStaged`, `OuterErrorFinalStaged`, `EvalFinalStaged` also bundleable?

TalDerei: no, `final_mask()` covers tail of the trace `[skip_gates + num_gates, R::n())`, so obviously these windows nest with each other, not _disjoint_. they overlap with each other, and bundling would sum overlapping notches.

crates/ragu_pcd/src/internal/native/claims.rs
TalDerei, line 1: we should nail down the soundness argument for the bundled impl. like the bundled identity shouldn't be strictly weaker than the per-stage revdots it replaces? cc @ebfull

crates/ragu_pcd/src/internal/native/mod.rs
```diff
@@ -339,23 +329,37 @@ pub fn register_all<'params, C: Cycle, R: Rank, const HEADER_SIZE: usize>(
             PreambleStage => {
                 registry.register_bonding(stages::preamble::Stage::<C, R, HEADER_SIZE>::mask()?)
             }
```
TalDerei, line 331: think `PreambleStage` could also be bundled into either `OuterError` / `InnerError` or `Query` / `Eval` group for more savings.

**this hasn't been implemented yet, locally tracking.**

TalDerei:
there are some open questions commented below and follow-up work that i'll tackle (in the same PR). but first opening this up for discussion before further changes are made.

crates/ragu_primitives/src/allocator.rs
```diff
@@ -83,6 +83,11 @@ impl<'dr, D: Driver<'dr>> Allocator<'dr, D> for () {
 ///
 /// Dropping a `Standard` with tokens still in the pool is safe:
 /// the driver already assigned $D = 0$ for those gates.
+///
+/// Per-gate this produces the `(a, 0, 0, d)` trace pattern. This
+/// pattern is load-bearing for stage-mask grouping in `ragu_circuits`:
+/// changing it (e.g., placing data in `b` or `c`) would silently
+/// break the bundled bonding-claim identity.
```
alxiong, line 90: nit: do we need explain it in this context? add more confusion than clarity imo. suggest removing.

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -123,68 +125,112 @@ impl<R: Rank> StageMask<R> {
...
         })
     }
 
+    /// Bundles an additional stage window into this mask, occupying the
+    /// same registry slot as the original. Panics if the new window
+    /// overlaps any existing notch — disjointness is required for the
+    /// bundled bonding identity to remain sound.
+    pub fn with_notch(mut self, skip_gates: usize, num_gates: usize) -> Result<Self> {
+        Self::check_notch(skip_gates, num_gates)?;
+        if num_gates > 0 {
```
alxiong, line 141: nit: just delegate `assert(num_gates > 0)` into `self.check_notch()`?

on this note, why would we ever allow an empty notch/stage, right?

crates/ragu_circuits/src/staging/mask.rs
```diff
@@ -665,6 +763,33 @@ mod tests {
...
+                .unwrap();
+
+            let p = Fp::random(&mut rand::rng());
+            let x = Fp::random(&mut rand::rng());
+            let y = Fp::random(&mut rand::rng());
+            let q = Fp::random(&mut rand::rng());
+
+            let mut sum_poly = mask_a.notch_project(p);
+            sum_poly += &mask_b.notch_project(p);
+            prop_assert_eq!(sum_poly.eval(q), combined.notch_project(p).eval(q));
```
alxiong, line 1001: btw, I also used this way to check equality before. should we implement `PartialEq` for sparse poly somewhere? and clean up all the unit tests that has to do probabilistic equality test.

crates/ragu_circuits/src/staging/mod.rs
```diff
@@ -434,3 +434,28 @@ pub trait StageExt<F: Field, R: Rank>: Stage<F, R> {
...
 impl<F: Field, R: Rank, S: Stage<F, R>> StageExt<F, R> for S {}
+
+/// Bundles multiple stage windows into a single bonding polynomial that
+/// occupies one registry slot. Each `(skip_gates, num_gates)` entry
+/// corresponds to one bundled stage; the resulting polynomial is the sum
+/// of the per-stage notches plus the shared global term added once by the
+/// [`Registry`](crate::registry::Registry).
+///
+/// Soundness depends on stage rxs living only in the `(a, 0, 0, d)` wire
+/// pattern — currently provided by
```
alxiong, line 446: nit: `rxs` -> "stage traces `rxs`", 
nit2: let's clarify that it's not about something magical/particular about `a,d` wires, but the mere fact that they grow in the same direction, so that $g+m$ succinctly and consistently across the 4 (n-size sub-trace) signal its region occupied.

alxiong (commented):
Only a quick first pass, would like to think about its soundness more carefully tmr. 

- should explain the math in the book (maybe add to Sean's corner on the bonding poly) 
- ~~what stop us from further merging preamble into the stage mask group?~~

I understand the completeness of this "grouping" as follows (which can be reworded and put into book):
`<<rx1, s1>> = 0`, where `s1 = s_global - notch_1`; now similarly for rx2 and s2. Now define a bundled stage mask: `s3 = s_global - notch_1 - notch_2`, and the claim is that `<<rx1, s3>> = 0`, because we can use linearity and decompose the LHS to `<<rx1, s1>> + <<rx1, -notch_2>>`, and the second term is also zero, because when stages are non-overlapping and the trace rx1 has all zeros on the region where `-notch_2` is non-zero. 


In terms of soundness, I need to convince myself that the following is impossible: 
in the first stage, i choose a bad rx1, then since I already know the sampled challenge, y and x, I will know what the stage mask will realize to on any position, (i.e. values of `x^g+m`), then i can carefully pick `rx2`, such that `<<rx1, s3>> = 0`, but `<<rx1, s1>> != 0`.

alxiong (changes requested):
revisiting my own comment in https://github.com/tachyon-zcash/ragu/pull/688#pullrequestreview-4189901815,

> I understand the completeness of this "grouping" as follows (which can be reworded and put into book):
<<rx1, s1>> = 0, where s1 = s_global - notch_1; now similarly for rx2 and s2. Now define a bundled stage mask: s3 = s_global - notch_1 - notch_2, and the claim is that <<rx1, s3>> = 0, because we can use linearity and decompose the LHS to <<rx1, s1>> + <<rx1, -notch_2>>, and the second term is also zero, because when stages are non-overlapping and the trace rx1 has all zeros on the region where -notch_2 is non-zero.

I think the logic in this PR is wrong.
imagine you i have a `rx1` that occupies a slot in `s2`'s unmasked range, then it would pass the `<<rx1, s3>> = 0` test, (Because it's still within the union range of `s1` and `s2`), but its original check is violated, i.e. `<<rx1, s1>> != 0` 

This is unsound. 
(unless i misinterpret the work here, I don't see any changes made to the `verifier.rs`, thus the verifier still effectively checking `<<rx1, s3>>=0 && <<rx2, s3>>=0` after the bundling of `s1` and `s2` masks into `s3`. )

@TalDerei 

<details>
<summary>Trivial test for the unsound case described above</summary>

```rust
    /// The bundled mask `s_global - notch_1 - notch_2` zeros out **both**
    /// stage windows, so a malformed `rx_1` carrying nonzero coefficients
    /// inside stage 2's notch slips through `revdot(rx_1, bundled_sy) = 0`.
    /// The original per-stage check `revdot(rx_1, stage1_sy) = 0` catches
    /// the same entry because `mask_1 = s_global - notch_1` is nonzero on
    /// stage 2's region.
    ///
    /// This demonstrates that the bundled bonding identity is only sound
    /// when an external invariant pins `rx_i` to zero on every other
    /// bundled stage's region. Today that invariant comes from the
    /// `(a, 0, 0, d)` standard allocator + the way `StageExt::rx` places
    /// witnesses; the bonding check itself does not enforce it.
    #[test]
    fn test_bundled_mask_misses_cross_stage_planted_rx() {
        // Stage 1: gates 1..100. Stage 2: gates 200..299. Disjoint.
        const STAGE1_SKIP: usize = 1;
        const STAGE1_NUM: usize = 100;
        const STAGE2_SKIP: usize = 200;
        const STAGE2_NUM: usize = 100;

        let stage1_mask = StageMask::<R>::new(STAGE1_SKIP, STAGE1_NUM).unwrap();
        let bundled_mask = StageMask::<R>::new(STAGE1_SKIP, STAGE1_NUM)
            .unwrap()
            .with_notch(STAGE2_SKIP, STAGE2_NUM)
            .unwrap();

        let y = Fp::random(&mut rand::rng());

        // Plant a nonzero entry on the a-wire of gate 250 — which lies
        // INSIDE stage 2's notch (200..300) and OUTSIDE stage 1's notch
        // (1..101).
        const PLANTED_GATE: usize = 250;
        assert!(PLANTED_GATE >= STAGE2_SKIP && PLANTED_GATE < STAGE2_SKIP + STAGE2_NUM);
        assert!(PLANTED_GATE < STAGE1_SKIP || PLANTED_GATE >= STAGE1_SKIP + STAGE1_NUM);

        let mut view = sparse::View::<Fp, R, _>::wiring();
        view.a.resize(PLANTED_GATE + 1, Fp::ZERO);
        view.a[PLANTED_GATE] = Fp::from(0xDEADBEEFu64);
        let malicious_rx_1 = view.build();

        // Bundled bonding check: mask is zero on stage 2's region, so the
        // planted entry contributes nothing. The malformed rx_1 passes.
        let mut bundled_sy = super::global_project::<Fp, R>(y);
        bundled_sy += &bundled_mask.notch_project(y);
        let bundled_check = malicious_rx_1.revdot(&bundled_sy);
        assert_eq!(
            bundled_check,
            Fp::ZERO,
            "soundness gap: bundled mask is zero on stage 2's region, so a planted \
             entry of rx_1 inside stage 2 is invisible to the bundled bonding check",
        );

        // Per-stage bonding check: stage 1's mask alone is *one* on stage 2's
        // region, so the same planted entry is visible and the check rejects.
        let mut stage1_sy = super::global_project::<Fp, R>(y);
        stage1_sy += &stage1_mask.notch_project(y);
        let stage1_check = malicious_rx_1.revdot(&stage1_sy);
        assert_ne!(
            stage1_check,
            Fp::ZERO,
            "the original per-stage check rejects the malformed rx_1 — bundling lost \
             a constraint that the separate-mask protocol enforced",
        );
    }
```
</details>

TalDerei:
> revisiting my own comment in [#688 (review)](https://github.com/tachyon-zcash/ragu/pull/688#pullrequestreview-4189901815),

as discussed OOB, intuited some soundness concerns with the impl per https://github.com/tachyon-zcash/ragu/pull/688#discussion_r3150023849. will review, but i'd like to find some remediation to keep these savings!

TalDerei:
> revisiting my own comment in [#688 (review)](https://github.com/tachyon-zcash/ragu/pull/688#pullrequestreview-4189901815)

here's the way i've been thinking about a potential fix here. bundling only applies on the mask-side (only collapses which masks get registered), not the rx-side. the rx's are still per stage fields in the proof, and the verifier can access them per-stage still. 

so here's the idea: keep the bundled stage masks / bonding claims, but the fix would be adding a per-stage support check on each rx in the final verifier. so for each stage, the verifier can iterate over the rx's coefficients and just assert each coefficient index lives in that stage's own window. that's basically negligible extra verifier work.

in that context, the bundled revdot catches plants / injections (per your example) outside it's union, but still misses the cross-stage injections inside the union (which is fine), and the support checks cover the later gap, and hence we recover the same well-formedness guarantees. 

pushed some commits to that effect, and not totally convinced it's the right approach but just offering a _shape_ of a solution.

TalDerei:
will set this as a draft awaiting @ebfull's review on the direction for these kinds of optimizations.

alxiong:
>  so for each stage, the verifier can iterate over the rx's coefficients and just assert each coefficient index lives in that stage's own window.

but the verifier can't perform these checks when moving to compressed proof where the verifier never receive `rx` but only `rx` commitment and can only query its openings.

TalDerei:
> > so for each stage, the verifier can iterate over the rx's coefficients and just assert each coefficient index lives in that stage's own window.
> 
> but the verifier can't perform these checks when moving to compressed proof where the verifier never receive `rx` but only `rx` commitment and can only query its openings.

yeah succinctness throws a wrinkle here...that's why this only a "shape" directionally of a solution.

TalDerei:
from all the analysis we've done, this does seem scary and premature, so closing for now and it's still tracked in https://github.com/tachyon-zcash/ragu/issues/185. thanks @alxiong!

PR #690, opened 2026-04-28
Improve formalization UX on the lean side

mitschabaude:
This PR improves the Lean-side formalization UX for Ragu instances and circuits.

- Formal instances are less boilerplate-heavy:
  - `Input`, `Output`, `inputLen`, and `outputLen` are inferred from `ProvableType.size`.
  - Instance `Spec` defaults to the reimplementation spec.
  - All explicit `Spec` and `same_spec` fields were removed.
  - Instance files were updated for newer Clean APIs.

- Ragu is updated to Clean main and its newer `GeneralFormalCircuit` shape:
  - verifier-side assumptions now live in `Assumptions`;
  - prover-only assumptions now live in `ProverAssumptions`;
  - the old `CompletenessSpec` role is now `ProverSpec`;
  - this allowed moving top-level spec premises into `Assumptions` while keeping prover-only witness conditions separate.

- Hint handling was refactored:
  - `FormalInstance` is the single formal-instance wrapper: it uses the hint-aware `GeneralFormalCircuit.WithHint` shape, and plain `GeneralFormalCircuit`s embed via `.toWithHint`.
  - `FormalInstance` input sizing uses `Value Input` size.
  - Circuits now use Clean prover hints / `UnconstrainedDep` inputs rather than separate hint callback arguments.
  - `Core.Mul`, allocators, invert circuits, and point allocation were ported to prover-hint style.
  - The old point `Hints.lean` glue was deleted.

- Several circuit APIs were aligned with the Rust circuits being formalized:
  - `DivNonzero` derives its witness internally.
  - `Invert` now has the Rust-like interface: input only, inverse witness derived internally.
  - `Boolean.And` now mirrors the Rust `dr.mul()` shape directly via `Core.mul`.
  - `Boolean.ConditionalEnforceEqual` had one prover assumption removed: its prover-side premise now matches the verifier spec, so it could become a `FormalAssertion`.
  - Low-level circuits got explicit `output` fields to reduce downstream proof noise.
  - Extra arithmetic instances for `Value` / `ProverValue` were added in `Ragu.Core`.

- Point circuit specs and assumptions were updated:
  - Top-level premises moved from specs to `Assumptions` where they are verifier-side assumptions.
  - `Double` had its prover assumptions weakened to match its verifier assumptions, so it could become a plain `FormalCircuit`.
  - `AddIncomplete`, `Double`, and `DoubleAndAddIncomplete` specs were rewritten to avoid awkward option matches where possible.
  - Touched circuit declarations moved toward `where` style.

- Core multiplication naming was aligned with Rust:
  - `Core.AllocMul` was renamed to `Core.Mul` / `Core.mul` on the Lean side.
  - The Rust Lean extractor target was renamed from `core_alloc_mul` / `CoreAllocMulInstance` to `core_mul` / `CoreMulInstance`.
  - The autogen module is now `Ragu.Instances.Autogen.Core.Mul`.

- Upstream docs and agent skills were merged and refreshed:
  - FV docs now describe `ProverAssumptions`, `ProverSpec`, and `GeneralFormalCircuit.WithHint` instead of the old `CompletenessSpec` model.
  - The full example now uses current `DivNonzero` + `Core.mul` style.
  - The `/fv-docs` and `/fv-review` skills were updated for formal-instance naming, `Unconstrained` / `UnconstrainedDep` hint inputs, and `Core.Mul`.


qa/lean/Ragu/Core.lean
```diff
@@ -41,39 +41,75 @@ def FlatOperation.eraseCompute {F : Type} [Field F] :
...
     = (exportedOperations input).toFlat.map FlatOperation.eraseCompute
 
-  same_output : ∀ (input : Vector (Expression (F p)) inputLen),
+  same_output : ∀ (input : Vector (Expression (F p)) (size (Value Input))),
     (input |> deserializeInput |> reimplementation |>.output 0 |> serializeOutput) = exportedOutput input
 
-  -- NOTE: this can be relaxed by proving that the reimplementation spec implies the instance spec instead
-  same_spec : ∀ input : Input (F p), ∀ output : Output (F p),
+  -- NOTE: this can be relaxed by proving that the reimplementation spec implies the instance spec instead.
+  same_spec : ∀ input : Value Input (F p), ∀ output : Output (F p),
```
TalDerei: question, any reason the formal instance would overwrite the spec and diverge from reimpl? not sure I understand that capability here?

qa/lean/Ragu/Circuits/Element/DivNonzero.lean
```diff
@@ -1,26 +1,27 @@
...
+  assertZero (y - denominator)
   return quotient
 
-def Assumptions (hintReader : ProverHint (F p) → Core.AllocMul.Row (F p))
-    (input : Inputs (F p)) (_data : ProverData (F p)) (hint : ProverHint (F p)) :=
-  let r := hintReader hint
-  r.y = input.y ∧ r.x * r.y = input.x ∧ (input.y ≠ 0 ∨ input.x = 0)
+def ProverAssumptions (input : Input (F p))
+    (_data : ProverData (F p)) (_hint : ProverHint (F p)) :=
+  input.y ≠ 0
```
TalDerei, line 24: meta observation; referencing https://github.com/tachyon-zcash/ragu/pull/642#discussion_r3103079289, i’ve kind of relaxed the notion of completeness being necessary since the ragu witness gen diverges from clean reimpl witness gen anyways. if we removed completeness and witness gen and marking as sorry, there's further simplification opportunities. but I think it's fine keeping anyway because afaict is still has some benefits (catches over-constrained gadgets at FV time and sanity checks the assumptions, which collectively seem useful).

qa/lean/docs/src/clean/core/example.md
```diff
@@ -8,92 +8,81 @@ Let's start with a basic template definition.
...
+  output _ offset := varFromOffset field offset
   localLength _ := 3
 
-theorem generalSoundness (idx : ℕ)
-    : GeneralFormalCircuit.Soundness (F p) (elaborated idx) GeneralSpec := by
-  circuit_proof_start [
-    Core.AllocMul.circuit, Core.AllocMul.Assumptions, Core.AllocMul.Spec, GeneralSpec
-  ]
+theorem soundness : GeneralFormalCircuit.Soundness (F p) elaborated Assumptions Spec := by
+  circuit_proof_start
```
TalDerei, line 47: ACK – `Assumptions` are now for both soundness and completeness (previously `GeneralFormalCircuit.Soundness` only took spec). prior to this change, `Assumptions` was for completeness only, but not it's seen by both soundness / completeness. and it seems like instead of the soundness preconditions being in the spec, we strip it out and put it in the assumptions and that lives as like a sibling to the spec.

TalDerei (approved):
simplification looks great! wasn't able to upstream to this PR, so filed some optional skill suggestions in https://github.com/tachyon-zcash/ragu/pull/692, feel free to cherry-picky.

qa/lean/Ragu/Core.lean
mitschabaude: Yeah I agree, these fields should be removed. There is no reason to diverge and in fact same_spec means you can't diverge

qa/lean/docs/src/clean/core/example.md
mitschabaude: No, Assumptions is seen only by soundness. For completeness, there are separate ProverAssumptions.

In GFC:

* soundness shows assumptions -> spec
* completeness shows prover assumptions -> prover spec

(Apart from the constraints) 

In FormalCircuit and FormalAssertion, these are collapsed:

* There is no prover spec
* In FC, prover assumptions = assumptions
* In FA, prover assumptions = (assumptions and spec)

qa/lean/Ragu/Core.lean
mitschabaude: removed in [a05b9c2](https://github.com/tachyon-zcash/ragu/pull/690/commits/a05b9c23339d31a1b2017a6191366eecab5988ac)

qa/lean/docs/src/clean/core/example.md
mitschabaude: I corrected this and a few other misunderstandings present in the skills here: [0942fdd](https://github.com/tachyon-zcash/ragu/pull/690/commits/0942fddf441d0074ccc45f776c5e99101e946348) (after merging your skill suggestions)

qa/lean/Ragu/Circuits/Element/DivNonzero.lean
mitschabaude: yeah IMO the completeness proofs are often either similar or easier than the soundness proofs, and they're nice to keep

mitschabaude:
@TalDerei thanks for the feedback!

- removed the unused formal instance specs
- Merged your skill updates, and had my own pass over them to correct a few things
- Pushed one more commit bringing a simplification due to another clean update: [3efc3f5](https://github.com/tachyon-zcash/ragu/pull/690/commits/3efc3f5d50214ef8fb0c577b4a44b7b5ca2572c9)

qa/lean/docs/src/clean/core/example.md
TalDerei: ah slight misunderstanding, this makes sense!

TalDerei:
@mitschabaude did another pass through the skills, thanks for correcting them :)

Issue #614, opened 2026-03-24
Determine utility of constraint counts method


TalDerei:
References https://github.com/tachyon-zcash/ragu/pull/610#discussion_r2977403874; `constraint_counts` and `segment_records` are ambiguous, determine their consumption.

PR #693, opened 2026-04-30
chore(circuits): remove ambiguous constraint_counts in favor of metrics

LucidSamuel:
Went with removal over redefinition. `constraint_counts` was only load-bearing for the PCD regression tests and metrics benchmarks. Both now use the feature-gated `ragu_circuits::testing::synthesis_counts()` helper, which runs the same synthesis analysis as registry compilation.

The returned raw counts have explicit semantics: they include the reserved SYSTEM gate, public-output constraints, and final ONE constraint. Normal builds do not enable `test-utils`.

Refs #614.

### DIFF
--- a/crates/ragu_circuits/Cargo.toml
+++ b/crates/ragu_circuits/Cargo.toml
@@ -35,6 +35,7 @@ modern-deps = [
 ]
 multicore = ["maybe-rayon/threads", "ragu_arithmetic/multicore", "std"]
 std = []
+test-utils = []
 
 [lib]
 bench = false
@@ -60,6 +61,7 @@ ragu_testing = { path = "../ragu_testing", version = "0.0.0" }
 [[bench]]
 name = "circuits"
 harness = false
+required-features = ["test-utils"]
 
 [[bench]]
 name = "trace_criterion"
--- a/crates/ragu_circuits/benches/circuits.rs
+++ b/crates/ragu_circuits/benches/circuits.rs
@@ -9,7 +9,7 @@ use ragu_arithmetic::Cycle;
 use ragu_circuits::{
     Circuit, CircuitExt,
     polynomials::{ProductionRank, TestRank, sparse},
-    registry::{CircuitIndex, Registry, RegistryBuilder},
+    registry::{Registry, RegistryBuilder},
 };
 use ragu_pasta::{Fp, Pasta};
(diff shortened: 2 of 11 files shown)


TalDerei (approved):
👍

TalDerei:
This seems to sufficiently addressing @ebfull's concerns in https://github.com/tachyon-zcash/ragu/pull/610#discussion_r2977403874. But on second pass it seems to just trade one public API for another (method on circuit ext rather than registry) to serve cross crate tests / benchmarks. Trying to think through a better approach that probably relies on some kind of flag gating.

Opted for testing helper file which is feature-gated with `testing-utils` and consumed cross-crate rather expanding public API surfaces. CI already uses `--all- features` so this isn't silently discarded in our test suite. so tldr,  ambiguous production `constraint_counts` API is replaced by an explicit, feature-gated testing helper. using my judgement here in cleaning up the review queue. cc @ebfull

LucidSamuel:
@TalDerei did a run with Fable for alternatives and it agreed that a feature-gated helper is the cleanest boundary. 

--
I’d make it narrower than the current CircuitExt::metrics() 
API: expose ragu_circuits::testing::synthesis_counts() under a non-default testing-utils feature, returning named gate and constraint counts. 

CircuitMetrics, segment_records(), and the extension method would remain private. The helper documentation would explicitly include the SYSTEM gate, public-output bindings, and final ONE constraint.

For the benchmarks, rather than enabling the testing feature, I’d replace the two direct metrics cases with size-parameterized registration cases.
`RegistryBuilder::register_circuit()` runs the same metrics analysis through the production path, and the existing register benchmark already establishes that pattern. That leaves the PCD regression test as the only consumer of testing-utils.

ragu_pcd can enable ragu_circuits/testing-utils through its dev-dependency, leaving normal builds unaffected. A user-facing capacity or preflight API may still be useful, especially for Hook budgets, but that should be designed separately around budget and headroom rather than shaped by #693 test requirements.

TalDerei:
> did a run with Fable for alternatives and it agreed that a feature-gated helper is the cleanest boundary.

see latest commit.

LucidSamuel:
small manifest hygiene point:- the `ragu_pcd` dev-dependency could also mirror the normal dependency with default-features.
= false, features = ["alloc", "test-utils"], rather than implicitly enabling the `ragu_circuits` default feature set

PR #699, opened 2026-05-06
feat: add merkle path verification gadget

LucidSamuel:
Book page under `guide/primitives/merkle.md` with cross-links from `getting_started.md `and `writing_circuits.md`.


TalDerei (commented):
https://github.com/tachyon-zcash/ragu/pull/325#pullrequestreview-3658543108 (cc @ebfull) feels on point; expanding the API surface like this feels premature until there's a demonstrable caller.

LucidSamuel:
> [#325 (review)](https://github.com/tachyon-zcash/ragu/pull/325#pullrequestreview-3658543108) (cc @ebfull) feels on point; expanding the API surface like this feels premature until there's a demonstrable caller.

okay, is the concern about `conditional_swap` specifically, or the merkle module, or both?

PR #700, opened 2026-05-06
simplify non-trivial example

S1nus:
Observing that multiple `Step`s can output the same type of `Header`, we can remove the separate types for `LeafNode` and `InternalNode`, leaving only `Node`.

Then, we can replace `WitnessLeaf` with `InitNode`, that outputs the same `Node` type as `Hash2`.


TalDerei (approved):

TalDerei:
this will be subsumed by macro changes, so opting to close this.

Issue #73, opened 2025-10-31
Partial witness support in the Circuit API


ebfull:
One of the most common hacks and optimizations in modern ZK proof systems is partial witness assignments / multi-stage witness assignments. Essentially, the prover commits to part of their circuit witness assignment and the proving system responds with a random challenge that is used to optimize checks done within the circuit, to reduce gate count.

This is part of what motivates the staging polynomial API. Inside of recursion, we want to compute Poseidon over a large number of inputs, but it would require us to invoke the permutation function many times. The solution is to commit to a partial witness (containing the inputs to the hash) and obtain a binding commitment (collision resistant, via Pedersen hashing) to those values (via public inputs) that can be used as inputs to Poseidon instead.

This is generally useful in application circuits; the application might have a curve point (representing a polynomial commitment) that the prover must commit to _before_ obtaining a challenge point that Ragu will provide oracle access for.

The same exact mechanism used to achieve this for the staging polynomials in recursion is what we should be using (or providing) for applications.

PR #701, opened 2026-05-07
feat(pcd): circuit API for polynomial query

S1nus:
Expose an API to circuit developers to perform optimized polynomial query verification, by delegating them to the proof system's existing batching infrastructure.

Adds a `PolyQueryClaims` object that may be constructed during `fuse(...)`, to be passed through `Adapter` to `Step::witness(...)` alongisde the `Driver`, allowing the Step's `witness(...)` function to call a `enforce_polynomial_query(com: Point, x: Element, y: Element)` function.


crates/ragu_pcd/src/poly_query.rs
TalDerei, line 1: organizationally, in terms of the `ragu_pcd` colocation here, tracing the dependency DAG suggests that nothing in this impl depends on anything in this crate. this is a generic PCS-level concept that feels like it should live in `ragu_primitives` instead, especially if something from `ragu_circuits` eventually needs to import it.

crates/ragu_pcd/src/fuse/_01_application.rs
```diff
@@ -41,7 +41,7 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             &mut *rng,
         )?;
 
-        let ((left_header, right_header), output_data, step_aux) = aux;
+        let ((left_header, right_header), output_data, step_aux, _claims) = aux;
```
TalDerei: the fact that there's no fuse-side consumer suggests this is an incomplete impl; if claims are discarded and aren't wired into fuse, isn't the test fixtures example demonstrably futile?

crates/ragu_pcd/src/step/mod.rs
```diff
@@ -171,9 +171,17 @@ pub trait Step<C: Cycle>: Sized + Send + Sync {
...
+    /// `pq` is a [`PolyQueryClaims`] sink that steps may use to record
+    /// polynomial-commitment opening claims via
+    /// [`PolyQueryClaims::enforce_polynomial_query`]. Steps that don't need
+    /// polynomial-query verification simply ignore the parameter. The
+    /// framework collects the resulting claims through the adapter's `Aux`
+    /// for later fuse-time processing.
     fn witness<'dr, 'source: 'dr, D: Driver<'dr, F = C::CircuitField>, const HEADER_SIZE: usize>(
         &self,
         dr: &mut D,
+        pq: &mut PolyQueryClaims<'dr, D, C::NestedCurve>,
```
TalDerei: this witness signature extension doesn't feel like the right shape to me. why should this be a trait obligation where every step impl (including ones that will never touch polynomial queries) has to thread `_pq` through its signature? there should be a way (and this is a very crude idea) to maybe bundle the drive object `dr` and `pq` into a single context object `cx` that represents like everything from the "outside" world the step is allowed to touch (ie. the driver, the poly query sink, etc).

crates/ragu_pcd/src/poly_query.rs
```diff
@@ -0,0 +1,73 @@
...
+        let triple =
+            D::try_just(|| Ok((com.value().take(), *x.value().take(), *y.value().take())))?;
+        let current = core::mem::replace(&mut self.claims, D::just(Vec::new));
+        self.claims = current.and_then(|mut v| {
+            triple.map(|t| {
+                v.push(t);
+                v
+            })
+        });
+        Ok(())
```
TalDerei: feel like there’s a more idiomatic rust pattern than `core::mem::replace` + `and_then/map`? if `PolyQueryClaims` internally held a `Vec<DriverValue<D, (C, D::F, D::F)>>` type, `enforce_polynomial_query` simplifies to a fold pattern.

crates/ragu_pcd/src/step/internal/adapter.rs
```diff
@@ -61,6 +61,7 @@ impl<C: Cycle, S: Step<C>, R: Rank, const HEADER_SIZE: usize> Circuit<C::Circuit
         ),
         <S::Output as Header<C::CircuitField>>::Data,
         S::Aux<'source>,
+        Vec<(C::NestedCurve, C::CircuitField, C::CircuitField)>,
```
TalDerei: intuitively, widening the aux type signature and needing to destructure callsites for that field feels off, similar to the witness observation. maybe we should make the `Aux<'source>` associated type a single named type, `AdapterAux`?

```ts
  pub(crate) struct AdapterAux<'source, C: Cycle, S: Step<C>, const HS: usize> {
      pub headers: (
          FixedVec<C::CircuitField, ConstLen<HS>>,
          FixedVec<C::CircuitField, ConstLen<HS>>,
      ),
      pub output_data: <S::Output as Header<C::CircuitField>>::Data,
      pub step_aux: S::Aux<'source>,
      pub poly_query_claims: Vec<(C::NestedCurve, C::CircuitField, C::CircuitField)>,
  }

  type Aux<'source> = AdapterAux<'source, C, S, HEADER_SIZE>;
```

TalDerei (changes requested):
meta observation in relation to broader partial witness / polynomial oracle access support described in https://github.com/tachyon-zcash/ragu/issues/73. In this PR, the application can basically assert some polynomial opening triple that the batched PCS machinery can enforce and amortize the costs for. In #73, there’s a capability in the other direction: the application provides some curve point (say a commitment), and fuse can derive the challenge x via FS and hand back x to the application as "oracle access" along with y = f(x), and the application can use x and y inside the circuit to reduce gate constraints (for instance, a smaller hash / check compared to doing something over the entire polynomial). In that context, maybe a more general symmetric impl pattern / design can emerge here, but that’s just a vague intuition. cc @ebfull

crates/ragu_pcd/src/fuse/_01_application.rs
S1nus: fuse-side consumer is WIP

S1nus:
> meta observation is in relation to broader partial witness / polynomial oracle access support described in #73. In this PR, the application can basically assert some polynomial opening triple that the batches PCS machinery can enforce and amortize the costs for. In #73, there’s a capability in the other direction: the application provides some curve point (say a commitment), and fuse can derive the challenge x via FS and hand back x to the application as oracle access along with y = f(x), and the application can use x and y inside the circuit to reduce gate constraints (for instance, a smaller hash / check compared to doing something over the entire polynomial). In that context, maybe a more general impl pattern can emerge here, but that’s just a vague intuition. cc @ebfull

in this case, the openings aren't random

crates/ragu_pcd/src/fuse/_01_application.rs
TalDerei: I think maybe they should be part of the same PR and land as a unit to demonstrate utility.

TalDerei:
> in this case, the openings aren't random

yeah that's the distinction I implied by saying "the application asserts some polynomial opening triple", but the broader symmetry note still stands though.

crates/ragu_pcd/src/step/mod.rs
S1nus: Good idea

turbocrime: conceptually, the driver is already responsible for context, yes? so even wrapping both in a broader 'context parameter' seems like it would be negative, imo

turbocrime (commented):
i think @TalDerei's commentary is good and i don't have much to add.

~~in general, i'd like to use a more gadget-focused api rather than a step-focused api~~

crates/ragu_pcd/src/fuse/_01_application.rs
S1nus: agree, just wanted review on API before wiring it in.

crates/ragu_pcd/src/poly_query.rs
S1nus: disagree because it's an API that only makes sense in the context of PCD circuit development

TalDerei:
warrant another review pass?

crates/ragu_pcd/src/poly_query.rs
S1nus: addressed this by changing `DriverValue<Vec` to `Vec<DriverValue`

TalDerei:
closes as superseded by https://github.com/tachyon-zcash/ragu/pull/783 (includes these commits).

PR #703, opened 2026-05-08
feat(pcd): circuit API for polynomial queries

S1nus:
Ragu will expose an API to circuit developers that enables optimized checking of PCS evaluations by performing the checks outside of the circuit, as part of the batched polynomial queries that occur during recursion

This PR adds a place in `Step::witness(...)` for implementors to return a `Vec` of polynomial evaluation claims that will be deferred to the proof system, and an example PCD tree that utilizes it.

a followup PR will integrate the deferred claims to `fuse(...)`


S1nus:
closing in favor of https://github.com/tachyon-zcash/ragu/pull/701

PR #704, opened 2026-05-13
perf: share child proof rx polynomials

LucidSamuel:
## Summary

Resolves the TODO at `proof/mod.rs:114` `Proof::as_child_stage_rx()` now clones `Arc` handles instead of deep-cloning 7 sparse polynomials per child (14 total per fuse call).

- `ChildStageRx` fields and bridge rx fields on `Proof`/`ProofBuilder` store `Arc<sparse::Polynomial<F, R>>` — `Clone` is a refcount bump
- `ProofBuilder` wraps polynomials in `Arc` at set time; bridge accessors return `&Arc<...>` so callers can `Arc::clone` for shared ownership
- Internal `Index` impls still return `&sparse::Polynomial` via `.as_ref()`
    — no API churn outside `ragu_pcd::proof`

also simplifies the trivial-proof self-child construction in `mod.rs:685`, which was deep-cloning the same  polynomials twice for left and right children.

### DIFF
--- a/crates/ragu_pcd/src/proof/builder.rs
+++ b/crates/ragu_pcd/src/proof/builder.rs
@@ -17,7 +17,7 @@ use ragu_circuits::{
 };
 use ragu_core::Result;
 
-use super::{Cached, Proof, SharedPolynomial};
+use super::{Cached, Proof};
 use crate::internal::nested;
 
 /// Produces `pub(crate) fn $name(&mut self, v: $ty)` that sets an `Option`
@@ -112,7 +112,7 @@ macro_rules! bridge_pair {
             commitment: C::NestedCurve,
         ) {
             assert!(self.$rx.is_none(), concat!("double-set: ", stringify!($rx)));
-            self.$rx = Some(rx);
+            self.$rx = Some(Arc::new(rx));
             self.$commitment = Some(commitment);
         }
 
@@ -163,7 +163,7 @@ macro_rules! explicit_commitment_getter {
     };
 }
 
-/// Produces `pub(crate) fn $rx(&self) -> Result<&sparse::Polynomial<...>>`
+/// Produces `pub(crate) fn $rx(&self) -> Result<&Arc<sparse::Polynomial<...>>>`
 /// and `pub(crate) fn $commitment(&self) -> Result<C::NestedCurve>` that
 /// lazily derive a cached bridge. On first call, `$rx` builds a
 /// `nested::stages::$stage::Witness` from the given getter methods, derives
@@ -177,9 +177,9 @@ macro_rules! explicit_commitment_getter {
 macro_rules! cached_bridge {
     ($rx:ident, $commitment:ident,
      $idx:expr, $stage:ident, { $($wit_field:ident : $getter:ident()),* }) => {
-        pub(crate) fn $rx(&self) -> Result<&sparse::Polynomial<C::ScalarField, R>> {
+        pub(crate) fn $rx(&self) -> Result<&Arc<sparse::Polynomial<C::ScalarField, R>>> {
             if let Some(rx) = self.$rx.get() {
-                return Ok(rx.as_ref());
+                return Ok(rx);
             }
             let rx = nested::stages::$stage::Stage::<C::HostCurve, R>::rx(
                 self.bridge_alpha_power($idx),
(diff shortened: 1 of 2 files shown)


alxiong:
Haven't reviewed, just bring to your attention our previous attempt in https://github.com/tachyon-zcash/ragu/pull/515 and the decision on choosing `Arc` among other possible smart pointers https://github.com/tachyon-zcash/ragu/issues/178#issuecomment-3973673762

the key thing is maintaining the "Copy/Clone-on-write" efficiency via `Arc::make_mut()`
(obviously the referenced PR is outdated, so maybe just the core idea and CoW behavior we need to ensure. could you first do an AI review pass to double check the CoW adherence whenever possible?)
@LucidSamuel

LucidSamuel:
> Haven't reviewed, just bring to your attention our previous attempt in #515 and the decision on choosing `Arc` among other possible smart pointers [#178 (comment)](https://github.com/tachyon-zcash/ragu/issues/178#issuecomment-3973673762)
> 
> the key thing is maintaining the "Copy/Clone-on-write" efficiency via `Arc::make_mut()` (obviously the referenced PR is outdated, so maybe just the core idea and CoW behavior we need to ensure. could you first do an AI review pass to double check the CoW adherence whenever possible?) @LucidSamuel

thanks for flagging. I reviewed #515 / #178 and did a focused AI review pass on the CoW behavior.

The #178 rule still applies: if an Arc-wrapped polynomial is ever mutated after sharing, that mutation should go through `Arc::make_mut()`.

I didn’t find any write-after-share path, the builder wraps owned polys once with `Arc::new`, `Proof::as_child_stage_rx()` only `Arc::clone``s handles, and consumers only get `&sparse::Polynomial` via `bridge_at()` / Index. I also checked for make_mut/get_mut/mutable shared-poly accessors and found none.

so `Arc::make_mut()` isn’t needed in this narrow child-handoff path. this just avoids the 7 deep clones per child / 14 per fuse call.

>also pushed a small cleanup commit for stale comments

crates/ragu_pcd/src/proof/mod.rs
```diff
@@ -43,9 +43,11 @@ use crate::{
 /// Wraps a value that can be recomputed from primary proof data. Used to
 /// distinguish commitment caches from primary polynomial fields at the type
 /// level. Immutable once constructed.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 struct Cached<T>(T);
 
+type SharedPolynomial<F, R> = Arc<sparse::Polynomial<F, R>>;
```
alxiong: I'd drop this type alias, just `Arc<T>` type is not complex enough to justify redirection.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -416,9 +416,48 @@ impl<'params, C: Cycle, R: Rank> ProofBuilder<'params, C, R> {
...
+    }
+
+    pub(crate) fn shared_nested_points_rx(&self) -> SharedPolynomial<C::ScalarField, R> {
+        Arc::clone(
+            self.nested_points_rx
+                .as_ref()
+                .expect("nested_points_rx not set"),
+        )
+    }
+
```
alxiong: imo, we shouldn't have these extra getters, just keep the old `ref_getter!` then just explicitly `Arc::clone` at the call site instead.

crates/ragu_pcd/src/proof/builder.rs
```diff
@@ -503,25 +542,49 @@ impl<'params, C: Cycle, R: Rank> ProofBuilder<'params, C, R> {
...
-        bridge_inner_error_commitment
-    );
     bridge_pair!(
         set_bridge_f_rx,
         bridge_f_commitment,
         bridge_f_rx,
         bridge_f_commitment
     );
 
+    pub(crate) fn set_bridge_s_prime_rx(
```
alxiong: don't really like these manual hand-expansion for setters, maybe make every bridge rx field `Option<SharedPolynomial<...>>` (or `OnceCell<SharedPolynomial<...>>` for the cached ones). That includes `bridge_preamble_rx` and `bridge_f_rx`, which the current PR leaves un-`Arc`'d just because they aren't shared into children.

Then we can update the macro `bridge_pair` for them uniformly.

alxiong (dismissed):

alxiong (commented):
ACK. (my previous comments are properly addressed)

TalDerei:
> * Introduced `SharedPolynomial<F, R>` type alias (`Arc<sparse::Polynomial<F, R>>`)

readme is stale, `SharedPolynomial` isn't in the diff.

TalDerei (approved):
looks good!

TalDerei:
keeping open, deferring something like this to @ebfull.

ebfull (approved):

PR #705, opened 2026-05-14
perf: benchmark registry eval restrictions

LucidSamuel:
Resolves the `_09_eval.rs` TODO by measuring the proposed registry-evaluation optimization instead of changing the fuse hot path speculatively.

The TODO suggested that registry evaluations in `_09_eval` could theoretically be more efficient if computed with registry
assistance. but those restriction polynomials already exist because they are committed earlier and used later for accumulation.

Measured locally:

- `registry::wxy_separate3`: ~5.16-5.34 ms
- `registry::restriction_eval3`: ~690-716 us

so changing `_09_eval` to recompute through the registry would regress performance rather than improve it.

### DIFF
--- a/crates/ragu_pcd/benches/criterion/registry.rs
+++ b/crates/ragu_pcd/benches/criterion/registry.rs
@@ -1,4 +1,4 @@
-use criterion::{Criterion, criterion_group, criterion_main};
+use criterion::{Criterion, black_box, criterion_group, criterion_main};
 use ff::Field;
 use ragu_arithmetic::Cycle;
 use ragu_circuits::polynomials::ProductionRank;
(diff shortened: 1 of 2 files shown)


alxiong:
Thanks for trying this out. I also observe the same regression while updating #595 
thus I drop those TODO comment in https://github.com/tachyon-zcash/ragu/pull/595/changes/744a269c09055f658c0cceee13dde5cedc6f0a0a

imo, we should close this since 595 already has the changes. (cc the PR owner @TalDerei)

TalDerei:
> imo, we should close this since 595 already has the changes. (cc the PR owner @TalDerei)

yeah duplicative against that PR

PR #706, opened 2026-05-14
refactor: share fuse query ordering

### LINKED ISSUE #603: compute_f and poly_queries duplicate query ordering manually
follow-up to #601, which added query-count guards for `compute_f` and `poly_queries()`.

those guards catch add/remove drift, but the static query prefix is still duplicated manually between
`_08_f.rs` and `compute_v.rs`. most of the remaining ordering is already shared through
`RxIndex::ALL` and `InternalCircuitIndex::ALL`; the static prefix is the last manually mirrored piece.

the long-term fix is to share the query specification itself, not just the count, using a common enum or descriptor list consumed by both sides..

-- pending #592 / #593 merge, since those changes are likely to reshape the query list.

LucidSamuel:
makes the static fuse query prefix a shared internal descriptor list instead of manually duplicating the order in two places.

this adds:
  - `StaticFQuery`
  - `STATIC_F_QUERIES`

both sides now consume that same static descriptor list:
- prover-side `compute_f` maps each `StaticFQuery` to its `factor_iter(...)`
- circuit-side `poly_queries` maps each `StaticFQuery` to `(p(u), v, denominator)`

the dynamic suffixes remain driven by the existing canonical lists:
  - `RxIndex::ALL`
  - `InternalCircuitIndex::ALL`
  
fixes #603

### DIFF
--- a/crates/ragu_pcd/src/fuse/_08_f.rs
+++ b/crates/ragu_pcd/src/fuse/_08_f.rs
@@ -116,51 +116,59 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
             idx.circuit_index().omega_j()
         };
 
-        let static_query_iter = |query| match query {
-            StaticFQuery::LeftP => factor_iter(left.native_p_poly().iter_coeffs(), left.u()),
-            StaticFQuery::RightP => factor_iter(right.native_p_poly().iter_coeffs(), right.u()),
-            StaticFQuery::LeftRegistryXyAtW => {
-                factor_iter(left.native_registry_xy_poly().iter_coeffs(), w)
-            }
-            StaticFQuery::RightRegistryXyAtW => {
-                factor_iter(right.native_registry_xy_poly().iter_coeffs(), w)
-            }
-            StaticFQuery::RegistryWx0AtLeftY => {
-                factor_iter(s_prime.registry_wx0_poly.iter_coeffs(), left.y())
-            }
-            StaticFQuery::RegistryWx1AtRightY => {
-                factor_iter(s_prime.registry_wx1_poly.iter_coeffs(), right.y())
-            }
-            StaticFQuery::RegistryWx0AtY => factor_iter(s_prime.registry_wx0_poly.iter_coeffs(), y),
-            StaticFQuery::RegistryWx1AtY => factor_iter(s_prime.registry_wx1_poly.iter_coeffs(), y),
-            StaticFQuery::RegistryWyAtLeftX => {
-                factor_iter(registry_wy.poly.iter_coeffs(), left.x())
-            }
-            StaticFQuery::RegistryWyAtRightX => {
-                factor_iter(registry_wy.poly.iter_coeffs(), right.x())
-            }
-            StaticFQuery::RegistryWyAtX => factor_iter(registry_wy.poly.iter_coeffs(), x),
-            StaticFQuery::RegistryXyAtW => {
-                factor_iter(builder.native_registry_xy_poly().iter_coeffs(), w)
-            }
-            StaticFQuery::RegistryXyAtLeftCircuitId => factor_iter(
-                builder.native_registry_xy_poly().iter_coeffs(),
-                left.circuit_id().omega_j(),
-            ),
-            StaticFQuery::RegistryXyAtRightCircuitId => factor_iter(
-                builder.native_registry_xy_poly().iter_coeffs(),
-                right.circuit_id().omega_j(),
-            ),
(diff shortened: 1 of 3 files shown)


crates/ragu_pcd/src/internal/native/mod.rs
```diff
@@ -420,3 +499,27 @@ pub fn register_all<'params, C: Cycle, R: Rank, const HEADER_SIZE: usize>(
 
     Ok(registry)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{InternalCircuitIndex, NUM_F_QUERIES, RxIndex, STATIC_F_QUERIES};
+
+    #[test]
+    fn f_query_count_matches_static_and_dynamic_sections() {
```
alxiong: unnecessary. statically and visually inspectable.

crates/ragu_pcd/src/fuse/_08_f.rs
```diff
@@ -115,41 +116,51 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
+            StaticFQuery::LeftAbAAtXz => factor_iter(left[RxComponent::AbA].iter_coeffs(), xz),
+            StaticFQuery::LeftAbBAtX => factor_iter(left[RxComponent::AbB].iter_coeffs(), x),
+            StaticFQuery::RightAbAAtXz => factor_iter(right[RxComponent::AbA].iter_coeffs(), xz),
+            StaticFQuery::RightAbBAtX => factor_iter(right[RxComponent::AbB].iter_coeffs(), x),
+            StaticFQuery::CurrentAAtXz => factor_iter(builder.native_a_poly().iter_coeffs(), xz),
+            StaticFQuery::CurrentBAtX => factor_iter(builder.native_b_poly().iter_coeffs(), x),
+        };
+
+        let mut iters: Vec<_> = Vec::with_capacity(native::NUM_F_QUERIES);
+        iters.extend(STATIC_F_QUERIES.into_iter().map(static_query_iter));
```
alxiong: nit: simply `STATIC_F_QUERIES.into_iter().map(|query| match query { ... }).collect();`, 

no need to declare `static_query_iter` first, then construct the `iters`, then extends. it

alxiong (commented):
ACK

crates/ragu_pcd/src/internal/native/mod.rs
LucidSamuel: makes sense. I added this as a lil guardrail while refactoring because the old query list was hand-maintained in two places, and I wanted something that would catch accidental drift in the static/dynamic split.

would it better to drop the whole test  module, or keep only the uniqueness check?

crates/ragu_pcd/src/internal/native/mod.rs
```diff
@@ -322,6 +322,81 @@ pub enum RxComponent {
...
 }
 
+/// Static prefix of the polynomial-query order used to construct and verify
+/// the fuse quotient polynomial $f(X)$.
+///
+/// These queries are consumed by both the prover-side `compute_f` path and the
+/// `compute_v` circuit. Keeping this prefix in one ordered list prevents the
+/// two implementations from drifting independently; the dynamic suffixes are
+/// still driven by [`RxIndex::ALL`] and [`InternalCircuitIndex::ALL`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
ebfull: Can these be removed? I doubt any of them are necessary.

crates/ragu_pcd/src/internal/native/mod.rs
```diff
@@ -322,6 +322,81 @@ pub enum RxComponent {
...
+pub(crate) enum StaticFQuery {
+    /// Left child proof $p(u)=v$ check.
+    LeftP,
+    /// Right child proof $p(u)=v$ check.
+    RightP,
+    /// Left child registry-xy polynomial queried at the current `w`.
+    LeftRegistryXyAtW,
+    /// Right child registry-xy polynomial queried at the current `w`.
+    RightRegistryXyAtW,
+    /// Current `m(w, x_0, Y)` queried at the left child's `y`.
```
ebfull: These should use math formula notation since they're in doc comments.

ebfull (changes requested):
Good refactoring! Just some nits.

TalDerei (approved):
ACK structural refactoring pending requested changes 👍

crates/ragu_pcd/src/internal/native/mod.rs
alxiong: i feel like only the uniqueness is sufficient. but not super strongly opinionated here. also fine with dropping the entire test module, since the enum variants by definition are unique.

TalDerei:
confirmed semantic equivalence (refactor is behaviorally no-op). merging since @ebfull's nits have been addressed.

PR #710, opened 2026-05-18
chore(fv): make lean reimpls polymorphic

### LINKED ISSUE #686: Make Lean reimpls polymorphic
Referencing https://github.com/tachyon-zcash/ragu/pull/672#discussion_r3138786691 from @mitschabaude, Clean reimpls _already_ support polymorphism (`(k : N)`, `Circuit.foldl`). The extractor, however, produces a concrete Lean constraint trace, and that extracted artifact is _monomorphic_ by construction (fixed shape), not polymorphic. We mistakenly let that constraint flow back into the Lean model for some gadgets. 

We have two Clean reimpls in #672 (`EnforceRootOfUnity` / `Fold`) that hit this extractor limitation: the rust gadgets are generic over some k, but the reimpls hard-code a length so their soundness/completeness theorems only cover that shape.

Anyways, for simplicity, we're checking (1) Reimpl === Spec (polymorphic Lean reimpl that works for all k), but (2) Rust's emitted trace (autogen) === Reimpl is only per k, and that's exactly the gap. So we'll have one polymorphic reimpl, but a few monomorphic autogen and formal instance impls.
                                                                                           
To fix this, we'd need to:
- make those reimpl + theorem polymorphic — parameterize main on k, prove soundness/completeness universally.
- add thin extraction instances that apply the polymorphic reimpl at concrete values, and include a few of these examples for high assurance.

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/686. For some gadget impls like `EnforceRootOfUnity` / `Fold`, we'll need a polymorphic reimpl plus a few monomorphic autogen and formal instance impls (for assurance, since this is an extractor limitation). Recall our IRL discussion during the conference on this, cc @gio54321 @alxiong (@mitschabaude for visibility).

### DIFF
--- a/qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
+++ b/qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
@@ -14,17 +14,16 @@ Modeled as a `FormalAssertion`: the gadget is constraint-only (no
 output), `Spec` acts as both assumption for completeness and conclusion
 for soundness.
 
-The body iteratively squares the input via `squareIter`, matching the
-Rust `for _ in 0..k { value = value.square(dr)?; }` loop. -/
-def squareIter : (k : ℕ) → Expression (F p) → Circuit (F p) (Expression (F p))
-  | 0, x => pure x
-  | k+1, x => do
-    let sq ← Mul.circuit ⟨x, x⟩
-    squareIter k sq
-
-def main (k : ℕ) (input : Expression (F p)) : Circuit (F p) (Var unit (F p)) := do
-  let final ← squareIter k input
-  assertZero (final - 1)
+The `k+1` branch uses `Circuit.foldl (.finRange (k+1))` so that
+`ConstantOutput` (the body's output is a fresh wire, independent of
+`acc`) is auto-discharged and `foldl.output_eq` / `foldl.soundness` /
+`foldl.completeness` fire under `circuit_norm`. The `k = 0` branch is
+handled explicitly because `Inhabited (Fin 0)` doesn't exist. -/
+def main : (k : ℕ) → Expression (F p) → Circuit (F p) (Var unit (F p))
+  | 0, input => assertZero (input - 1)
+  | k + 1, input => do
+    let final ← Circuit.foldl (.finRange (k + 1)) input fun x _ => Mul.circuit ⟨x, x⟩
+    assertZero (final - 1)
 
 def Assumptions (_ : F p) := True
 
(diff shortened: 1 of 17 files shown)


qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
```diff
@@ -1,52 +1,154 @@
...
+Rust `for _ in 0..k { value = value.square(dr)?; }` loop. -/
+def squareIter : (k : ℕ) → Expression (F p) → Circuit (F p) (Expression (F p))
+  | 0, x => pure x
+  | k+1, x => do
+    let sq ← Mul.circuit ⟨x, x⟩
+    squareIter k sq
+
+def main (k : ℕ) (input : Expression (F p)) : Circuit (F p) (Var unit (F p)) := do
+  let final ← squareIter k input
+  assertZero (final - 1)
```
mitschabaude: it's cute you got a recursive circuit to work with induction proofs! I think it should be easier with our built-in `Circuit.foldl` loop:
```suggestion
def main (k : ℕ) (input : Expression (F p)) : Circuit (F p) Unit := do
  let final ← Circuit.foldl (.finRange k) input fun x _ => Mul.circuit ⟨x, x⟩
  assertZero (final - 1)
```

qa/lean/Ragu/Circuits/Element/Fold.lean
```diff
@@ -4,49 +4,193 @@ import Ragu.Circuits.Element.Mul
...
+      Circuit (F p) (Expression (F p))
+  | 0, acc, _, _ => pure acc
+  | m+1, acc, rest, s => do
+    let scaled ← Mul.circuit ⟨acc, s⟩
+    hornerStep m (scaled + rest[0]) rest.tail s
 
-def Assumptions (_input : Input (F p)) := True
+def main : (n : ℕ) → Var (Input n) (F p) → Circuit (F p) (Expression (F p))
+  | 0, _ => pure 0
+  | _+1, input => hornerStep _ input.xs[0] input.xs.tail input.s
```
mitschabaude: same comment as above, I think using `Circuit.foldl` would make this easier

mitschabaude (commented):
clean has a couple of loop constructs with simp support in `circuit_proof_start` / `circuit_norm`. We use these whenever we need a loop:
`Circuit.forEach`, `Circuit.map`, `Circuit.mapFinRange`, `Circuit.foldl`, `Circuit.foldlRange`.

- `Circuit.foldlRange` has the worst simp behavior of these but is also the most general, it can emulate any `for i = ...` loop.
- `foldl` assumes the output of the loop body, as an expression, does not depend on its input. this is usually the case if your circuit creates fresh variables for the output like `Mul.circuit` does. It should simplify fairly nicely and should be usable for most of your circuits.

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
TalDerei: yeah thought this was nice :) converting to built-ins doesn't cleanly translate, PR'd https://github.com/Verified-zkEVM/clean/pull/373 to investigate.

TalDerei:
@mitschabaude super helpful feedback, wasn't aware 

> clean has a couple of loop constructs with simp support in `circuit_proof_start` / `circuit_norm`. We use these whenever we need a loop:

should have looked at the authoritative docs (rather than our own which didn't mention them), these are really helpful!

qa/lean/Ragu/Circuits/Element/EnforceRootOfUnity.lean
TalDerei: finally got this to work using `Circuit.foldl (.finRange k)`!

ebfull (approved):

PR #711, opened 2026-05-18
feat(fv): endoscalar gadget coverage

TalDerei:
Builds on https://github.com/tachyon-zcash/ragu/pull/710 and extends coverage to endoscalar gadget (`Alloc, Lift, GroupScale, Extract`). Currently, `Alloc, Lift, and GroupScale` are implemented, and `Extract` can't be verified until it's separately reworked. Also more carefully audited the specs here.

Most of the diff is dominated by the ridiculously long autogen from the codegen step, but this generated code will be replaced in a subsequent pass in https://github.com/tachyon-zcash/ragu/issues/786. **Update:** This is now removed after rebasing over https://github.com/tachyon-zcash/ragu/pull/795.


qa/lean/Ragu/Circuits/Endoscalar/Alloc.lean
```diff
@@ -0,0 +1,67 @@
...
+the corresponding bit of the 128-bit prover hint `value`. Mirrors the
+Rust loop `for i in 0..Uendo::BITS { Boolean::alloc((value >> i) & 1) }`
+in `crates/ragu_primitives/src/endoscalar.rs::Endoscalar::alloc`.
+
+`Uendo` is `pub use u128 as Uendo` in `crates/ragu_arithmetic/src/lib.rs:102`,
+so `Uendo::BITS = u128::BITS = 128`. This Lean reimpl is monomorphic at 128 —
+no polymorphism needed.
+
+TODO: extraction instance deferred. The 512-operation autogen (128 Booleans ×
+4 ops each) exceeds what Clean's current simp/tactic infrastructure can reduce
```
TalDerei: this blows up the constraint count or something on the autogen-side extraction instance / autogen / reimpl, so deferring for now.

crates/ragu_primitives/src/fv_utils.rs
TalDerei, line 1: motivated by https://github.com/tachyon-zcash/ragu/issues/799, and proposing we extend this to other higher level gadgets if we want true parity as gadgets change and become more optimized.

TalDerei:
merging this since it's isolated FV work, and will follow-up with extract impl once endoscalar primitive in https://github.com/tachyon-zcash/ragu/pull/767 lands. https://github.com/tachyon-zcash/ragu/issues/799 is follow-up work to actually use ragu defs rather than sometimes duplicating them in the `qa/` to avoid needing to modify the public API surface or peppering `unstable-fv` flags. cc @ebfull

PR #713, opened 2026-05-18
feat(pcd): poly query API with `Context` object

S1nus:
alternative to https://github.com/tachyon-zcash/ragu/pull/701.

In this variant, `dr` and `pq` are wrapped into one object called `Context`, that may be instantiated with or without `pq`.


S1nus:
closing, https://github.com/tachyon-zcash/ragu/pull/701 adds this functionaltiy.

PR #715, opened 2026-05-19
Cleanups

ebfull:


crates/ragu_circuits/src/staging/bonding.rs
```diff
@@ -75,6 +81,15 @@ where
...
         }
 
+        // Defensive cross-check on the type-level invariant: `Output = ()` and
+        // `Write<F> for ()` together guarantee no output-binding constraints
+        // are emitted, so the instance polynomial $k(Y)$ has degree zero.
+        let metrics = crate::metrics::eval(&self)?;
+        assert_eq!(
+            metrics.degree_ky, 0,
+            "bonding circuit must have empty instance polynomial",
+        );
```
TalDerei, line 91: seems like a tautological assertion (i've traced the logic and can't see this ever triggering, unless under some kind of trait relaxation, since there are structural type invariants that prevent it and make it no-op), so not sure it warrants a full extra circuit synthesis pass here.

crates/ragu_circuits/src/metrics.rs
```diff
@@ -194,8 +194,6 @@ pub struct CircuitMetrics {
     pub(crate) num_gates: usize,
 
     /// The degree of the instance polynomial $k(Y)$.
-    // TODO(ebfull): not sure if we'll need this later
-    #[allow(dead_code)]
     pub(crate) degree_ky: usize,
```
TalDerei, line 199: vaguely recalling this field coming from a downstream suggestion after looking at what other arkworks PCD libraries do, and they threaded the public instances degree in their version of metrics. doesn't actually seem necessary, we can probably just remove this field honestly.

crates/ragu_primitives/src/consistent.rs
```diff
@@ -1,25 +1,26 @@
...
 
-/// Trait that enforces a gadget's internal constraints on existing wires.
+/// Implemented for gadgets that are capable of emitting constraints to
+/// re-express all of their invariants on the wires inside `self`.
 ///
-/// Some gadgets require internal invariants for correctness; a [`Point`] must
-/// satisfy its curve equation, a [`Boolean`] must be 0 or 1. This trait enforces
-/// those constraints on wires allocated elsewhere, separating allocation from
-/// constraint enforcement.
+/// Not every gadget can implement this.
```
TalDerei: nit; this wording seems vague.

book/src/implementation/staging.md
```diff
@@ -1,6 +1,145 @@
 # Staging Implementation
 
-> **TODO**: This section will explain the implementation of staged traces,
-> including how the staging mechanism from the
-> [protocol design](../protocol/extensions/staging.md) is realized in the
-> `ragu_pcd` crate.
+The [protocol chapter on staging](../protocol/extensions/staging.md) defines
+the math: a trace polynomial $r(X)$ is decomposed into stage polynomials
+$a(X), b(X), \ldots$ plus a final trace $r'(X)$, each independently
+committable and each guaranteed well-formed by a [`StageMask`][stagemask].
```
TalDerei:
```suggestion
committable and each guaranteed well-formed by a [stage mask][stagemask].
```

book/src/guide/gadgets/index.md
```diff
@@ -40,39 +45,35 @@ This gadget is a _compositional_ gadget: it contains another gadget (a
...
-  the definition of [`Maybe<T: Send>`][maybe-trait]), this property almost
-  always holds.
-* **They must be `'static`.** Specifically, when the driver's lifetime `'dr` is
-  the static lifetime `'static` the gadget itself must be `'static`. This
-  property is guaranteed by the Rust type system, and so gadget implementations
-  do not need to carefully reason about it. In practice, any references a gadget
-  contains must be `'static`.
-* **They must be `Clone`.** All gadgets should be cloneable. This is commonly
-  necessary anyway, but drivers may need to clone gadgets generically when
-  performing various transformations.
```
TalDerei, line 75: this API contract still holds if we wanted to keep this.

TalDerei (approved):
some minor observations

crates/ragu_circuits/src/staging/bonding.rs
ebfull: It doesn't, Claude's code review skill flagged it for me also. I thought I'd keep it for now since this metrics pass is only called when bonding circuits are turned into objects, which should occur rarely and only during startup.

ebfull: (Bonding _circuits_ are not actually materialized during proving.)

crates/ragu_circuits/src/metrics.rs
ebfull: I deliberately kept the assertion above purely to remove this `dead_code` so that we could keep this.

book/src/guide/gadgets/index.md
ebfull: I couldn't find a pretty location to put it when I was rewriting the book entry, I think it's so minor that because it's a supertrait we could leave it to the Rust docs.

Issue #707, opened 2026-05-17
Rebalance internal circuits to support the largest header size


TalDerei:
Related to https://github.com/tachyon-zcash/ragu/issues/53.

Our test bench encodes `HEADER_SIZE = 65`, which is essentially the cap that our internal recursion circuits can currently support (scales by some number of muls / increase in H). If an application circuit's budget exceeds `HEADER_SIZE - 1`, it'll throw an encoding error. We should be rebalancing our circuits to support the largest possible header size, eg. migrating the ky computation from `Hashes1` to `OuterCollapseCircuit` should raise the limit (back-of-envelope: cap moves to `HEADER_SIZE = 105`).

TalDerei:
closed by #712.

PR #720, opened 2026-05-21
perf: step capacity and proving-cost investigation

turbocrime:
Test harnesses that empirically map a Ragu PCD step's **budget** (how much fits) and **proving cost** (what moves the clock) under `ProductionRank` (`R<13>`, 2048-gate cap). Each file is one experiment with one knob; they print tables rather than assert, so run with `--nocapture`. Cargo runs the test binaries serially, so one command is fine; `--test-threads=1` serializes execution. Timing numbers are release + `multicore` on my machine.

Run all of them:

```sh
cargo test --features multicore -p ragu_pcd --release \
  --test distinct_headers \
  --test distinct_steps \
  --test endoscalar_capacity \
  --test num_steps \
  --test num_witnesses \
  --test poseidon_capacity \
  --test poseidon_squeeze_capacity \
  --test uniform_step_cost \
  -- --nocapture --test-threads=1
```

### Poseidon permutations

One step performs K chained Poseidon permutations; K rises until it busts the gate cap.

```sh
cargo test --features multicore -p ragu_pcd --release --test poseidon_capacity -- --nocapture --test-threads=1
```

|  K | gates |
| -: | ----: |
|  1 |   290 |
|  2 |   578 |
|  3 |   866 |
|  4 |  1154 |
|  5 |  1442 |
|  6 |  1730 |
|  7 |  2018 |
|  8 |  bust |

Constant **+288 gates per permutation** → the 2048-gate budget holds ~7 Poseidons.

### Endoscalar ops

Same, with K chained endoscalar operations.

```sh
cargo test --features multicore -p ragu_pcd --release --test endoscalar_capacity -- --nocapture --test-threads=1
```

|  K | gates |
| -: | ----: |
|  1 |   588 |
|  2 |  1043 |
|  3 |  1498 |
|  4 |  1953 |
|  5 |  bust |

Constant **+455 gates per op** → ~4 endoscalars per step. Budget a step in these units; if you're over, split the work across steps (the cost section shows why splitting is nearly free).

### Repeated squeezes

One step squeezes K field elements from a single sponge; reports gates as K grows.

```sh
cargo test --features multicore -p ragu_pcd --release --test poseidon_squeeze_capacity -- --nocapture --test-threads=1
```

|  K | gates |
| -: | ----: |
|  1 |   290 |
|  4 |   290 |
|  5 |   578 |
|  8 |   578 |
|  9 |   866 |
| 16 |  1154 |
| 28 |  2018 |
| 29 |  bust |

Squeezing is near-free within a permutation's rate: gates hold flat across 4 squeezes and only step up when a fresh permutation is forced. Pulling many outputs from one sponge costs far less than the same number of permutations.

## Cost — what actually moves per-proof time

The recurring result: within a fixed application, per-proof cost is dominated by the fixed recursion pipeline, and almost nothing done *inside* a step changes it.

### Per-step gate work

A fixed 5-step application (two leaf seeds, two sibling fuses, a final fuse); one step's Poseidon count is swept while the others stay minimal, timing every step. Below: a swept fuse and a swept leaf (seed), each vs its own gate count.

```sh
cargo test --features multicore -p ragu_pcd --release --test uniform_step_cost -- --nocapture --test-threads=1
```

| gates | fuse avg(ms) | seed avg(ms) |
| ----: | -----------: | -----------: |
|     2 |        197.8 |        352.1 |
|   290 |        199.8 |        361.3 |
|   866 |        203.8 |        353.6 |
|  1442 |        204.1 |        355.7 |
|  2018 |        210.8 |        359.9 |

Filling a step from 2 → 2018 gates (a 1000× swing) moves its time ~7% (fuse) / ~2% (seed). Gate count is a *capacity* constraint, not a *speed* one — no runtime reason to shave gates until near the cap. A `seed` costs ~1.75× a `fuse` (the base-case path); the full output also shows operand order and leaf-vs-internal inputs make no difference.

### Witness (wire) count

A depth-2 tree (4 leaves → 2 fuses → 1 root) where each leaf allocates N witness wires and does no hashing; times the whole tree as N grows.

```sh
cargo test --features multicore -p ragu_pcd --release --test num_witnesses -- --nocapture --test-threads=1
```

| witnesses | gates | avg(ms) |
| --------: | ----: | ------: |
|         1 |     2 |  1986.9 |
|       512 |   257 |  1981.9 |
|      1024 |   513 |  2007.1 |
|      2048 |  1025 |  2025.5 |
|      4092 |  2047 |  2047.3 |

Allocating many witness wires is ~free per proof — holding large amounts of witness data in a step is cheap. (Incidental limits: the wire ceiling is ~4092 allocations, since `gates ≈ N/2 + 1` against a strict `< 2048` bound, and pure allocations add no constraints — `constr` stays 13.)

### Tree node count

Balanced binary trees of increasing depth, built from a fixed 3-step reuse set; times the whole tree against node count.

```sh
cargo test --features multicore -p ragu_pcd --release --test num_steps -- --nocapture --test-threads=1
```

| depth | nodes | avg(ms) | per-node |
| ----: | ----: | ------: | -------: |
|     1 |     3 |   887.2 |    295.7 |
|     2 |     7 |  1969.8 |    281.4 |
|     3 |    15 |  4139.4 |    276.0 |
|     4 |    31 |  8541.6 |    275.5 |

The one knob that scales cost: total proving time is **linear** in the number of executed nodes, at a ~constant ~276 ms/node. Cost tracks the size of the computation (how many folds you run), exactly as constant-size PCD proofs predict.

### Distinct step types vs distinct headers

A left-leaning chain folds a fresh seed into the accumulator at each of N levels, so every registered step is actually executed. `distinct_steps` uses N distinct step types that share one header; `distinct_headers` threads a distinct header through each level. Node count is `2N+1`, so the per-node column isolates the marginal cost of distinctness.

```sh
cargo test --features multicore -p ragu_pcd --release --test distinct_steps -- --nocapture --test-threads=1
```

| steps | nodes | circuits | log2 | avg(ms) | per-node |
| ----: | ----: | -------: | ---: | ------: | -------: |
|     1 |     3 |       18 |    5 |   887.6 |    295.9 |
|     4 |     9 |       21 |    5 |  2542.1 |    282.5 |
|     8 |    17 |       25 |    5 |  4786.3 |    281.5 |
|    16 |    33 |       33 |    6 |  9323.6 |    282.5 |

```sh
cargo test --features multicore -p ragu_pcd --release --test distinct_headers -- --nocapture --test-threads=1
```

| levels | nodes | circuits | log2 | avg(ms) | per-node |
| -----: | ----: | -------: | ---: | ------: | -------: |
|      1 |     3 |       17 |    5 |   976.9 |    325.6 |
|      4 |     9 |       20 |    5 |  2794.8 |    310.5 |
|      8 |    17 |       24 |    5 |  4835.6 |    284.4 |
|     16 |    33 |       32 |    5 |  9294.4 |    281.6 |

Both converge to ~282 ms/node — the same as the reused-type tree above (`num_steps`, ~276). Executing distinct step types, or threading a distinct header per level, costs nothing per node beyond a plain node (the elevated per-node at N=1 is just the seed-heavy small chain). Application breadth — many distinct registered step types and headers, and crossing the `log2_circuits` boundary — does not tax proving; only the executed node count does.

## Net

Proving cost scales only with the number of **executed** steps (tree size), linearly. Everything structural inside or around a step — gate work, witness count, step-type count, header count — is essentially free per proof, swamped by the fixed recursion pipeline. Design for the gate *capacity* cap, and split work across more steps freely; it costs nothing per proof beyond the extra nodes themselves.

### DIFF
--- a/crates/ragu_pcd/tests/endoscalar_capacity.rs
+++ b/crates/ragu_pcd/tests/endoscalar_capacity.rs
@@ -1,0 +1,122 @@
+//! How many endoscalar `group_scale` calls fit in one application step
+
+use ff::Field;
+use ragu_circuits::{polynomials::ProductionRank, registry::CircuitIndex};
+use ragu_core::{
+    Result,
+    drivers::{Driver, DriverValue},
+    gadgets::{Bound, Kind},
+    maybe::Maybe,
+};
+use ragu_pasta::{EpAffine, Fp, Pasta};
+use ragu_pcd::{
+    ApplicationBuilder,
+    header::{Header, Suffix},
+    step::{Encoded, Index, Step},
+};
+use ragu_primitives::{
+    Element, Endoscalar, Point,
+    allocator::{Allocator, Standard},
+};
+
+const HEADER_SIZE: usize = 2;
+
+/// 1-element input header, sized to fit a single `Fp` plus the suffix slot.
+struct InputHeader;
+
+impl<F: Field> Header<F> for InputHeader {
+    const SUFFIX: Suffix = Suffix::new(0);
+    type Data = F;
+    type Output = Kind![F; Element<'_, _>];
+
+    fn encode<'dr, D: Driver<'dr, F = F>, A: Allocator<'dr, D>>(
+        dr: &mut D,
+        allocator: &mut A,
+        witness: DriverValue<D, Self::Data>,
+    ) -> Result<Bound<'dr, D, Self::Output>> {
+        Element::alloc(dr, allocator, witness)
+    }
+}
+
+/// Step performing exactly K endoscalar `group_scale` calls chained against
+/// a running point accumulator. Each iteration scales the accumulator by the
+/// same allocated scalar, so the total work is K independent scalings.
+struct ScaleK<const K: usize>;
+
+impl<const K: usize> Step<Pasta> for ScaleK<K> {
+    const INDEX: Index = Index::new(0);
+    type Witness<'source> = (EpAffine, u128);
+    type Aux<'source> = ();
+    type Left = InputHeader;
+    type Right = ();
+    type Output = InputHeader;
+
+    fn witness<'dr, 'source: 'dr, D: Driver<'dr, F = Fp>, const HS: usize>(
+        &self,
+        dr: &mut D,
+        witness: DriverValue<D, (EpAffine, u128)>,
+        left: DriverValue<D, Fp>,
+        _right: DriverValue<D, ()>,
+    ) -> Result<(
+        (
+            Encoded<'dr, D, Self::Left, HS>,
+            Encoded<'dr, D, Self::Right, HS>,
+            Encoded<'dr, D, Self::Output, HS>,
+        ),
+        DriverValue<D, <Self::Output as Header<Fp>>::Data>,
+        DriverValue<D, Self::Aux<'source>>,
+    )>
+    where
+        Self: 'dr,
+    {
+        let allocator = &mut Standard::new();
+        let left = Encoded::new(dr, allocator, left)?;
+        let right = Encoded::from_gadget(());
+
+        let (p, r) = witness.cast();
+        let mut acc = Point::alloc(dr, p)?;
+        let r = Endoscalar::alloc(dr, r)?;
+        for _ in 0..K {
+            acc = r.group_scale(dr, &acc)?;
+        }
+
+        let l: &Element<'dr, D> = left.as_gadget();
+        let output_gadget: Element<'dr, D> = l.clone();
+        let output_data = output_gadget.value().map(|v| *v);
+        let output = Encoded::from_gadget(output_gadget);
+        Ok(((left, right, output), output_data, D::unit()))
+    }
+}
+
+fn measure<const K: usize>() -> Result<(usize, usize)> {
+    let pasta = Pasta::baked();
+    let app = ApplicationBuilder::<Pasta, ProductionRank, HEADER_SIZE>::new()
+        .register(ScaleK::<K>)?
+        .finalize(pasta)?;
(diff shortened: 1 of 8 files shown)


TalDerei (approved):

crates/ragu_pcd/tests/endoscalar_capacity.rs
```diff
@@ -0,0 +1,123 @@
...
+    ApplicationBuilder,
+    header::{Header, Suffix},
+    step::{Encoded, Index, Step},
+};
+use ragu_primitives::{
+    Element, Endoscalar, Point,
+    allocator::{Allocator, Standard},
+};
+
+const HEADER_SIZE: usize = 2;
```
TalDerei, line 22: depends on the header size obviously, which scales constraints (by some number of muls / increase in H) per https://github.com/tachyon-zcash/ragu/issues/707. maybe we should pin this to a more reasonable H.

turbocrime: effect seems to be quite small. i had some tests scaling the header as an axis but it would require quite a large header to move the bust boundary, so i deleted them.

turbocrime: practically, header gate effect appears negligible for the sizes i'm concerned with, so i'm not motivated to explore deeply.

this was written to investigate the gadget cost bounds, and it serves.

TalDerei:
we're trying to minimize review queue pressure, and this doesn't strictly need to land. now that the numbers are calculated, we can extract the metrics into the book (tracked in #824). when we're at the stage where we're doing more benchmarking and fine tuning performance, we can PR some version of this.

PR #721, opened 2026-05-21
Bump zkcrypto patches

ebfull:
* This uses the upstream `ff`'s own `0.14.0` release branch now that `addchain` was released and that branch has been updated.
* Removes the `addchain` patch; don't need it now.
* `group`'s own `0.14.0` release branch is also aligned with these changes now, except that there are some additional trait changes in that branch that are left out here.
* `pasta_curves` is just simply bumped to use (and account for) the `group` update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,8 +4,9 @@ version = 4
 
 [[package]]
 name = "addchain"
-version = "0.2.0"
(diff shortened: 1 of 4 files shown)


TalDerei (approved):
ACK on _partial_ dep upstream to [zkcrypto](https://github.com/zkcrypto). will later move off the remaining forks to upstream `zkcrypto/group` / `zcash/pasta_curves`.

PR #732, opened 2026-05-24
chore(ci): cargo vet workflow

TalDerei:
stacks on https://github.com/tachyon-zcash/ragu/pull/731/. 

`cargo vet` expects its config directory to live in workspace root, so had to create an ephemeral symlink inside the CI job that points to `qa/supply-chain/`. also configures `Supply chain / cargo vet` status checks in branch protections so any version drift blocks merge. 

**update:** https://mozilla.github.io/cargo-vet/commands.html#--store-path-store_path for path override instead of using a symlink.

### DIFF
--- a/.github/workflows/vet.yml
+++ b/.github/workflows/vet.yml
@@ -16,37 +16,10 @@ permissions:
   contents: read
 
 jobs:
-  # Detect whether anything that could affect vet/audit changed. The downstream
-  # jobs always evaluate (so branch-protection required-check semantics work)
-  # but short-circuit when nothing relevant changed.
-  changes:
-    name: detect changes
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      pull-requests: read
-    outputs:
-      lockfile: ${{ steps.filter.outputs.lockfile }}
-    steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
-        with:
-          fetch-depth: 0
-          persist-credentials: false
-      - uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3.0.3
-        id: filter
-        with:
-          filters: |
-            lockfile:
-              - '**/Cargo.lock'
-              - '**/Cargo.toml'
-              - 'qa/supply-chain/**'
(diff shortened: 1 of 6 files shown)


.github/workflows/vet.yml
```diff
@@ -0,0 +1,99 @@
...
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  # Detect whether anything that could affect vet/audit changed. The downstream
+  # jobs always evaluate (so branch-protection required-check semantics work)
+  # but short-circuit when nothing relevant changed.
```
ebfull: Will this make PRs that _don't_ modify the e.g. `Cargo.lock` (and therefore skip the action) themselves violate the branch protection requirements?

TalDerei:
@ebfull feel free to push exemption updates or whatever to unvetted deps to pass CI.

.github/workflows/vet.yml
```diff
@@ -0,0 +1,99 @@
...
+              - '**/Cargo.lock'
+              - '**/Cargo.toml'
+              - 'qa/supply-chain/**'
+              - '.github/workflows/vet.yml'
+
+  vet:
+    name: cargo vet
+    needs: changes
+    if: needs.changes.outputs.lockfile == 'true' || github.event_name == 'push'
+    runs-on: ubuntu-latest
```
ebfull, line 21: add `timeout-minutes: 15` (for example) wherever that's supposed to go

ebfull (changes requested):

.github/workflows/vet.yml
TalDerei: it won't violate the branch protection requirements, that's explicitly what https://github.com/tachyon-zcash/ragu/pull/732/commits/c187440161f0aa9c61d6a2ed1c026ef8b7f87f6b was supposed to prevent. but for simplicity, will change to always run unconditionally.

TalDerei: simplified in https://github.com/tachyon-zcash/ragu/pull/732/commits/e6f4951625c7fb384b920495125dd5401c9bf2a6. these are fast workloads, so savings are negligible anyway.

ebfull (approved):

PR #736, opened 2026-05-24
chore(ci): manual dispatch trigger for bench workflows

TalDerei:
Prior to this change, benches didn't block CI on regression analysis (it's not in the status checks list) and [summaries](https://github.com/tachyon-zcash/ragu/pull/729) were moved into the CI workflow directly where they'd be inspectable. Regardless, I find myself basically ignoring _gungraun_ benchmarks as they're currently configured. Part of it is the presentation, which I don't find very useful, and the occasional noise (like [this](https://github.com/tachyon-zcash/ragu/pull/497#issuecomment-3937683913)), and it's hard to tell the significance of a measurement because it may be on a nano/millisecond scale. The regression analysis is wired to use gungraun. 

I personally use valgrind more often because gungraun has limitations around measuring wall-clock time (only measures instruction counts which are deterministic to be fair) and doesn't support multi-core workloads (reference https://github.com/tachyon-zcash/ragu/pull/510). 

Anyways, personally in favor of using a manual trigger in CI for things we know affect performance. This configures a lightweight build job for PRs (compile only, no valgrind/gungraun) and the existing bench job behind manual dispatch, so regressions only surface on manual triggers. These workloads can also be run locally.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -4,6 +4,12 @@ on:
   push:
     branches: [main]
   pull_request:
+  workflow_dispatch:
+    inputs:
+      base_sha:
+        description: 'Baseline commit SHA for regression comparison (defaults to latest main)'
+        required: false
+        default: ''
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -18,6 +24,7 @@ env:
 jobs:
   changes:
     name: detect changes
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     outputs:
       rust: ${{ steps.filter.outputs.rust }}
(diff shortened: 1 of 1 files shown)


TalDerei:
after investigating the hung up CI jobs, nothing in the queue suggested there was an ingestion problem where runners were stuck. it seems the issue was that the branch protection rules expected a check by specific name, but the workflow never fired because the test job was skipped, so those other jobs were basically stuck waiting. I don't think this was new, but it seems to have surfaced from enabling status checks in our ruleset and it's now only surfacing on PRs with purely workflow related changes...

anyways, our branch protection suggest that these tests are required on every PR, so we can relax path filters to unconditionally run them. that seems too strict and unnecessary, so I instead opted to enable the test OS matrix (instead of individual test runners) as the required check in branch protection and it correctly skips them in this PR. rulesets have been updated accordingly.

ebfull (approved):

Issue #728, opened 2026-05-23
Implement the 'patcher' technique for soundness fuzzing


TalDerei:
Inspired by [zksecurity's paper](https://dl.acm.org/doi/pdf/10.1145/3713081.3731718) "Towards Fuzzing Zero-Knowledge Proof Circuits", they define a technique called the 'Patcher' that uses a mutation-based approach for finding more aggressive under-constrained bugs.

Soundness bugs require witnesses that the circuit should reject but instead incorrectly accepts. Random mutation alone rarely works because bugs typically occur in corner cases, so the paper proposes: "A promising approach is, to begin with a valid witness and then introduce slight modifications that make it invalid according to the oracle, thus exposing under-constrained soundness issues."

We've already implemented that in the `fuzz_witness_cheat` target in https://github.com/tachyon-zcash/ragu/pull/708, which does a single witness mutation (mid stream during circuit synthesis) and then downstream constraint can check it and fail. It basically compares that the honest fingerprint (fingerprint from honest witness) doesn't equal cheat fingerprint (fingerprint from mutated witness).

The patcher technique extends this for _actual_ under-constrained coverage; it swaps one and adjusts other witness values downstream so the constraints still hold (letting the cheat propagate further instead of being immediately caught). For this, we'd build a constraint graph that records the synthesis pass, and then patcher walks graph and does this mutation + repair work. We'll also need a fingerprint concept here to compare outputs. This can then be possibly wired into testing our internal recursion circuits as the next target (or dummy circuits, but whatever it is we need some kind of real circuit substrate, call it `fuzz_circuit_cheat` that the underlying circuit has `enforce_zero` calls!). 

It's worth noting that our fuzzing today as it currently exists doesn't really check for underconstrained bugs, but instead basically catches gadget API robustness bugs (that random compositions of gadget calls don't crash / panic, OOB, or misbehave under arbitrary inputs). Again `fuzz_witness_cheat` is an idiomatic example of this which composes random op calls. 

--------------

<img width="414" height="255" alt="Image" src="https://github.com/user-attachments/assets/2102261d-9398-4c3f-8904-c84f0162fc4f" />

zmanian:
I want to take a crack at this.

TalDerei:
closed by https://github.com/tachyon-zcash/ragu/pull/794.

PR #737, opened 2026-05-24
Implement patcher soundness fuzzing

zmanian:
## Summary
- Rework the patcher MVP to live entirely under `qa/fuzz`.
- Add a pure `qa/fuzz/src/patcher.rs` op-stream repair planner with unit coverage for downstream witness-slot repair behavior.
- Update `fuzz_witness_cheat` to run honest, unpatched-cheat, and patched-cheat `Simulator` paths, then signal only when the patched cheat preserves the observable fingerprint after excluding directly patched support slots.
- Leave the production-crate driver hooks, general circuit patcher, and recursion/PCD targets for a follow-up if the fuzz-local target proves insufficient.

Addresses #728

## Test Plan
- `cargo test --manifest-path qa/fuzz/Cargo.toml patcher`
- `cargo check --manifest-path qa/fuzz/Cargo.toml --bin fuzz_witness_cheat`
- `cargo check --manifest-path qa/fuzz/Cargo.toml --bins`
- `rustfmt --edition 2024 --check qa/fuzz/src/lib.rs qa/fuzz/src/patcher.rs qa/fuzz/fuzz_targets/fuzz_witness_cheat.rs`
- `git diff --check`

### DIFF
--- a/qa/fuzz/Cargo.toml
+++ b/qa/fuzz/Cargo.toml
@@ -141,9 +141,8 @@ test = false
 doc = false
 bench = false
 
-# Mid-stream witness cheating: honest vs cheat paths through the Simulator.
-# Currently a Simulator-robustness fuzzer; intended as a soundness oracle once
-# the patcher work lands (see issue tracking the patcher technique).
+# Mid-stream witness cheating: honest vs unpatched/patched cheat paths through
+# the Simulator, with downstream witness-slot repair for soundness signals.
 [[bin]]
 name = "fuzz_witness_cheat"
 path = "fuzz_targets/fuzz_witness_cheat.rs"
--- a/qa/fuzz/README.md
+++ b/qa/fuzz/README.md
@@ -41,7 +41,7 @@ All four share an essentially identical `Op` enum and dispatch — see the
 |---|---|
 | `fuzz_element_ops` | Completeness — random gadget compositions must not crash and must produce internally-consistent witnesses. The substrate. |
 | `fuzz_witness_coverage` | Same as `fuzz_element_ops` plus a post-run witness-state hash spread across coverage branches. Biases the fuzzer toward distinct internal witness states. Opt-in POC (~28% throughput cost). |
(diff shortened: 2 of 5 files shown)


TalDerei:
I had thought this would need a dedicated driver impl, as the PR implements in some form. will review

TalDerei:
after briefly inspecting the diff, same feedback as https://github.com/tachyon-zcash/ragu/pull/738#issuecomment-4530732916. this introduces some pretty invasive changes to our crates outside of `qa/fuzz`. few other process notes: the patcher changes should also live under `qa/fuzz`, and although the closing issue suggests the recursion circuits an idiomatic target for this fuzzing workflow, that was more idealistic than practical (probably) at this stage so the actual target would be dummy circuit impls in the same fuzzing directory. anyways, will think through a refactoring here.

zmanian:
Reworked this PR into the smaller fuzz-local version:

- diff is now limited to `qa/fuzz`
- removed the production crate driver/hooks and circuit/PCD patcher target from this PR
- `fuzz_witness_cheat` now uses a local op-stream patcher helper with unit tests
- follow-up can discuss minimal `unstable-fuzzing` hooks only if the local target is too weak

Verification on the rewritten branch:

- `cargo test --manifest-path qa/fuzz/Cargo.toml patcher`
- `cargo check --manifest-path qa/fuzz/Cargo.toml --bin fuzz_witness_cheat`
- `cargo check --manifest-path qa/fuzz/Cargo.toml --bins`
- `rustfmt --edition 2024 --check qa/fuzz/src/lib.rs qa/fuzz/src/patcher.rs qa/fuzz/fuzz_targets/fuzz_witness_cheat.rs`
- `git diff --check`

zmanian:
Follow-up context: I opened #739 as the next, stacked step after this PR.

The split is:

- #737 keeps the initial patcher MVP to a pure `qa/fuzz` op-stream oracle. It answers: can we mutate a witness slot, repair downstream-created support slots, and detect same-observable-fingerprint behavior without touching production crates?
- #739 checks the next question at a real `Circuit::witness` boundary, but still entirely inside `qa/fuzz`: it adds a fuzz-local dummy `Circuit` target that reuses this PR's patcher policy and compares a selected public output element instead of the whole final stack.

So #739 is not trying to revive the broader production-driver/API changes from the earlier diff. It is meant to validate whether the `qa/fuzz`-local approach is strong enough before considering any `unstable-fuzzing` hooks in core/primitives/circuits.

TalDerei:
superceeded by https://github.com/tachyon-zcash/ragu/pull/794 and https://github.com/tachyon-zcash/ragu/issues/793, which implement this in a more piecemeal fashion.

Issue #558, opened 2026-03-10
Fuzzing infrastructure support


TalDerei:
Synergetic with proptests in https://github.com/tachyon-zcash/ragu/issues/523; Inspired by Aztec's fuzzing talk https://www.youtube.com/watch?v=ItVUCP1El2E, I researched different fuzzing frameworks (black, white, grey) and stumbled on [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz) which is a [grey-box](https://www.fuzzingbook.org/html/GreyboxFuzzer.html) fuzzer used in industry. It's a wrapper around Google's [libFuzzer](https://llvm.org/docs/LibFuzzer.html) which uses LLVM's [SanitizerCoverage](https://clang.llvm.org/docs/SanitizerCoverage.html) to track branch coverage and mutate towards unexplored paths. After integrating cargo-fuzz, we can use an LLM to generate the harness code to fuzz targets. We can also append to our `qa/` subdirectory any scaffolding as necessary beyond the artifacts fuzzing generates. This can be configured as a dev-dep (probably requires nightly) in the `ragu_testing` crate.

TalDerei:
this was already closed by initial fuzzing scaffolding efforts.

PR #738, opened 2026-05-25
Add floor planner fuzzing and target inventory helper

### LINKED ISSUE #726: Fuzz floor planner
Follow-up to https://github.com/tachyon-zcash/ragu/issues/709, fuzzing the floor planner when it's fully implemented (current impl is just a prefix-sum).

zmanian:
## Summary
- add public floor-planner inspection helpers and a new `fuzz_floor_planner` target over real circuit topologies, including nested routines
- derive local and scheduled fuzz target inventories from `qa/fuzz/Cargo.toml` via `list-targets.sh`
- update fuzz README target count and CI notes

Closes #726.
Partially addresses #558 by removing target-list drift between `Cargo.toml`, `fuzz.sh`, and continuous fuzzing.

## Verification
- `cargo check --manifest-path qa/fuzz/Cargo.toml --bins`
- `cargo test -p ragu_circuits`
- `rustfmt --edition 2024 --check crates/ragu_circuits/src/floor_planner.rs qa/fuzz/fuzz_targets/fuzz_floor_planner.rs`
- `bash -n qa/fuzz/fuzz.sh qa/fuzz/list-targets.sh`
- `git diff --check`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p ragu_circuits --document-private-items`

### DIFF
--- a/.github/workflows/fuzz-cron.yml
+++ b/.github/workflows/fuzz-cron.yml
@@ -32,8 +32,26 @@ permissions:
   contents: read
 
 jobs:
+  targets:
+    name: enumerate fuzz targets
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.targets.outputs.targets }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+
+      - name: List fuzz targets
+        id: targets
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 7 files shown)


TalDerei:
closing as premature since floor planner isn't fully implemented (currently just uses a naive prefix sum), as the closing issue suggests. moreover, we shouldn't expand the API surface like this outside of the `qa/fuzz` subdirectory. if we absolutely need to, things should be behind the `unstable-fuzzing` flag.

Issue #466, opened 2026-02-14
Exercise error-swallowing behavior with malicious circuit


TalDerei:
Exercise class of completeness failures by demonstrating error-swallowing (error stack propagation) tests with a malicious circuit, references https://github.com/tachyon-zcash/ragu/pull/465#issuecomment-3900849899. Also, circuit synthesis still has insufficient test coverage (only a few tests from https://github.com/tachyon-zcash/ragu/issues/350), and that should be expanded.

TalDerei:
closed by #466.

PR #741, opened 2026-05-26
Improvements and clarifications about security analysis

ebfull:
Used this as practice to flex my Lean muscles a bit.


crates/ragu_primitives/src/element.rs
TalDerei, line 1: commit message in https://github.com/tachyon-zcash/ragu/pull/741/changes/c059526e872e0db10c3d0e9bd571b5fdfb600eb3 is misleading, includes "the Lean reimpl + autogen traces to match", but the commit doesn't introduce any of that.

crates/ragu_primitives/src/invertible.rs
```diff
@@ -147,6 +213,83 @@ impl<'dr, D: Driver<'dr>> Consistent<'dr, D> for Invertible<'dr, D> {
...
+///
+/// A `NonzeroBank` is only constructible by [`NonzeroBank::scope`] (which
+/// always discharges the proof) or by internal callers that assert every
+/// fold is nonzero by structural argument.
+pub struct NonzeroBank<'dr, D: Driver<'dr>> {
+    // `None` is unchecked mode: `fold` emits no constraint and the bank
+    // discharges to a no-op when dropped. `Some(p)` is discharging mode:
+    // `fold` multiplies the new factor into `p` and the closing scope
+    // constrains `p != 0`.
+    product: Option<Element<'dr, D>>,
```
TalDerei, line 248: this is an interesting way to model bank state (discharging or unchecked mode) using an `Option`.

crates/ragu_primitives/src/invertible.rs
```diff
@@ -0,0 +1,494 @@
...
+        dr: &mut D1,
+        a: &Bound<'dr, D2, Self>,
+        b: &Bound<'dr, D2, Self>,
+    ) -> Result<()> {
+        // Soundness: comparing only the element suffices because
+        // `alloc_with_advice` enforces `element * inverse = 1`, so equal
+        // elements have equal inverses in any satisfied assignment.
+        Gadget::enforce_equal(&a.element, dr, &b.element)
+    }
+}
```
TalDerei, line 222: this soundness note caught my eye and I traced the commits prior to its introduction; reloading context, `enforce_consistent` (from the `Consistent` trait defined in https://github.com/tachyon-zcash/ragu/pull/388) internally calls this `enforce_equal_gadget`, and it was created to run at the staging boundary (where prior constraints have been stripped and wires rebound). in that context, it needs to re-emit gadget constraints / invariants on its own wires, without relying on any prior context or trusting anything. if I follow that logic, this isn't technically what this method is doing anymore.

the prior method impl

```ts
  fn enforce_equal_gadget(dr, a, b) -> Result<()> {
      Gadget::enforce_equal(&a.element, dr, &b.element)?;  
      Gadget::enforce_equal(&a.inverse, dr, &b.inverse)?;   
      Ok(())
  }
```

had an extra constraint that was subsequently stripped. now I understand the optimization here, and it's safe because `Invertible`s carry their own constraint that inverse satisfies (`element * inverse = 1`) in the current driver. but, `enforce_consistent` isn't supposed to assume that. by making that constraint, we're violating the `Consistent` API contract's defintion since it has to work as if nothing has been emitted before. 

if i extrapolate, there's a latent bug here if an `Invertible` say flows through the staging boundary (appears in  `Stage::OutputKind` or inside a `#[derive(Consistent)]`). that isn't the case now, but just noting that we should guard against these extensions. 

i'd add back the extra constraint to avoid violating the `Consistent` contract.

crates/ragu_primitives/src/invertible.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    /// Opens a scope in which nonzero proofs can be batched. The discharge
+    /// runs on the success path of `body`; on any `Err` the scope
+    /// short-circuits without discharging — but no [`Nonzero`] minted inside
+    /// the body can outlive the `Err` return, so soundness is preserved.
+    ///
+    /// This costs one gate and two constraints for the final discharge,
+    /// plus one gate per [`fold`](Self::fold) in discharging mode, on top
+    /// of whatever the body itself emits.
```
TalDerei: this doc comment is a bit misleading

```suggestion
    /// Opens a scope in which nonzero proofs can be batched. The discharge
    /// runs on the success path of `body`; on any `Err` the scope
    /// short-circuits without discharging. Callers must not capture Nonzeros 
    /// minted inside the body into outer state that survives the Err return — if 
    /// they do, those Nonzeros will exist without a backing discharge constraint.
    /// The type system does not enforce this; it is a caller obligation.
    ///
    /// This costs one gate and two constraints for the final discharge,
    /// plus one gate per [`fold`](Self::fold) in discharging mode, on top
    /// of whatever the body itself emits.

File: crates/ragu_primitives/src/invertible.rs [hunk 4]
```diff
@@ -0,0 +1,494 @@
...
+        Self::new_unchecked(self.element.negate(dr))
+    }
+
+    /// Squares this element. The result is nonzero since `F` is an integral
+    /// domain.
+    ///
+    /// This costs one gate.
+    pub fn square(&self, dr: &mut D) -> Result<Self> {
+        Ok(Self::new_unchecked(self.element.square(dr)?))
+    }
```
TalDerei, line 54: nit; some methods in this file have a misleading cost, should be "one gate and two constraints".

qa/lean/Ragu/Contrib/EndoscalarProof.lean
TalDerei, line 1: this file stood out to me from a purely organizational standpoint, because it falls outside the ragu <> Clean FV pipeline entirely. this is configured as pure lean lemmas, which don't seem to be used outside of this file. in theory, some of these lemmas may be needed by an actual reimpl (eg. a `lift_endoscalar` formalization), and would need to be exported as theorems since they're currently private.

maybe this and `CircuitFlattening.lean` should live in like a helper subdirectory, or maybe `qa/lean/Ragu/Specs/`?

qa/crates/lean_extraction/src/main.rs
```diff
@@ -182,6 +194,11 @@ static EXPORT_TARGETS: &[ExportTarget] = &[
     },
 ];
 
+/// Hand-written Lean modules that are not extracted from Rust but must still be
+/// imported by the top-level `Ragu.lean`, so that the Lean build (and its
+/// `--wfail` `sorry` check) covers them.
+static EXTRA_IMPORTS: &[&str] = &["Ragu.EndoscalarProof"];
```
TalDerei: nit; this is a small lean extraction codegen change to support those individual files you added. but this seems a little hacky because if you forget to append to `EXTRA_IMPORTS`, then there's a file that sits in the tree that's silently typechecked by nothing. there should be some auto discovery impl here that doesn't rely on like appending files to this static var.

qa/lean/Ragu/Circuits/NonzeroBank/ScopeK2.lean
TalDerei, line 1: the clean reimpl here is monomorphic (hardcodes K = 2), but Clean supports polymorphism. our polymorphism rule in our skills mention that the reimpl should be parametric in K, and the extraction surface (autogen + formal instance + rust extractor) staying monomorphic at possibly multiple Ks. reference #686 and #710 for context.

qa/lean/Ragu/Circuits/Element/DivNonzero.lean
```diff
@@ -5,59 +5,93 @@ import Ragu.Circuits.Core.Mul
...
-deriving ProvableStruct
+  inverse : UnconstrainedDep field F
+deriving CircuitType
 
--- quotient * denominator = numerator, with denominator = y, numerator = x
+/-- `Element::divide(&y_nonzero)` after `y.enforce_nonzero(...)`.
+
+  This reimpl inlines both halves so the new wire `a` allocated by
+  `Invertible::alloc` is in scope when the divide step links its
+  denominator. Step by step:
```
TalDerei: we should compose subgadgets rather than duplicatively inlining in the reimpl. in this context, `DivNonzero.main` can delegate to `EnforceNonzero.circuit` and `Divide.circuit`.

crates/ragu_primitives/src/invertible.rs
```diff
@@ -0,0 +1,494 @@
...
+}
+
+impl<'dr, D: Driver<'dr>> Consistent<'dr, D> for Invertible<'dr, D> {
+    fn enforce_consistent(&self, dr: &mut D) -> Result<()> {
+        let value = D::just(|| *self.element.value().take());
+        let inverse_value = D::just(|| *self.inverse.value().take());
+        Self::alloc_with_advice(dr, value, inverse_value)?.enforce_equal(dr, self)
+    }
+}
```
ebfull, line 205: Fascinating. Codex flagged that `enforce_equal_gadget` relying on invariants to hold for both operands (which seems fair and normal) means that `Consistent` implementations cannot safely use it for comparison, since the `Self` instance is not guaranteed to hold any invariants.

This is unsound.

qa/lean/Ragu/Circuits/Point/AddIncomplete.lean
```diff
@@ -1,124 +1,158 @@
...
-    simp [Spec.Point.add_incomplete, h]
-
-    let h_lemma := Lemmas.add_incomplete_preserves_membership ⟨input_P1_x, input_P1_y⟩ ⟨input_P2_x, input_P2_y⟩ curveParams
-    simp [Spec.Point.add_incomplete, h] at h_lemma
-    specialize h_lemma h_P1_mem h_P2_mem
-    ring_nf at ⊢ h_lemma
-    simp_all only [id_eq, inv_pow, and_self]
-  · grind
+    Element.EnforceNonzero.circuit, Element.EnforceNonzero.Spec,
+    circuit_norm] at h_holds ⊢
```
TalDerei: this diverges from the pattern we've been using for how subcircuit lemmas get applied (even though technically it's structurally similiar). it should pass subcircuit lemmas to `circuit_proof_start` like this:

```ts
  circuit_proof_start [
    Element.Mul.circuit, Element.Mul.Assumptions, Element.Mul.Spec,
    Element.Divide.circuit, Element.Divide.Assumptions, Element.Divide.Spec,
    Element.Square.circuit, Element.Square.Assumptions, Element.Square.Spec,
    Element.EnforceNonzero.circuit, Element.EnforceNonzero.Spec]
```
same with `Point/DoubleAndAddIncomplete.lean`.

qa/lean/Ragu/Circuits/Element/Divide.lean
```diff
@@ -0,0 +1,77 @@
...
+/-- Verifier side: the circuit enforces `quotient · y = x`. When
+`y ≠ 0`, this pins `quotient = x / y`. When `y = 0 ∧ x ≠ 0` no witness
+exists; when `y = 0 ∧ x = 0` the quotient is unconstrained but the
+premise is false. Callers are responsible for `(x, y) ≠ (0, 0)`. -/
+def Assumptions (input : Input (F p)) (_data : ProverData (F p)) :=
+  input.y ≠ 0 ∨ input.x ≠ 0
+
+def Spec (input : Input (F p))
+    (out : field (F p)) (_data : ProverData (F p)) :=
+  out = input.x / input.y
```
TalDerei, line 43: manually inspected all the specs across new lean targets, and they seem semantically correct.

TalDerei (commented):
left some observations, great work on the lean impls!

qa/crates/lean_extraction/src/main.rs
ebfull: Possibly easier to just add a "Supporting" module which imports from there, and `EXTRA_IMPORTS` here is substituted with code that only produces the single supporting module.

TalDerei:
I can optionally address some of my own feedback here. and don't forget to use our fv skills next time, they've gone through many iterations of review and refinement!

crates/ragu_primitives/src/invertible.rs
ebfull: #466

book/src/guide/gadgets/gadgetkind.md
```diff
@@ -34,19 +34,31 @@ associated type `Kind` that relates back to its corresponding
...
+Gadgets offer the
+[`GadgetKind::enforce_conservative_equal_gadget`][enforce-equal] method to
+specify conservative equality constraints by pairing corresponding wires in the
+gadget's canonical traversal.
+
+Most circuit code should avoid using this method directly. Prefer
+[`GadgetExt::enforce_equal`][gadgetext-enforce-equal], backed by
+[`GadgetEquals`][gadget-equals], for ordinary gadget comparisons. Conservative
+equality is mainly useful for infrastructure such as consistency checks and
+wire-substitution paths that need canonical pairwise constraints.
```
TalDerei, line 61: ACK – this is the context free version that contains every wire pair, and `enforce_equal` is the optimized version outside of the `Consistent` API contract. good separation of concerns.

TalDerei (commented):
post-hoc ACK on gadget equality and lean changes.

Issue #742, opened 2026-05-29
Restrict conservative gadget equality to wire-pair enforcement


ebfull:
Conservative equality for `Gadget` / `GadgetKind` is currently documented as pairwise equality over corresponding wires, but the implementor-facing API passes a full `Driver` into `GadgetKind::enforce_conservative_equal_gadget`. That makes the contract fragile: handwritten impls can accidentally allocate gates, add arbitrary constraints, or call invariant-aware equality helpers even though this path is supposed to be the raw conservative equality fallback.

Introduce a small gadget-module API for conservative wire equality, with a method name that stays explicit, e.g. `enforce_conservative_equal(&mut self, a: &Wire, b: &Wire)`. The decorated full-driver adapter should probably live inside `ragu_core::gadgets`, since this is an implementation detail of the gadget API rather than a new driver abstraction.

Then thread this restricted interface through the conservative equality APIs and derive output. `GadgetEquals` / `GadgetExt::enforce_equal` should remain the invariant-aware equality path; the goal here is only to make the conservative path enforce its pairwise-wire contract at the type level.

TalDerei:
Follow-up to https://github.com/tachyon-zcash/ragu/pull/741.

PR #743, opened 2026-05-29
chore(core): conservative gadget equality fragility

TalDerei:
Follow-up to https://github.com/tachyon-zcash/ragu/pull/741 and diverges from suggested approach in https://github.com/tachyon-zcash/ragu/pull/742. The API fragility basically comes from the fact that `enforce_conservative_equal_gadget` hands the implementor a full driver, but nothing in the API contract prevents say doing more than just enforcing wire-pair equality (like allocating mul gates or something). #742 suggests using a decorator to restrict that (reminds me of #600 where we added a constrained routing method / `BondingValidator` for `MultiStageCircuit` that caught these kinds of driver violations at runtime). The suggested approach in that issue would catch it at compile time. 

Anyways, this approach instead targets the `map_gadget` layer. It hinges on the observation that if the `map_gadget` is trusted to traverse all the wires, then it should be trusted to derive equality from that too. So instead of having a hand written `enforce_conservative_equal_gadget` impl per gadget, we maintain one generic impl that internally calls `map_gadget` and does two traversals over dedicated `WireMap` impls (`Collector` which collects the wires and then `EnforceEq` that does the wire equality). In that context, conservative gadget equality is inherited by all gadgets unconditionally. 

It doesn't touch the "invariant-aware" side.


TalDerei:
closing in favor of impl in https://github.com/tachyon-zcash/ragu/pull/744.

PR #744, opened 2026-05-29
chore(core): conservative gadget equality decorator

TalDerei:
Follow-up to https://github.com/tachyon-zcash/ragu/pull/741 and aligns with the suggested approach in https://github.com/tachyon-zcash/ragu/pull/742, and alternative to implementation in https://github.com/tachyon-zcash/ragu/pull/743.

### DIFF
--- a/book/src/guide/gadgets/gadgetkind.md
+++ b/book/src/guide/gadgets/gadgetkind.md
@@ -52,10 +52,9 @@ use it to count, substitute, or extract wires.
 Gadgets implement the
 [`GadgetKind::enforce_conservative_equal_gadget`][enforce-equal] method to
 specify conservative equality constraints by pairing corresponding wires in the
-gadget's canonical traversal. This is the raw fallback used where a gadget's
-invariants are *not* guaranteed to hold, so the contract is strict: it must
-constrain each pair of corresponding wires and do nothing else — no extra gates,
-no arbitrary constraints, and no invariant-aware equality shortcuts.
+gadget's canonical traversal. The contract is strict: it must constrain each
+pair of corresponding wires and do nothing else — no extra gates, no arbitrary
+constraints, and no gadget-specific shortcuts.
 
 To make that contract enforceable at the type level rather than by convention,
 the method does not receive a full [`Driver`][driver-trait]. Instead it is
@@ -68,10 +67,7 @@ macro generates exactly these calls.
 
 Most circuit code should avoid this path directly. Prefer
 [`GadgetExt::enforce_equal`][gadgetext-enforce-equal], backed by
-[`GadgetEquals`][gadget-equals], for ordinary gadget comparisons — it may
-constrain fewer wires by relying on a gadget's invariants. Conservative equality
-is reserved for infrastructure such as consistency checks and wire-substitution
-paths, which must re-establish constraints without trusting any prior invariant.
+[`GadgetEquals`][gadget-equals], for ordinary gadget comparisons.
 
 ## Safety
 
--- a/book/src/guide/gadgets/gadgetkind.md
+++ b/book/src/guide/gadgets/gadgetkind.md
@@ -49,16 +49,29 @@ use it to count, substitute, or extract wires.
 
 ## `enforce_conservative_equal_gadget`
 
-Gadgets offer the
+Gadgets implement the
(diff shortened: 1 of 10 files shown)


book/src/guide/gadgets/gadgetkind.md
```diff
@@ -49,16 +49,29 @@ use it to count, substitute, or extract wires.
...
+Gadgets implement the
 [`GadgetKind::enforce_conservative_equal_gadget`][enforce-equal] method to
 specify conservative equality constraints by pairing corresponding wires in the
-gadget's canonical traversal.
-
-Most circuit code should avoid using this method directly. Prefer
+gadget's canonical traversal. This is the raw fallback used where a gadget's
+invariants are *not* guaranteed to hold, so the contract is strict: it must
+constrain each pair of corresponding wires and do nothing else — no extra gates,
+no arbitrary constraints, and no invariant-aware equality shortcuts.
```
ebfull: I deliberately avoided describing in user-facing docs the idea of "invariants not holding" and suggesting to the reader / API consumer that valid gadgets can exist in an invariant violating state. Invariants can always be violated. Ragu repairs those violations when it breaks them, so the user does not need to see it.

I want to give the users a contract they can satisfy and be sure that the invariants are satisfied whenever they are exposed to the abstraction. And so I do not even want to tell them the reason why a function exists is to

crates/ragu_core/src/gadgets.rs
```diff
@@ -95,6 +95,31 @@ use super::{
...
+/// is the raw fallback used where a gadget's invariants are not guaranteed (for
+/// example consistency re-establishment and wire-substitution paths). Handing
+/// implementations this adapter instead of a full [`Driver`] enforces that
+/// contract at the type level rather than by documentation alone: the only
+/// operation it exposes is
+/// [`enforce_conservative_equal`](Self::enforce_conservative_equal).
+pub struct WireEqualizer<'a, 'dr, D: Driver<'dr>> {
+    dr: &'a mut D,
+    _marker: core::marker::PhantomData<&'dr ()>,
+}
```
TalDerei, line 109: one of the benefits of https://github.com/tachyon-zcash/ragu/pull/743 over this PR is technically, this restricts what conservative equality impls can do, but it doesn't prove that the handwritten impls visit every wire exactly once, so they can still omit calls, right? that'd require using the canonical `map_gadget` traversal.

ebfull (changes requested):
Mainly a doc nit that should be propagated through the changes.

crates/ragu_core/src/gadgets.rs
TalDerei: costs in #743 seem kinda negligible, but fine with either approach.

TalDerei:
pushed some doc / book updates, see other discussion thread as well.

crates/ragu_core/src/gadgets.rs
```diff
@@ -131,13 +149,32 @@ pub trait Gadget<'dr, D: Driver<'dr>>: Clone {
...
         dr: &mut D2,
         other: &Self,
     ) -> Result<()> {
-        Self::Kind::enforce_conservative_equal_gadget::<D2, D>(dr, self, other)
+        let mut eq = WireEqualizer {
+            dr,
+            _marker: core::marker::PhantomData,
+        };
+        self.enforce_conservative_equal_with(&mut eq, other)
+    }
```
ebfull: Can't this just immediately delegate to the `enforce_conservative_equal_gadget` implemented on the `Kind` type? Then the `Gadget` trait does not need the extra method exposed.

ebfull (changes requested):

crates/ragu_core/src/gadgets.rs
TalDerei: yeah nice, that kills a layer of misdirection (`enforce_conservative_equal` now delegates directly to `Kind::enforce_conservative_equal_gadget`). 

but couldn't delete `enforce_conservative_equal_with` since it's used for how we recursively traverse subgadget fields, so moved that into `WireEqualizer`. there's a way to get rid of that method, but it'd make the call sites pretty ugly. https://github.com/tachyon-zcash/ragu/pull/744/commits/81c54e2a3fa6702aa3a26599c49cdb3f6c7c24e3 was my interpretation of your suggested change.

ebfull (approved):

Issue #746, opened 2026-06-01
Ensure --all-features works


ebfull:
This is something we should do to ensure that `docs.rs` can build docs with all feature annotations available.

PR #747, opened 2026-06-01
arithmetic: support modern and legacy dependency modes

ebfull:
* use `modern-deps` feature to access the modern stack
* use `legacy-deps` feature to access the legacy stack (old rand/ff/group/pasta/etc.)
* intended to be temporary, `legacy-deps` will be deleted eventually
* can't use both at the same time (see also #746)


justfile
```diff
@@ -45,15 +45,20 @@ book COMMAND: _book_setup
 
 # run all tests
 test *ARGS:
-  cargo test --workspace --all-features {{ARGS}}
+  cargo test --workspace {{ARGS}}
+
+# check the temporary legacy dependency surface for external consumers
+check_legacy *ARGS:
+  cargo check -p ragu_arithmetic -p ragu_pasta -p ragu_core -p ragu_primitives -p ragu_circuits -p ragu_pcd --lib --no-default-features --features "alloc legacy-deps" {{ARGS}}
+  cargo check -p ragu --lib --no-default-features --features legacy-deps {{ARGS}}
```
ebfull, line 53: I don't use `just` so somebody else will need to review these.

qa/fuzz/Cargo.toml
```diff
@@ -26,9 +26,12 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = "0.4"
 arbitrary = { version = "1", features = ["derive"] }
+# This workspace cannot inherit repo-root [workspace.dependencies]. Fuzzing is
+# intentionally modern-only, so keep these direct ff/group/pasta_curves/rand
+# entries in sync with the repo-root modern dependency set.
 ff = { version = "0.14.0-pre.1", default-features = false }
 group = { version = "0.14.0-pre.1", default-features = false }
```
TalDerei: something like `^0.14.0-pre.1` would technically allow cargo to resolve to the `0.14.0` release (at least for `ff` release). not suggesting we block this PR on tagging a `0.14.0` group release (side note: group `0.13.0` was never tagged which seems like a release hygiene slip). anyways, we should note to bump pins and regenerate the lockfile. this only affects the `modern-deps` side.

Cargo.toml
```diff
@@ -67,26 +82,24 @@ rust-version = "1.90"
...
 
 [workspace.dependencies]
+# Keep the unaliased modern ff/group/pasta_curves/rand entries in sync with
+# qa/fuzz/Cargo.toml. The cargo-fuzz workspace is intentionally isolated and
+# cannot inherit these workspace dependencies.
 ff = { version = "0.14.0-pre.1", default-features = false }
 group = { version = "0.14.0-pre.1", default-features = false }
-pasta_curves = { version = "0.5.1", features = ["deferred"] }
+pasta_curves = { git = "https://github.com/ebfull/pasta_curves", rev = "4b2c768ec4b14b2c7fa0f6e527e453cde97805aa", features = ["deferred"] }
 rand = "0.10"
```
TalDerei, line 91: lifting same comment from https://github.com/zkcrypto/group/pull/63;

"running 'cargo audit' flagged an open RUSTSEC advisory on this rand version, and [0.10.1](https://github.com/rust-random/rand/releases/tag/0.10.1) was a released fix. we could bump this and regenerate the lockfile to silence the audit warning."

downstream consumers should bump this dep.

Cargo.toml
```diff
@@ -67,26 +82,24 @@ rust-version = "1.90"
...
-
-# NOTE: `qa/fuzz/Cargo.toml` is its own workspace root
-# (cargo-fuzz convention) and `[patch.crates-io]` does not propagate into
-# nested workspaces. If you add, remove, or change any patch below, mirror
-# the change in that file or fuzz builds will resolve different versions
-# than the rest of the workspace.
-[patch.crates-io]
-# `ebfull/pasta_curves` branch `ff-0.14-for-ragu` HEAD as of 2026-05-21.
-# Stacks on the prior `ff-0.14-with-rand-0.10-and-deferred-redux` work
-# (ff/group 0.14.0-pre.0, rand 0.10, deferred reduction, try_random rename)
```
TalDerei, line 89: we support the deferred reduction, but we're not using the new `try_random` fallible API anywhere?

.github/workflows/rust.yml
```diff
@@ -80,7 +80,25 @@ jobs:
...
+        with:
+          persist-credentials: false
+      - name: Setup Rust
+        uses: ./.github/actions/rust-setup
+        with:
+          cache-suffix: legacy-deps
+      - name: Check runtime crates with legacy dependencies
+        run: cargo check -p ragu_arithmetic -p ragu_pasta -p ragu_core -p ragu_primitives -p ragu_circuits -p ragu_pcd --lib --locked --no-default-features --features "alloc legacy-deps"
+      - name: Check root crate with legacy dependencies
+        run: cargo check -p ragu --lib --locked --no-default-features --features legacy-deps
```
TalDerei, line 101: I know the legacy feature flag is a temporary shim, but there's a CI gap here where it basically only does `cargo check --lib` and doesn't run test coverage.

crates/ragu_testing/Cargo.toml
TalDerei, line 1: noticed that `ragu_testing` and `lean_extraction` still consume upstream `ff/group` directly instead of via `ragu_arithmetic` (which implies to me they're only modern deps which is a slight asymmetry to how the other crates are wired). 

**update:** saw this comment which seems to track and acknowledge this asymmetry:

```
# This workspace cannot inherit repo-root [workspace.dependencies]. Fuzzing is
# intentionally modern-only, so keep these direct ff/group/pasta_curves/rand
# entries in sync with the repo-root modern dependency set.
```

TalDerei (approved):
left some observations, nothing strictly blocking.

justfile
TalDerei: `justfile` changes seem fine to me besides dropping the `--all-features` flag (tracked in https://github.com/tachyon-zcash/ragu/issues/746), don't use it often either.

qa/fuzz/Cargo.toml
ebfull: Going to fix these after the `group` release.

Cargo.toml
ebfull: The comment previously just mentioned it to track what was in the branch. In any case it's now obsolete.

Issue #759, opened 2026-06-01
Investigate gadget-level spec composability for formal verification


TalDerei:
Extrapolating some ruminations by @ebfull. 

Currently, we can (1) compose gadgets, and (2) propagate child specs into the parent's reasoning (this is just subcircuit lemma that surfaces `Assumptions (preconditions) -> Spec (postcondition)` as a unified hypothesis). 

But, Clean's framework doesn't currently support (3) composing specs into a parent spec, or check spec compatibility as a reusable first class thing. If we added that, the missing pieces would be:
- some compatibility lemma/relation (where one gadget specs output guarantees imply _another_ gadgets assumptions)
- spec composition combiner / mechanism that derives a candidate parent spec (so in theory some top level gadget that composes these gadget calls would have it's soundness proof cleanly collapse). 

We'd have a commensurate version of gadget-level spec invariants on the Ragu side. For instance if you define a gadget, you have to define a spec.

PR #761, opened 2026-06-02
chore(fv): lean spike for endoscalar extract

TalDerei:
**Not intending to merge, for demonstration purposes;** Lean proves that our endoscaling [extract](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_primitives/src/endoscalar.rs#L77) primitive hides an under-constrained bug in isolation (under the guise of an undocumented precondition, it's not currently exploitable in our circuits modulo negligible probability). 

It's worth noting this wasn't caught by our draft endoscalar gadget coverage in https://github.com/tachyon-zcash/ragu/pull/711, because I hadn't gotten to `Endoscalar::GroupScale` and `Endoscalar::Extract` impls yet! However, repeated stuck goals it couldn't resolve would have been a strong smell something was weird. The spec it'd be trying to prove is false, so the goals can't close.

Concretely, we couldn't prove _unconditional_ soundness in `ExtractStuckDemo` because we'd hit a stuck goal (trying to prove the false statement). Instead we proved _conditional_ soundness (`extract_sound_of_shifted_ne_zero`), alongside a "prove the counterexample" form of its negation in `extract_unsound`. Basically the negation proves the circuit is satisfiable but admits a forged witness it should reject.

Trying to extract the deeper philosophy here is challenging beyond trying to exhaustively enumerate all the gadget preconditions we're making (orthogonal to https://github.com/tachyon-zcash/ragu/issues/759), and that's because these preconditions are sometimes silently undocumented. We should more rigorously document the gadget specs / assumptions we're making as rust-side hygiene before throwing lean at it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -217,3 +217,31 @@ jobs:
           tool: mdbook@0.4.52,mdbook-katex@0.9.4,mdbook-mermaid@0.16.2,mdbook-admonish@1.20.0
       - name: Build book
         run: mdbook build ./book
+
+  # Aggregate gate: a single, stable required status check. It always runs —
+  # even on non-Rust PRs where every job below is skipped — and fails only if a
+  # needed job actually failed or was cancelled (skipped counts as success).
+  # Require THIS in the ruleset instead of the individual job names: a skipped
+  # matrix job (`test`) reports the un-expanded name `test (${{ matrix.os }})`,
+  # so `test (ubuntu-latest)` etc. never report and block lean-/docs-only PRs.
+  ci-required:
+    name: ci-required
+    if: always()
+    needs:
+      - changes
+      - fmt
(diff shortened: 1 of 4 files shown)


TalDerei:
the fact that `lake build` in CI compiles on the first commit means lakefile isn't building every file under `Ragu/` (so orphans can hide say in `Contrib/`). We need to change the lakefile glob so lake build compiles all submodules whether or not they're imported and  FV tree is fully covered by CI, to avoid silent verification gaps.

**Update:** now it properly fails :)

Cherry picked commits in https://github.com/tachyon-zcash/ragu/pull/763.

TalDerei:
closing since this was for demonstration purposes. this will be patched in https://github.com/tachyon-zcash/ragu/pull/767. cc @ebfull.

PR #763, opened 2026-06-02
chore(fv): ci gate fv glob

TalDerei:
Cherry picks commits from https://github.com/tachyon-zcash/ragu/pull/761. 

Closes lakefile gaps referenced in https://github.com/tachyon-zcash/ragu/pull/761#issuecomment-4605456359, and CI gaps from our strict status checks enabled in our rulesets (specifically the CI test matrix should skip for `qa/lean` only changes). Also lifts some learnings from that same PR into our `fv-review` skill.


ebfull (approved):

LucidSamuel:
quick check:- was legacy-check intentionally left out of ci-required.needs, perhaps because it is separately required by the ruleset?

asking bevause it appears to be the only Rust CI job not included in the aggregate gate and If ci-required is the sole required check, a legacy-only failure could leave it green. Is this intentional? 
@TalDerei

TalDerei:
> quick check:- was legacy-check intentionally left out of ci-required.needs, perhaps because it is separately required by the ruleset?
> 
> asking bevause it appears to be the only Rust CI job not included in the aggregate gate and If ci-required is the sole required check, a legacy-only failure could leave it green. Is this intentional? @TalDerei

yes intentional, i'm reworking our required checks because some of them were hanging so had to temp disable them and just make them non-required.

PR #764, opened 2026-06-03
fix(primitives): endoscalar extraction

TalDerei:
An internal [auditing tool](https://www.zkao.io/) found that an `Endoscalar::extract` bit was under constrained at `elem + i == 0` (both quadratic residue branches are zero there, which is trivially a square, so a malicious prover could forge that bit). The bug in our endoscaling primitive was demonstrated in Lean in https://github.com/tachyon-zcash/ragu/pull/761.

To fix this, make a false bit also prove `elem + i != 0` (the usual inverse trick), which constrains it. That adds one extra multiplication gate per bit (128 bits), which increases `ComputeV` constraints. It also includes regression tests.

### DIFF
--- a/crates/ragu_primitives/src/endoscalar.rs
+++ b/crates/ragu_primitives/src/endoscalar.rs
@@ -20,7 +20,7 @@ use ragu_arithmetic::{
     ff::{Field, PrimeField, WithSmallOrderMulGroup},
 };
 use ragu_core::{
-    Result,
+    Error, Result,
     drivers::{Driver, DriverValue, LinearExpression, emulator::Emulator},
     gadgets::Gadget,
     maybe::Maybe,
@@ -28,7 +28,9 @@ use ragu_core::{
 
 use crate::{
     Boolean, Element, NonzeroBank, Point,
+    allocator::Allocator,
     promotion::Demoted,
+    util::InternalMaybe,
     vec::{CollectFixed, ConstLen, FixedVec},
 };
 
@@ -86,11 +88,6 @@ impl<'dr, D: Driver<'dr>> Endoscalar<'dr, D> {
         let mut value = D::just(|| 0u128);
         let mut constant = D::F::ZERO;
 
-        let mut coeff_0 = D::F::ZERO;
-        let mut coeff_1 = D::F::ZERO;
-        let coeff_2 = D::F::MULTIPLICATIVE_GENERATOR;
-        let coeff_3 = D::F::ONE - D::F::MULTIPLICATIVE_GENERATOR;
-
         for i in 0..(u128::BITS as usize) {
             let (sqrt, bit) = D::try_just(|| {
                 let value = *elem.value().take() + constant;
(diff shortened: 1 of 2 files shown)


crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -231,6 +204,113 @@ impl<'dr, D: Driver<'dr>> Endoscalar<'dr, D> {
...
+    //            + vb * (1 - MULTIPLICATIVE_GENERATOR)
+    let coeff_2 = D::F::MULTIPLICATIVE_GENERATOR;
+    let coeff_3 = D::F::ONE - D::F::MULTIPLICATIVE_GENERATOR;
+    dr.enforce_zero(|lc| {
+        lc.add_term(&D::ONE, (constant * coeff_2).into())
+            .add_term(bit.wire(), (constant * coeff_3).into())
+            .add_term(elem.wire(), coeff_2.into())
+            .add_term(vb.wire(), coeff_3.into())
+            .sub(square.wire())
+    })?;
```
TalDerei, line 309: shifted down verbatim from `extract` method.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -231,6 +204,113 @@ impl<'dr, D: Driver<'dr>> Endoscalar<'dr, D> {
...
+        let shifted = *elem.value().take() + constant;
+        if *bit.snag() {
+            Ok(D::F::ZERO)
+        } else {
+            shifted
+                .invert()
+                .into_option()
+                .ok_or_else(|| Error::InvalidWitness("endoscalar false bit at zero".into()))
+        }
+    })?;
```
TalDerei, line 233:
```ts
  let inverse = D::just(|| {
      let shifted = *elem.value().take() + constant;
      if *bit.snag() {
          D::F::ZERO
      } else {
          shifted.invert().unwrap_or(D::F::ZERO)
      }
  });
```

also seems like a simpler alternative.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -231,6 +204,113 @@ impl<'dr, D: Driver<'dr>> Endoscalar<'dr, D> {
...
+    // `shifted * inverse = 0`, which `inverse = 0` always satisfies (complete).
+    let (shifted_wire, _, is_not_zero) = dr.mul(|| {
+        Ok((
+            Coeff::Arbitrary(*shifted.value().take()),
+            Coeff::Arbitrary(*inverse.snag()),
+            bit.value().not().coeff().take(),
+        ))
+    })?;
+    dr.enforce_equal(&shifted_wire, shifted.wire())?;
+    dr.enforce_zero(|lc| lc.add(&is_not_zero).sub(&D::ONE).add(bit.wire()))?;
```
TalDerei, line 287: standard inverse trick used in `is_zero`.

ebfull:
I would actually much rather just switch this to using a full boolean decomposition. This was brought up earlier in the zkSecurity audit: we'll just avoid possible non-canonical decompositions by requiring the highest bit to be unset, which will force the prover to grind.

The problem is just that this has a weird impact on the rest of the logic and may need some type-fu to do properly.

TalDerei:
> I would actually much rather just switch this to using a full boolean decomposition. This was brought up earlier in the zkSecurity audit: we'll just avoid possible non-canonical decompositions by requiring the highest bit to be unset, which will force the prover to grind.
> 
> The problem is just that this has a weird impact on the rest of the logic and may need some type-fu to do properly.

recalling that thread now where we discussed the output distribution of `extract`. what if we kept the existing QR based extraction and patched this zero case (since it's already PR'd to close the immediate hole in the primitive)? and then followed up with a canonical boolean decomposition impl, which may even be more efficient in terms of constraints so we can treat that as an optimization vector?

filed https://github.com/tachyon-zcash/ragu/issues/765.

ebfull:
It won't be more efficient to "patch" this, I would rather leave it as-is and document the soundness gap. But it is almost as efficient to just fully decompose anyway, which means the only downside is the "grinding" the prover has to do to avoid canonicity checks.

TalDerei:
> It won't be more efficient to "patch" this, I would rather leave it as-is and document the soundness gap. But it is almost as efficient to just fully decompose anyway, which means the only downside is the "grinding" the prover has to do to avoid canonicity checks.

yeah fair enough, documented in https://github.com/tachyon-zcash/ragu/pull/766.

TalDerei:
closing in favor of #766.

PR #760, opened 2026-06-02
Circuit code documentation policy

ebfull:
Inspired by review of #740 I started looking for methodical changes for the documentation policy for handling exceptional logic cases.

PR #767, opened 2026-06-03
chore(primitives): endoscalar boolean decomposition

### LINKED ISSUE #765: Replace QR based endoscalar extraction with a canonical boolean decomposition
References  https://github.com/tachyon-zcash/ragu/pull/764, possibly premature vector for now.

`Endoscalar::extract` is used to [generate](https://github.com/tachyon-zcash/ragu/issues/71) verifier challenges, but the output distribution of extract is _somewhat_ speculative. For context, there was an OOB discussion where I noted:

"In terms of cross-curve challenge derivation, we’ll extract 128-bits out of the challenges and endoscale group elements to get scalars on the other curve (recall both curves will share the same Fp-transcript). we also can’t instead naively use a field isomorphism because the primes differ obviously, and the reduction would require non-native arithmetic.

anyways, generally for QR testing, not sure what that joint distribution looks like there but isn’t the property we’re targeting unpredictability/indistinguishability rather than strictly uniformity.

noticed the [method](https://github.com/tachyon-zcash/ragu/blob/main/crates/ragu_primitives/src/endoscalar.rs#L231-L238) mentions “a random output of a secure algebraic hash function" but doesn't characterize the output distribution."

and @ebfull had mentioned that he had convinced himself at one point it had negligible distance from uniform and would be safe, but it requires confirming. 

Anyways, with our current QR based approach, it costs a sqrt/QR op per bit and has the zero case hole (that requires 128 extra gates to patch), and the distribution is conjectured to be near uniform. 

In the bit decomposition approach, it'd be a standard boolean decomposition gadget where we'd decompose the challenge into one bit fewer than the field width (it's canonical by construction so doesn't require a range check to avoid wrapping). This is _possibly_ cheaper in terms of constraints because it avoid square roots, and the statistical analysis is easier to reason about.

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/765. Cross-linking to audit findings [here](https://zkao.io/projects/cmpvhafwe001u01jr5wejqxni/findings?finding=cmpvufkq60057o6jwgod0bwmi).

Replaces the QR based `Endoscalar::extract` with a canonical boolean decomposition, closing the gap demonstrated in #761 (CI failure and the Lean kernel unable to discharge the goal, and once this lands, we can add a new formal definition in Clean for this gadget).

This decomposes a challenge into its `CAPACITY` (254) LE bits and takes the low 128 bits as the endoscalar. `decompose` internally does a binary expansion on the allocated bits and enforces equality against the element using the same LC technique as `multipack`. It's a canonical decomposition since we're decomposing into one fewer bit than the field width, so we don't need a range check and worry about wrapping the prime field.

A retry loop is baked into `EndoscalarChallenge::sample` which resamples the eval blinding and re-derives `pre_beta` until it's in range.

### DIFF
--- a/crates/ragu_pcd/src/fuse/_09_eval.rs
+++ b/crates/ragu_pcd/src/fuse/_09_eval.rs
@@ -9,14 +9,10 @@ use ragu_circuits::{
     polynomials::{Rank, sparse},
     staging::StageExt,
 };
-use ragu_core::{
-    Result,
-    drivers::{Driver, DriverTypes},
-    maybe::{Always, Maybe},
-};
+use ragu_core::{Result, drivers::Driver, maybe::Maybe};
 use ragu_primitives::{Element, EndoscalarChallenge, GadgetExt, Point};
 
-use super::{NativeSPrime, RegistryWy};
+use super::{NativeFuseEmulator, NativeSPrime, RegistryWy};
 use crate::{
     Application, Proof,
     internal::{native, transcript::Transcript},
@@ -76,21 +72,19 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
     /// matching `eval_rx` and advanced transcript are returned together. The old
     /// prefix is therefore unavailable to callers after sampling, so subsequent
     /// transcript use must explicitly continue from the accepted transcript.
-    pub(super) fn sample_pre_beta<'dr, D, RNG>(
+    pub(super) fn sample_pre_beta<'dr, RNG>(
         &self,
         rng: &mut RNG,
-        dr: &mut D,
-        transcript_prefix: Transcript<'dr, D, C::CircuitPoseidon>,
+        dr: &mut NativeFuseEmulator<C>,
+        transcript_prefix: Transcript<'dr, NativeFuseEmulator<C>, C::CircuitPoseidon>,
         eval_witness: &native::stages::eval::Witness<C::CircuitField>,
         builder: &ProofBuilder<'_, C, R>,
     ) -> Result<(
-        EndoscalarChallenge<'dr, D>,
+        EndoscalarChallenge<'dr, NativeFuseEmulator<C>>,
         sparse::Polynomial<C::CircuitField, R>,
-        Transcript<'dr, D, C::CircuitPoseidon>,
+        Transcript<'dr, NativeFuseEmulator<C>, C::CircuitPoseidon>,
     )>
     where
-        D: Driver<'dr, F = C::CircuitField, Wire = ()>,
-        D: DriverTypes<MaybeKind = Always<()>>,
         RNG: CryptoRngCore,
     {
         // Rejection-sample only the eval-stage blinding. This is the final
--- a/crates/ragu_pcd/src/fuse/_09_eval.rs
+++ b/crates/ragu_pcd/src/fuse/_09_eval.rs
@@ -4,20 +4,13 @@
 //! of every element that was also queried in the `query` stage. The evaluation
 //! $f(u)$ is derived from the aforementioned evaluations.
 
-use ragu_arithmetic::{CryptoRngCore, Cycle, ff::Field};
-use ragu_circuits::{
-    polynomials::{Rank, sparse},
-    staging::StageExt,
-};
-use ragu_core::{Result, drivers::Driver, maybe::Maybe};
-use ragu_primitives::{Element, EndoscalarChallenge, GadgetExt, Point};
+use ragu_arithmetic::Cycle;
+use ragu_circuits::polynomials::Rank;
+use ragu_core::{drivers::Driver, maybe::Maybe};
+use ragu_primitives::Element;
 
-use super::{NativeFuseEmulator, NativeSPrime, RegistryWy};
-use crate::{
-    Application, Proof,
-    internal::{native, transcript::Transcript},
-    proof::ProofBuilder,
-};
+use super::{NativeSPrime, RegistryWy};
+use crate::{Application, Proof, internal::native, proof::ProofBuilder};
 
 impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_SIZE> {
     pub(super) fn compute_eval<'dr, D>(
@@ -50,57 +43,5 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
                 registry_xy: builder.native_registry_xy_poly().eval(u),
             },
         }
-    }
-
-    /// Rejection-samples the eval-stage blinding until the derived `pre_beta`
-    /// is an in-range endoscalar challenge.
-    ///
-    /// This consumes the transcript prefix immediately before the eval
-    /// commitment. Each attempt clones that prefix, absorbs a candidate eval
-    /// commitment, and squeezes a candidate `pre_beta`; on acceptance, the
-    /// matching `eval_rx` and advanced transcript are returned together. The old
-    /// prefix is therefore unavailable to callers after sampling, so subsequent
-    /// transcript use must explicitly continue from the accepted transcript.
-    pub(super) fn sample_pre_beta<'dr, RNG>(
-        &self,
-        rng: &mut RNG,
-        dr: &mut NativeFuseEmulator<C>,
-        transcript_prefix: Transcript<'dr, NativeFuseEmulator<C>, C::CircuitPoseidon>,
-        eval_witness: &native::stages::eval::Witness<C::CircuitField>,
-        builder: &ProofBuilder<'_, C, R>,
-    ) -> Result<(
-        EndoscalarChallenge<'dr, NativeFuseEmulator<C>>,
-        sparse::Polynomial<C::CircuitField, R>,
-        Transcript<'dr, NativeFuseEmulator<C>, C::CircuitPoseidon>,
-    )>
-    where
-        RNG: CryptoRngCore,
-    {
-        // Rejection-sample only the eval-stage blinding. This is the final
-        // commitment before `pre_beta`, so changing it gives a fresh challenge
-        // without rebuilding the preceding fuse state. The grind lives inside
-        // `EndoscalarChallenge::sample`, which (re)derives `pre_beta` from each
-        // fresh blinding and retries until it lands in range.
-        let (pre_beta, (eval_rx, accepted_transcript)) = EndoscalarChallenge::sample(dr, |dr| {
-            // Fresh eval-stage blinding each attempt: this draw is the
-            // per-attempt entropy that makes `pre_beta` independent across
-            // retries.
-            let eval_rx = native::stages::eval::Stage::<C, R, HEADER_SIZE>::rx(
-                C::CircuitField::random(&mut *rng),
-                eval_witness,
-            )?;
-            let native_eval_commitment = eval_rx.commit_to_affine(C::host_generators(self.params));
-            let bridge_eval_commitment =
-                builder.candidate_bridge_eval_commitment(native_eval_commitment)?;
-
-            let mut candidate_transcript = transcript_prefix.clone();
-            let eval_commitment = Point::constant(dr, bridge_eval_commitment)?;
-            eval_commitment.write(dr, &mut candidate_transcript)?;
-            let pre_beta = candidate_transcript.challenge(dr)?;
-
-            Ok((pre_beta, (eval_rx, candidate_transcript)))
-        })?;
-
-        Ok((pre_beta, eval_rx, accepted_transcript))
     }
 }
--- a/crates/ragu_pcd/src/fuse/_09_eval.rs
+++ b/crates/ragu_pcd/src/fuse/_09_eval.rs
@@ -9,11 +9,19 @@ use ragu_circuits::{
     polynomials::{Rank, sparse},
     staging::StageExt,
 };
-use ragu_core::{Result, drivers::Driver, maybe::Maybe};
-use ragu_primitives::Element;
+use ragu_core::{
+    Result,
+    drivers::{Driver, DriverTypes},
+    maybe::{Always, Maybe},
+};
+use ragu_primitives::{Element, EndoscalarChallenge, GadgetExt, Point};
 
 use super::{NativeSPrime, RegistryWy};
-use crate::{Application, Proof, internal::native};
+use crate::{
+    Application, Proof,
+    internal::{native, transcript::Transcript},
+    proof::ProofBuilder,
+};
 
 impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_SIZE> {
     pub(super) fn compute_eval_witness<'dr, D>(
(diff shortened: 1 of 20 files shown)


crates/ragu_primitives/src/endoscalar.rs
TalDerei, line 1: @ebfull in terms of the grinding the prover has to do to avoid canonicity checks, should we if wrap the whole fuse attempt in a retry loop if for instance `pre_beta >= 2^CAPACITY`, or another idea which seems more surgical is to resample the eval stage commitment randomness until `pre_beta is < 2^CAPACITY`? there's probably a preferred way to do this rejection sampling in our current fuse pipeline.

TalDerei:
there's still more work to be done here, but opening it up for initial review comments.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -278,6 +278,42 @@ pub fn multipack<'dr, D: Driver<'dr, F: ragu_arithmetic::ff::PrimeField>>(
...
+///
+/// Returns `D::F::CAPACITY` booleans constrained so that
+/// `b_0 + b_1*2 + b_2*4 + ... == elem`. This is the inverse of [`multipack`].
+///
+/// `CAPACITY` bits represent exactly the values in `[0, 2^CAPACITY)`, a subset
+/// of `[0, p)`, so the decomposition is canonical by construction: every
+/// in-range element has a unique bit string and no wrapped alias is
+/// representable, requiring no range check. The trade-off is that the
+/// constraints are only satisfiable when `elem < 2^CAPACITY`; for the Pasta
+/// fields the excluded range has negligible density (about `2^-129`).
```
ebfull: Heads up: this doc buries the completeness condition which is worth being explicit about. I've tried in #760 to create policies that better match what I expect to see in doc strings like this.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -75,89 +74,43 @@ impl<'dr, D: Driver<'dr>> Endoscalar<'dr, D> {
...
+    /// The endoscalar is the low 128 bits of the element's canonical bit
+    /// decomposition. The element is decomposed in full, so those 128 bits are
+    /// uniquely determined by `elem` and the extraction is sound for every
+    /// input.
     ///
-    /// [#765]: https://github.com/tachyon-zcash/ragu/issues/765
+    /// The decomposition only admits elements below `2^CAPACITY`, so for the
+    /// negligible (~2^-129) fraction of larger elements the constraint is
+    /// unsatisfiable and an honest prover cannot prove. This out-of-range abort
+    /// is intentional and acceptable for a transcript-derived challenge.
```
ebfull: Ditto for these docs.

crates/ragu_primitives/src/endoscalar.rs
ebfull: It's tricky... I think we need a type for a Poseidon output (which we know, if the input is random within the prover's control, will not violate this precondition with high probability). The tricky part is somehow embedding this quality (the influence the prover has over the distribution of the output) into the APIs and type system. I want it to be clear when we're relying on this precondition and trace it back to what causes it to be true.

Anything that appears fragile is not going to work though.

ebfull: Ah, I know what to do: create a type that forces the prover to emulate first before they're allowed to access the type, and have it emulate the decomposition. I think a really simple way to do this is to use routines! Then, the API can force the caller to rewrite their code (as a routine) so that the retry loop has to be unified with the code that produces the value.

TalDerei: interesting idea, that seems to work! will mock out an API shape.

crates/ragu_primitives/src/boolean.rs
TalDerei: specified the completeness conditions in an explicit `# Completeness` section.

TalDerei:
**update:** still playing around with the retry loop mechanics.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -13,25 +13,258 @@
...
+    ///
+    /// # Errors
+    ///
+    /// A genuine synthesis failure in `produce` or in validating an in-range
+    /// candidate propagates immediately; the loop retries only on the expected
+    /// out-of-range condition and so cannot spin on a real error.
+    pub fn sample<T>(
+        dr: &mut NativeEmulator<F>,
+        mut produce: impl FnMut(&mut NativeEmulator<F>) -> Result<(Element<'dr, NativeEmulator<F>>, T)>,
+    ) -> Result<(Self, T)> {
```
TalDerei, line 230: `sample` returns challenge an `Element<'dr, NativeEmulator<F>>` (generic field element gadget), but in the future it could be a dedicated typed challenge (reference https://github.com/tachyon-zcash/ragu/pull/569).

TalDerei:
status quo hasn't changed, staging for review.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -294,18 +487,36 @@ pub fn lift_endoscalar<F: WithSmallOrderMulGroup<3>>(endo: u128) -> F {
...
+/// [`EndoscalarChallenge`] validation fail before [`Endoscalar::extract`] is
+/// reachable.
 ///
-/// Given a random output of a secure algebraic hash function, this extracts
-/// `k` bits of "randomness" from the value by checking whether `value + i`
-/// is a quadratic residue for each bit position `i`.
-pub fn extract_endoscalar<F: PrimeField + WithSmallOrderMulGroup<3>>(value: F) -> u128 {
+/// # Panics
+///
+/// Panics if `value` has not passed endoscalar challenge validation.
```
ebfull: I don't think we should be exposing a function like this in the API that will panic on some inputs, when we could just return an error or optional and locally unwrap in the places where the invariant is guaranteed.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -306,6 +306,47 @@ pub fn multipack<'dr, D: Driver<'dr, F: ragu_arithmetic::ff::PrimeField>>(
...
+/// of `[0, p)`, so the decomposition is canonical by construction: every
+/// in-range element has a unique bit string and no wrapped alias is
+/// representable, requiring no range check.
+///
+/// # Completeness
+///
+/// Honest proving succeeds only when `elem < 2^CAPACITY`; a larger element has
+/// no `CAPACITY`-bit decomposition, so the constraints are unsatisfiable. Over
+/// the Pasta fields a uniformly random element falls outside this range with
+/// negligible probability (about `2^-129`).
```
ebfull: Much of the doc comments _should_ be using math formulas / katex by policy.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -13,25 +13,258 @@
...
+    boolean::decompose,
     promotion::Demoted,
     vec::{CollectFixed, ConstLen, FixedVec},
 };
 
+/// A transcript challenge pre-validated for endoscalar extraction.
+///
+/// Carries the precondition required by [`Endoscalar::extract`]: the element's
+/// canonical representative is below `2^CAPACITY`, so it admits a canonical
+/// `CAPACITY`-bit decomposition with no separate in-circuit canonicity check.
```
ebfull: There's something wrong with this API and it has me concerned.

It implements `Gadget`, but it offers this contract imposing an invariant over its witness. But gadgets aren't supposed to do that _ever_. You need to change this, it shouldn't be a `Gadget` at all. It's certifying that it's an `Element` with some specific quality that the prover relies on... just create a newtype for that.

ebfull (changes requested):
EndoscalarChallenge violates my understanding of the gadget API.

crates/ragu_primitives/src/endoscalar.rs
TalDerei: yeah this is a gadget API/policy violation, but not a soundness issue (because the unified instance binds the challenge, and i made sure that `Endoscalar::extract` explicitly enforces its range / decomposition). the context here that motivated the design was https://github.com/tachyon-zcash/ragu/pull/767#discussion_r3357012696, so it's a gadget purely so it could be returned from the routine and passed across drivers as a typed challenge.

TalDerei: agreed OOB that routines here are unnecessary, they've been removed.

TalDerei: opted to make this a newtype rather than trying to keep the gadget and fix the policy violation (note the invariant you referenced is checked elsewhere in-circuit, this doc comment just made an overstatement that it was checked here).

crates/ragu_primitives/src/boolean.rs
TalDerei: fixed, aligned with our policy.

crates/ragu_primitives/src/boolean.rs
```diff
@@ -304,6 +304,57 @@ pub fn multipack<'dr, D: Driver<'dr, F: ragu_arithmetic::ff::PrimeField>>(
...
+    dr: &mut D,
+    allocator: &mut impl Allocator<'dr, D>,
+    elem: &Element<'dr, D>,
+) -> Result<Vec<Boolean<'dr, D>>> {
+    let repr = elem.value().map(|v| v.to_repr());
+    let bits = (0..D::F::CAPACITY as usize)
+        .map(|i| {
+            let bit = repr
+                .as_ref()
+                .map(move |repr| (repr.as_ref()[i / 8] >> (i % 8)) & 1 == 1);
```
ebfull: `PrimeField::to_repr` does not specify an endianness; little-endian byte order is a property of the pasta_curves implementation we use, not a contract, which is why it has to be restated as a field requirement here and in `endoscalar_in_range` and `extract_endoscalar`. `PrimeFieldBits::to_le_bits` provides the same bits with the order guaranteed by the trait. Suggest requiring `PrimeFieldBits` and iterating `to_le_bits` for the witness bits. That deletes the field-requirement paragraphs and turns misuse into a type error instead of a runtime surprise. It needs the ff `bits` feature; bitvec is already in Cargo.lock, so the dependency tree does not grow. The bound will surface on `Cycle::CircuitField` via compute_v; I would mirror it on `ScalarField` so the two field bounds stay symmetric.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -13,25 +13,250 @@
...
+pub fn endoscalar_in_range<F: PrimeField>(value: F) -> bool {
+    let repr = value.to_repr();
+
+    // Recompose the low CAPACITY bits exactly as `decompose` does, then compare
+    // against the original value. The two agree iff no bit at index >= CAPACITY
+    // is set, i.e. iff `value < 2^CAPACITY`.
+    let mut acc = F::ZERO;
+    let mut gain = F::ONE;
+    for i in 0..(F::CAPACITY as usize) {
+        if (repr.as_ref()[i / 8] >> (i % 8)) & 1 == 1 {
```
ebfull: Related to the `to_le_bits` suggestion on `decompose`: the consequence of a wrong endianness assumption is worst here. On a field whose repr is not little-endian, this predicate is false for essentially every element, so `sample` rejects candidates indefinitely instead of failing loudly. This is also the third copy of the `repr.as_ref()[i / 8] >> (i % 8)` indexing (the `decompose` witness closure and the value fold in `extract_element` are the others). With `to_le_bits` the predicate reduces to checking that no bit at index >= CAPACITY is set, and the three copies collapse into one.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -13,25 +13,250 @@
...
+        D::F: PrimeField,
+    {
+        // Emulator drivers never evaluate the decomposition constraints, so
+        // also reject an out-of-range witness value directly; `try_just` runs
+        // only during witness generation and emits nothing, leaving circuit
+        // structure untouched.
+        D::try_just(|| {
+            if !endoscalar_in_range(*elem.value().take()) {
+                return Err(Error::InvalidWitness(
+                    "endoscalar challenge must satisfy value < 2^CAPACITY".into(),
```
ebfull: Out-of-range is the one failure of this constructor that callers need to detect programmatically, since the correct response is resampling rather than propagation, and a Display string is not a stable thing to match on. Suggest a dedicated public unit error type (`EndoscalarRangeError` or similar) boxed as the source here, plus a small accessor on `ragu_core::Error`, something like `invalid_witness_source::<T>() -> Option<&T>`, so callers do not downcast through the variant by hand. The next two comments build on this.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -13,25 +13,250 @@
...
+    /// [`sample`]: EndoscalarChallenge::sample
+    /// [`from_element`]: EndoscalarChallenge::from_element
+    pub(crate) fn try_from_element(
+        dr: &mut NativeEmulator<F>,
+        elem: Element<'dr, NativeEmulator<F>>,
+    ) -> Result<Option<Self>> {
+        // Classify the candidate by value first, so an out-of-range challenge
+        // never reaches the binding decomposition (which would conflate it with
+        // a genuine error). Only in-range challenges are constrained.
+        if !endoscalar_in_range(**elem.value().snag()) {
```
ebfull: With a typed source on the range failure, this pre-classification can go away: call `from_element` and translate the range error into `Ok(None)`, propagating everything else. The conflation this comment guards against no longer exists, because the error itself identifies the condition on any driver. That leaves `from_element` as the single place the range rule is checked, and `endoscalar_in_range` becomes an implementation detail of it.

crates/ragu_primitives/src/lib.rs
```diff
@@ -37,7 +37,12 @@ pub mod vec;
 
 pub use boolean::{Boolean, multipack};
 pub use element::{Element, multiadd};
-pub use endoscalar::{Endoscalar, extract_endoscalar, lift_endoscalar};
+// `endoscalar_in_range` is an internal predicate; it is re-exported only for the
+// fuzz harness (which models challenge rejection) under `unstable-fuzzing`, so it
+// does not widen the normal public API.
+#[cfg(feature = "unstable-fuzzing")]
```
ebfull: With the typed range error available, the fuzz target can model transcript rejection by calling `extract_endoscalar` and discarding inputs that fail with that specific error. That exercises the same classification path production code uses, instead of a pre-filter that has to be kept in agreement with it, and this feature and re-export can then be dropped.

crates/ragu_primitives/src/endoscalar.rs
```diff
@@ -294,18 +472,43 @@ pub fn lift_endoscalar<F: WithSmallOrderMulGroup<3>>(endo: u128) -> F {
     acc
 }
 
-/// Extracts an endoscalar from a random field element.
+/// Extracts an endoscalar from a validated field element.
```
ebfull, line 484: This is a protocol-level change to what extraction means, not just an implementation change: extraction is now defined only below 2^CAPACITY, and the prover re-derives the challenge on a miss. `book/src/protocol/extensions/endoscalar.md` still specifies extract(s in F) as total, with a pseudorandom-extraction property over all of F. Please update that page to cover the range restriction, the low-bits definition, and the prover's re-randomization obligation. The extraction property can then be stated as exact uniformity over the in-range subset.

crates/ragu_pcd/src/fuse/_09_eval.rs
```diff
@@ -43,14 +45,31 @@ impl<C: Cycle, R: Rank, const HEADER_SIZE: usize> Application<'_, C, R, HEADER_S
...
+
+    /// Samples a fresh eval-stage blinding and returns its rx together with the
+    /// bridge eval commitment.
+    ///
+    /// The `pre_beta` rejection loop calls this once per attempt, re-deriving
+    /// the challenge from a fresh commitment until it lands in range. Only the
+    /// accepted rx is committed to the builder (via `set_native_eval_rx`), so
+    /// this computes the commitment without populating the builder's cached eval
+    /// cells.
+    pub(super) fn sample_eval_commitment<RNG: CryptoRngCore>(
```
ebfull: Non-blocking. On the accept path these commitments are computed twice: once here for classification, and again later when the builder's lazy caches derive them from the stored rx (`native_eval_commitment` and the cached bridge eval pair). Everything is deterministic and the shared macro keeps the derivations from drifting, so it is only duplicate work once per fuse. If a variant of `set_native_eval_rx` that also stores the accepted commitments is cheap to add, it would remove that; fine to leave as is otherwise.

ebfull (changes requested):
The core of this looks right. A CAPACITY-bit decomposition is canonical without a separate range check because 2^CAPACITY < p, so constraining pre_beta to its decomposition and taking the low 128 bits closes the #765 gap, and putting the rejection loop inside `EndoscalarChallenge::sample` means native code cannot obtain a challenge from a rejected candidate. The boundary tests at 2^CAPACITY - 1 and 2^CAPACITY, the wireless-emulator rejection test, and sharing the rx derivation between the candidate and cached bridge eval paths are all good.

Three requests, detailed inline:

1. Build the bit access on `PrimeFieldBits::to_le_bits` instead of `to_repr`, so the little-endian assumption becomes a trait contract rather than a documented field requirement. This is the main one.
2. Report the out-of-range failure with a typed error source instead of a Display string. This also simplifies `try_from_element` and removes the need for the `unstable-fuzzing` feature.
3. Update the endoscalar page in the book, which still specifies extraction as total on F.

None of these are soundness problems on the Pasta fields. They are about making the assumptions this change relies on explicit and checkable.

crates/ragu_primitives/src/endoscalar.rs
TalDerei: yeah agreed, making this explicitly clear (lifting from your book entry in the [packable](https://github.com/tachyon-zcash/ragu/compare/main...ebfull:ragu:packable) branch).

crates/ragu_pcd/src/fuse/_09_eval.rs
TalDerei: filed https://github.com/tachyon-zcash/ragu/issues/838 so it doesn't block this PR.

TalDerei:
@v12sec review this

TalDerei:
@v12sec review this

TalDerei:
> **Complete:** Audit complete. V12 found one issue worth reviewing.

non-blocking; flags any issue for smaller fields than ragu's field modulus.

TalDerei:
working on FV impl of this gadget in https://github.com/tachyon-zcash/ragu/pull/848.

ebfull (approved):

PR #768, opened 2026-06-03
chore(pcd): restrict base-case detection to the bootstrap step

### LINKED ISSUE #827: Sentinal for base case detection
References https://github.com/tachyon-zcash/ragu/pull/768: "Anyways, I think base case detection should use a special sentinel (a reserved internal header suffix distinct from (), or even stronger maybe pair the sentinel with a fixed internal circuit id or proof kind flag), or have it be wired to only work under seed path?"

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/827. I'm contending that our escape hatch mechanism for base case detection is too broad. The motivation is to impose a restriction confining the base case to only the leaves of the PCD graph (rather than allowing it at arbitrary internal nodes as before). Without this, we seem to muddy any security analysis we'd want to do.

-----------------------

**Some supplemental historical context:**

Analyzed our approach for bootstrapping recursion from first principles. For context, lifting a thread from an earlier OOB async discussion that documents the landscape of design decisions around the trivial proof API and base case handling:

* 1. Trivial proofs that verify (partial status quo): instead of duplicating logic, trivial proofs would be minimal (zero polynomials, deterministic blindings) but still verify (somehow?),
* 2. Trivial proofs skip verification: the recursive circuit includes a base case selector that detects a trivial header and conditionally skips the verification check,
* 3. Remove trivial from API: the base case becomes implicit; rather than exposing a trivial method, we’d introduce some internal base_step(witness) —> PCD that creates minimal trivial proof, runs merge like a leaf step, and returns a non-trivial proof that does verify (~ rerandomization),
* 4. Proof enum: distinguish between trivial and real variants, with special handling for trivial proofs during merge,
* 5. Cached trivial + multiple merge passes: application caches one trivial proof at startup and runs merge multiple times for sufficient randomization,

For reference, [ark-pcd](https://github.com/arkworks-rs/pcd) takes approach (2): their dummy proofs are real proofs of a trivial circuit, and instead of making trivial proofs verify, verification is conditionally skipped via a base case selector inside the recursive circuit.

Our approach (which was implemented by https://github.com/tachyon-zcash/ragu/pull/279) models a strictly _weaker_ approach to (2). Strictly weaker in the sense that there's some conceptual ambiguity with the way we're interpreting our base case detection mechanism I think. 

Currently, suffix 1 is defined as the internal suffix for the () header. The base case triggers when both child headers are () in the PCD tree, not only during _initial_ bootstrapping. That's confusing, no? Because we need some way to bootstrap the recursion, but the problem is the predicate chosen for that exception _seems_ too broad. If suffix is used, it ideally should be a dedicated internal base case sentinel that normal application outputs can't produce, not the reusable () header suffix. In that context, currently any recursive fuse whose left and right input header types are () gets base case treatment. 

https://github.com/tachyon-zcash/ragu/pull/267#pullrequestreview-3608409505 anticipated that base cases will need to accept dummy/invalid bootstrap children (trivial PCDs that can't verify on their own), selected by a child header suffix. But we're also allowing other PCD proofs in the chain to trigger the selector.

There's a flow that demonstrates this:
  1. Start with two trivial `Pcd<()>` inputs,
  2. seed/fuse them through a step whose `Output = ()`, producing a valid `Pcd<()>`,
  3. Corrupt that produced proof so it fails verify on its own,
  4. Fuse two corrupted `Pcd<()>` through another step with `Left = () `and `Right = ()`,
  5. The parent proof verifies.

Also rerandomize where left/input header's are () also currently triggers the same base case predicate.


crates/ragu_pcd/src/internal/native/stages/preamble.rs
```diff
@@ -138,14 +143,19 @@ impl<'dr, D: Driver<'dr, F = C::CircuitField>, C: Cycle, const HEADER_SIZE: usiz
...
         &self,
         dr: &mut D,
         allocator: &mut impl Allocator<'dr, D>,
     ) -> Result<Boolean<'dr, D>> {
-        let suffix = &self.output_header[HEADER_SIZE - 1];
-        suffix.is_equal(dr, allocator, &Element::one())
+        let bootstrap = Element::constant(dr, D::F::from(Suffix::internal(2).get()));
+        self.output_header[HEADER_SIZE - 1].is_equal(dr, allocator, &bootstrap)
     }
 }
```
TalDerei: basically, base case detection now checks for a private internal `Dummy` header (only the `Bootstrap` step can declare it).

crates/ragu_pcd/src/lib.rs
```diff
@@ -171,14 +206,65 @@ impl<'params, C: Cycle, R: Rank, const HEADER_SIZE: usize>
...
+            &mut rng,
+            step::internal::bootstrap::Bootstrap::new(),
+            (),
+            app.dummy_pcd(),
+            app.dummy_pcd(),
+        )?;
+        app.bootstrap = Some(pcd.into_parts().0);
+
+        Ok(app)
+    }
```
TalDerei, line 253: this seems like a nicer internal symmetry where finalize owns bootstrapping: an internal `Bootstrap` step fuses two dummy `Pcd<Dummy>` inputs (where `Dummy` is a reserved sentinel), produces / caches a valid `Pcd<()>`, and supplies it as verified children to `seed`. for one extra fuse during finalization, we don't allow arbitrary base case triggers in the PCD graph (specifically, only this internal `Bootstrap` step can trigger the base case, beneath the leaves of the application graph).

see tree visualization in https://claude.ai/code/artifact/43b623b1-e9f1-4217-9f2e-b320a47303b6 (there's a self-fold rerandomization optimization there that would save caching an uncompressed trivial PCD proof. **update: rerandomization isn't touched anymore**).

book/src/implementation/proofs.md
```diff
@@ -114,20 +114,21 @@ let (proof, aux) = app.seed(&mut rng, MyLeafStep { ... }, witness)?;
...
+once over two synthesized _trivial proofs_ — which do not verify on their own —
+to produce the bootstrap proof, which carries the reserved `Leaf` header.
 
-Trivial proofs are not meant to verify independently—they exist
-solely to provide valid input structure for `seed()` when no real
-child proofs are available.
+The bootstrap proof attests nothing, and any prover can produce one; it exists
+only so that the recursion has a valid leaf to start from. No other step may
+produce `Leaf`, and a step may declare it for both inputs or neither, so the
+base case sits only beneath the leaves of the graph.
```
TalDerei: for seed calls, we now instantiate with valid `Pcd<Leaf>` instead of `Pcd<()>`.

crates/ragu_pcd/src/internal/native/stages/preamble.rs
TalDerei, line 160: basically, base case detection now uses a dedicated private `Dummy` header (declared by internal Bootstrap step).

seems more elegant than overloading ordinary () application headers. before, any application step with both trivial headers would trigger base case detection; now it's confined to internal dummy headers for a special internal bootstrap step.

TalDerei:
> There's a flow that demonstrates this:
> 
> 1. Start with two trivial `Pcd<()>` inputs,
> 2. seed/fuse them through a step whose `Output = ()`, producing a valid `Pcd<()>`,
> 3. Corrupt that produced proof so it fails verify on its own,
> 4. Fuse two corrupted `Pcd<()>` through another step with `Left = () `and `Right = ()`,
> 5. The parent proof verifies.

the crux of what this PR prevents where the parent proof no longer verifies.

TalDerei:
@v12sec review this

.github/workflows/rust.yml
```diff
@@ -180,10 +180,10 @@ jobs:
         uses: ./.github/actions/rust-setup
         with:
           targets: i686-unknown-linux-gnu
-          apt-packages: gcc-multilib
+          cargo-tools: cross@0.2.5
           cache-suffix: release
       - name: Run 32-bit tests
-        run: cargo test --release --all --locked --target i686-unknown-linux-gnu
+        run: cross test --release --all --locked --target i686-unknown-linux-gnu
```
TalDerei: because apt regularly hangs CI, but we may want to revert.

TalDerei:
@v12sec review this

LucidSamuel:
Hey @TalDerei I checked the other cached value in this path, and `seeded_trivial_pcd` still looks problematic. It is initialized once from the caller’s RNG, stored in the OnceCell, and reused as the right child of every rerandomization. fuse then copies its seven stage polynomials into each parent and embeds its commitments in the bridge preamble.

probed and observed that outputs from the same Application matched in all 7 right-child stage polynomials and all 30 embedded commitment coordinates, even after rerandomizing twice while Independently seeded applications matched 0/7 and 0/30.

I'm thinking a fixed, domain-separated seed... possibly initialized beside the bootstrap would make this a public constant instead. 

~ just an observation as I think this will only may matter once Tachyon’s global PROOF_SYSTEM uses real Ragu, though I haven’t established whether compression preserves the tag.

ebfull (approved):

PR #780, opened 2026-06-05
feat(ragu): port mock ragu

TalDerei:
This ports the mock ragu API built by @turbocrime from [tachyon](https://github.com/tachyon-zcash/tachyon/tree/main/crates/mock_ragu) into the ragu crate (workspace root) so that it can be integrated with Zebra. The migration targets mock ragu on current tachyon main, and cherry picks other polynomial / framework hooks API changes made to mock in https://github.com/tachyon-zcash/tachyon/pull/124 (reference PR description and closing issue https://github.com/tachyon-zcash/tachyon/issues/130 for context). The manifest dep changes are minimal and purely additive. 

There are obvious divergences that break parity (primarily in the sense here that the [framework hooks](https://github.com/tachyon-zcash/ragu/pull/701) are currently a fabricated surface and there's no ragu counterpart, _yet_). I'll flag some examples of these divergences below, but we could either keep those changes, which are encapsulated in the [second commit](https://github.com/tachyon-zcash/ragu/pull/780/commits/9d891ec5780f14f7e57a6784bccf9cd2e5894a94), or abandon them as premature until actual polynomial oracle access support lands with an agreed shape, and then we can align the mock against those changes. The direction we take requires consensus before landing this migration.


src/tests/application.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    fn witness<'source>(
+        &self,
+        _ctx: &mut StepCtx<'_>,
+        witness: Self::Witness<'source>,
+        _left: <Self::Left as Header>::Data<'source>,
+        _right: <Self::Right as Header>::Data<'source>,
+    ) -> Result<(<Self::Output as Header>::Data<'source>, Self::Aux<'source>)> {
+        Ok((TestHeaderData { value: witness }, ()))
+    }
```
TalDerei, line 51: this is one of the liberties the mock takes by anticipating the oracle query changes. real `Step::witness` takes a `Driver`, not a `ctx` object.

src/mock/polynomial.rs
```diff
@@ -0,0 +1,212 @@
...
+        for (i, &c) in coeffs.iter().enumerate() {
+            new_coeffs[i + 1] += c;
+            new_coeffs[i] += c * root.neg();
+        }
+        coeffs = new_coeffs;
+    }
+    coeffs
+}
+
+/// Mirrors `ragu_circuits::polynomials::unstructured::Polynomial`.
```
TalDerei, line 49: real poly commits are unblinded per https://github.com/tachyon-zcash/ragu/pull/621, but mock blinds them.

src/mock/application.rs
```diff
@@ -0,0 +1,148 @@
...
+    ) -> Result<(Pcd<S::Output>, S::Aux<'source>)> {
+        let left_proof = left.proof;
+        let right_proof = right.proof;
+
+        let mut hooks = FrameworkHooks::new();
+        let mut ctx = StepCtx::new(&mut hooks);
+        let (output_data, aux) = step.witness(&mut ctx, witness, left.data, right.data)?;
+
+        // TODO just like the real crate :D
+        let _claims = hooks.into_outputs();
```
TalDerei, line 104: the fact that we're not doing anything with the PCS opening claims and just discarding them maybe suggests these API additions are premature. but still good to mock some kind of shape...so it may actually be useful to keep.

src/mock/polynomial.rs
turbocrime: we can remove the blinding param, it's not used in any of the present iteration

src/mock/relations.rs
turbocrime, line 1: my local branch of mock_ragu has some newer work here, and is still different from https://github.com/tachyon-zcash/ragu/pull/701

TalDerei: can you optimistically upstream to this PR? treating these changes as speculatively landing in the port anyway until there's an agreement on the scope of this mock.

turbocrime: yeah, i'll push things over.

turbocrime (approved):
pulled my current working version into this PR.

it had lots of additional lifetime parameters, which i think you removed? or, they may have been added later. see da9bd5495c7182dd8649ea4d7e1e3037339abade

turbocrime:
i have some additional tests for the mock that i'll bring over in a minute

TalDerei:
> it had lots of additional lifetime parameters, which i think you removed?

these were GAT alignments against ragu in https://github.com/tachyon-zcash/ragu/pull/780/commits/15512a3d46091ecc424757c4981d3d3628c28f96. the lifetimes were removed and then re-added by your latest batch.

TalDerei:
we had also discussed possibly feature gating the mock API (right now the ragu crate is the mock unconditionally)?

**update:** was playing around with this in https://github.com/tachyon-zcash/ragu/commit/02f0c8a1aab7bca34313c11364f668662dc2e2a8 which we can optionally cherry-pick.

Cargo.toml
```diff
@@ -104,6 +99,7 @@ ff = { version = "0.14", default-features = false }
 group = { version = "0.14", default-features = false }
 pasta_curves = { git = "https://github.com/ebfull/pasta_curves", rev = "5a84fc0d13f738f54c43fb1f9e28ecf1d358dcf5", features = ["deferred"] }
 rand = "0.10"
+rand_core = { version = "0.10", default-features = false }
```
TalDerei, line 114: `rand_core` version currently [diverges](https://github.com/ZcashFoundation/zebra/blob/main/Cargo.toml#L117) from zebra, but as do other deps like `rand`. we'll eventually align dep versions. cargo `vet / audit` came back clean.

TalDerei:
latest commit properly gates mock against [legacy/modern stack](https://github.com/tachyon-zcash/ragu/pull/747). mock needs the modern crypto APIs, so the feature gates I decided on for the legacy / modern deps are:
- `--features modern-deps` where mock is included and uses modern deps,
- `--features legacy-deps` where mock is excluded from build and uses legacy deps

the mock is off by default, and zebra can pull it in by pulling the `mock` feature, and it only uses modern deps when enabled.

TalDerei:
optimistic ACK OOB by @ebfull; will merge to complete port and we can iterate on scope accordingly.

PR #781, opened 2026-06-05
chore(primitives): document gadget preconditions/contracts

### LINKED ISSUE #779: Validate floor plan invariants
Independent of the current floor plan construction, the wiring evaluators accept raw `ConstraintSegment` slices. They check local segment counts after synthesis, but if say a malformed floor plan reaches sx / sy / sxy, or trace assembly, technically invalid global layout assumptions can produce incorrect polynomial placement or panic via assertions far later in the call stack. It'd be nice to have more informative error handling here where we add floor plan validation at evaluator/assembly boundaries and return a structured error for invalid plans. We should maybe have a floor plan variant on the existing `Error` type?

TalDerei:
Closes https://github.com/tachyon-zcash/ragu/issues/779. Cross-linking to audit finding [here](https://zkao.io/projects/cmpvhafwe001u01jr5wejqxni/findings?finding=cmpxwba8000khpqjvf4lwe8na).

Wiring eval / trace assembly previously trusted segment offsets and could panic deep in the call stack. They now call `floor_planner::validate` up front before doing any work, and if validation fails they propagate a dedicated `MalformedFloorPlan` error instead of trusting the offsets or panicking later. Validation moves earlier in the call stack, and the actual boundaries where errors are emitted are now more compartmentalized. 

It internally propagates `Result`s, checks that `sx / sy / sxy` are equivalent, and only then returns the sparse polynomial itself.

**Update**: see [update](https://github.com/tachyon-zcash/ragu/pull/781#issuecomment-4651639674) below for the state of this PR.

### DIFF
--- a/crates/ragu_circuits/src/floor_planner.rs
+++ b/crates/ragu_circuits/src/floor_planner.rs
@@ -142,12 +142,16 @@ pub fn floor_plan(segment_records: &[SegmentRecord]) -> Vec<ConstraintSegment> {
         constraint_start += record.num_constraints();
     }
 
-    // Every circuit has a root segment, so `segment_records` (and therefore
-    // `result`) is always non-empty; an empty plan is a caller bug. The
-    // prefix-sum construction above is valid by construction, so this is a
-    // defensive self-check that also guards future floor planners that reorder
-    // or pack segments.
-    validate(&result).expect("floor planner must produce a valid floor plan");
+    // Hand-optimized mask `WiringObject`s (e.g. `StageMask`) have no segments
+    // and ignore the floor plan, so the registry legitimately calls this with
+    // empty `segment_records`; return the empty plan rather than tripping
+    // `validate`'s root-segment requirement. For non-empty plans the prefix-sum
+    // construction above is valid by construction, so this self-check is
+    // defensive and also guards future floor planners that reorder or pack
+    // segments.
+    if !result.is_empty() {
+        validate(&result).expect("floor planner must produce a valid floor plan");
+    }
 
     result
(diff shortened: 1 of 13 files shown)


crates/ragu_circuits/src/floor_planner.rs
```diff
@@ -73,6 +75,50 @@ pub struct ConstraintSegment {
     pub(crate) num_constraints: usize,
 }
 
+pub(crate) fn validate(floor_plan: &[ConstraintSegment]) -> Result<()> {
```
TalDerei: note that once we replace our current dummy DFS prefix sum impl, this method body will need to be updated as well.

TalDerei:
**update:** in discussing with @ebfull, this is one class of bugs that can be caught by using and propagating errors, but there are others that can't. we can't prevent all the different ways the circuit writer may exploit non-determinism (say with interior mutability, which our circuit API doesn't protect against since this is a limitation of the type system, as the [closing issue](https://github.com/tachyon-zcash/ragu/issues/779) already points out). this is particularly reminiscent of the observation made in https://github.com/tachyon-zcash/ragu/pull/610#discussion_r2977385683. so instead of these error handling changes we should more strictly document this as an invariant. 

maybe we should have a "determinism" subsection in the circuit code documentation policy https://github.com/tachyon-zcash/ragu/pull/760? if that inherits the scope here, this can close accordingly. **(This is what i opted for instead, minimal doc comment over code changes).**

TalDerei:
restaging for review.

ebfull (approved):

PR #783, opened 2026-06-06
feat(pcd): FrameworkHooks with derive_challenge

S1nus:
Moved derive_challenge out of #701 due to complexity.

adds just the circuit-facing API for derive_challenge and enforce_poly_query hooks, leaving the fuse plumbing and changes to the recursion circuits as followup work.


TalDerei:
@ebfull see review feedback in #701 for historical context, cc @ebfull

PR #789, opened 2026-06-09
fix: prevent application suffix wrapping

### LINKED ISSUE #776: Prevent suffix wrapping
Currently, if an application developer writes something like `Suffix::new(usize::MAX - 1)` in their `Application` circuit, it will overflow usize and violate our claim that application suffixes can't alias reserved internal suffixes. In that context, we can either add a const assert to `Suffix::new() `to guard against this (like the assertion in `Suffix::internal()`), or it becomes a non-issue when our [Application proc macro](https://github.com/tachyon-zcash/ragu/pull/585) changes land and we remove this suffix handling from the application circuit layer and abstract it as an internal implementation detail.

LucidSamuel:
fixes #776 

guards `Suffix::new()` against application suffix values that would overflow when the reserved internal suffix offset is added.

- changes

- Add a const-compatible assertion to `Suffix::new()`
- Document the panic condition
- Test the maximum valid application suffix
- Test that an overflowing suffix is rejected

### DIFF
--- a/crates/ragu_pcd/src/header.rs
+++ b/crates/ragu_pcd/src/header.rs
@@ -38,7 +38,18 @@ pub struct Suffix {
 
 impl Suffix {
     /// Creates a new application-defined [`Header`] suffix.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `value` is large enough that adding the internal suffix
+    /// offset would overflow `usize`, which would alias reserved internal
+    /// suffixes.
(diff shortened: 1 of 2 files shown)


TalDerei:
reloaded context here, seems to reference https://github.com/QED-it/ragu/issues/67.

TalDerei:
this becomes moot once #585's application macro owns suffix assignment, so closing as unplanned.

PR #563, opened 2026-03-12
chore: API hardening

TalDerei:
Inspired by https://github.com/tachyon-zcash/ragu/pull/541, further trait sealing and private fields with public accessors.

PR #790, opened 2026-06-10
feat(pcd): expose proof accessors

turbocrime:
consumers may need to handle pcd components separately.


alxiong (approved):

TalDerei (approved):
the fields behind these accessors were actually pub before API hardening in https://github.com/tachyon-zcash/ragu/pull/563 (which itself was inspired by https://github.com/tachyon-zcash/ragu/pull/541). since callers are now expanding to downstream consumers, exposing these seems reasonable. longer term we'll also want canonical serialization.

TalDerei:
status quo hasn't changed here.

PR #791, opened 2026-06-11
chore(ci): fix weekly automated fuzzing workflow

TalDerei:
References action failure in https://github.com/tachyon-zcash/ragu/actions/runs/27245117745.

After this lands on main, need to manually trigger `workflow_dispatch` to rerun job and fix our continuous fuzzing workflow.

### DIFF
--- a/.github/workflows/fuzz-cron.yml
+++ b/.github/workflows/fuzz-cron.yml
@@ -95,7 +95,12 @@ jobs:
             ${{ runner.os }}-fuzz-corpus-${{ matrix.target }}-
 
       - name: Install cargo-fuzz
-        run: cargo install cargo-fuzz
+        # Pinned version + --locked: an unlocked install re-resolves
+        # transitive deps, whose MSRV can drift past the pinned nightly
+        # (cargo-fuzz 0.13.2 pulls cargo-platform 0.3.3, which wants
+        # rustc >= 1.91). Revisit this pin when rotating the nightly in
+        # .github/actions/rust-nightly-setup.
+        run: cargo install --locked cargo-fuzz@0.13.1
 
       - name: Run fuzz target
         working-directory: qa/fuzz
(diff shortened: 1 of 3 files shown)


CodeMongerrr:
I duplicated this from the other direction before realising it was already open (#814, now closed in favour of this). Three things I turned up that might be worth having here.

**It works.** I tested the pin on the repo's pinned stable rather than assuming: `cargo install --locked cargo-fuzz@0.13.1` builds cleanly on 1.90.0. For completeness, `--locked` on its own is not enough, since cargo-fuzz 0.13.2's own lockfile already pins `cargo-platform 0.3.3`, so the version pin is doing the real work here.

**Scope, in case it is useful for prioritising.** This is not just the run linked in the description. Every scheduled run since 2026-06-10 has failed the same way, 16 in a row, last green 2026-06-07. All 17 targets, every time.

**The comment's revisit trigger may point at the wrong thing.** The note says to revisit the pin when rotating the nightly in `rust-nightly-setup`, but the install step has no `RUSTUP_TOOLCHAIN`, so it runs on the default toolchain, which is the **stable 1.90.0** from `rust-toolchain.toml`. That is also what the error reports (`rustc 1.90.0 is not supported`). The pinned nightly is currently 1.96, comfortably past `cargo-platform`'s 1.91 floor, so rotating the nightly will not affect this pin either way. The thing that would is bumping the MSRV in `rust-toolchain.toml`. Might be worth pointing the reminder there instead, or setting `RUSTUP_TOOLCHAIN` on the install step so it tracks the same toolchain the targets run on.

**Unrelated, but adjacent:** the failure is quieter than it looks. `Extract run stats` wraps its redirect in `|| true` and the crash upload is `if-no-files-found: warn`, so a run where fuzzing never started still uploads a stats artifact and reports no crashes. From the Actions list it reads a lot like a clean run, which is probably part of why this sat for eight weeks. Happy to open a separate issue for that if you think it is worth hardening.

TalDerei:
superseded by #792.

PR #792, opened 2026-06-11
chore(fuzz): cron coverage fixes and crash regression infra

TalDerei:
There were some fuzzing targets missing from the CI matrix, which this adds. Also splits the corpus cache into restore/save so it can load interesting crash inputs for future workloads.

We run these automated workflows three times a week. It already runs, `fuzz-cron.yml` which looks for bugs and mutates inputs to find crashes and grow the corpus. After this lands, it'll also run `fuzz-coverage.yml` which produces a coverage report (it replays the corpus once and reports which code / lines the fuzzers are actually reaching, so we have some stats even when the fuzzing didn't find anything!). Before this coverage report, we only collected crash artifacts when a bug was found, saved that in the corpus as memory, and then next run picks up from it.

### DIFF
--- a/.github/workflows/fuzz-coverage.yml
+++ b/.github/workflows/fuzz-coverage.yml
@@ -1,0 +1,124 @@
+name: Fuzz coverage
+
+# Weekly per-target coverage reports over the cron-accumulated corpora.
+# Replays each target's cached corpus under a coverage-instrumented
+# build and uploads an llvm-cov per-file report, so coverage drift is
+# visible over time without running anything locally. The union ("which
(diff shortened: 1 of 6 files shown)


TalDerei:
this fixes the fuzzing CI job that runs 3 times a week and has been broken for 2 months, so optimistically merging. cc @ebfull

the way this works basically is scheduled fuzz runs accumulate corupus (file of past runs), then the weekly coverage workflow consume them and upload real per-target reports. action compute costs are free since it's a public repo and we're using standard runners and we shouldn't run into runner storage limits.

Issue #793, opened 2026-06-11
Extend underconstrained fuzzing to internal recursion circuits


TalDerei:
Follow-up to https://github.com/tachyon-zcash/ragu/issues/728 and depends on https://github.com/tachyon-zcash/ragu/pull/794;

Extend the patcher technique to production circuits. **This is where the real under-constrained fuzzing checks happen against our internal recursion circuits!**

- Run the patcher against our real internal recursion circuits and more complex circuits.

  The patcher engine currently lives inside the `fuzz_advice_patcher` target and only knows how to fuzz the small generated "substrate" circuits. To aim it at the real recursion circuits we need to pull it out, teach it to find its footing on circuits it didn't generate, and give it an oracle that means something once a circuit has real public inputs and outputs.

  - [x] **Make the patcher engine reusable:** Move the recording driver, the repair solver, and the advice-finding logic out of the fuzz target and into `ragu_testing`, so it can run against *any* `Circuit` through its public API, not just the generated ones. Part of this is teaching it to work out which wires are free advice from the constraint graph itself (today the substrate just hands it that list).
  - [x] **Add a real soundness oracle:** Once a circuit has public inputs and outputs, "the same inputs always give the same outputs" is a genuine spec we get for free: pin the inputs, let a malicious prover wiggle the internal advice, repair, and flag it if an output moves while every constraint still holds. We couldn't do this on the generated circuits, they have no public I/O, so there's nothing to pin against.
  - [x] **Close the two driver gaps:** The recording driver panics on routines today and uses the simplest allocator, so the allocator's wire-pooling path never runs. The real recursion circuits use both, and they're the only place we can actually exercise them.
  - [x] **Run it on the real circuits:** Point the engine at the internal recursion circuits in `ragu_pcd` and let it hunt. If it bogs down on the larger ones, switch the solver from re-scanning every constraint each pass to a worklist that only revisits what actually changed.

- Add the new fuzz targets to `fuzz-coverage.yml`.

PR #794, opened 2026-06-11
feat(fuzz): implement the patcher technique for catching under-constrained bugs

TalDerei:
This consolidates a few different local branches that contributed to this structure. It implements the patcher technique in https://github.com/tachyon-zcash/ragu/issues/728 from [ZKSecurity's paper](https://dl.acm.org/doi/pdf/10.1145/3713081.3731718) to exercise underconstrained fuzzing. Follow-up work in https://github.com/tachyon-zcash/ragu/issues/793 extends this to internal recursion circuits; this lays the groundwork for that. 

- We _already_ had `fuzz_witness_cheat`, which simply tests the prover's computation by swapping values during witness generation and checking if the simulator misbehaves.
- `fuzz_witness_pinning` which for the first time tests the constraints themselves (takes a valid witness, tampers with a value like a malicious prover, and sees if existing constraints reject it).
- `fuzz_circuit_cheat` is a weaker patcher implementation that mutates one witness input, recompute downstream by rerunning the gadgets, assembles the trace and checks the revdot identity against a native oracle.
- `fuzz_advice_patcher` is the real patcher engine this PR closes: a dedicated recording driver captures the constraint graph, then mutates a free advice wire and repairs by walking the graph and solving for the single unknown in each constraint (not by re-running gadget logic), comparing the result to a native shadow. So it tries walking the constraint graph and solving for adjustments so the constraints remain satisfied, and that's exactly what we're trying to exploit to catch under-constrained bugs.

Also, libfuzzer and cargo test now both operate over the same shared op grammar under `ragu_testing` rather than each fuzz target having their own different grammars.


TalDerei:
Opting to merge this since changes are purely consolidated in `qa/fuzz`. cc @ebfull

PR #802, opened 2026-06-22
refactor(mock): improve alignment

turbocrime:
The mock now re-exports more actual ragu modules.

Drops the relations module as a consumer concern.

## Changes

- **Use ragu polynomials.** `Polynomial` delegates to `ragu_arithmetic`, drops the `Commitment` wrapper in favor of bare `Eq` Adds `ragu_pasta` as a `mock`-feature dep.
- **`Polynomial::MAX`** from `ProductionRank` (`1 << ProductionRank::RANK`).
- **Constraint utils** (`src/mock/constraint.rs`): free functions representing constraining methods of ragu's `Element` and `Point`
- **Poseidon sponge** (`src/mock/sponge.rs`): an out-of-circuit `Sponge` that drives the real `ragu_primitives` sponge with a wireless emulator. Replaces ad-hoc `halo2_poseidon` use in consumers.
- **Export ragu's errors.** Deletes the mock's `Error(&'static str)` and re-exports `ragu_core::{Error, Result}`
- **Drop the `relations` module.** Removes `src/mock/relations.rs` and its tests, no longer part of the mock surface.


src/mock/step.rs
```diff
@@ -70,16 +68,16 @@ impl Index {
 
     /// Parses an [`Index`] from its `get()` value.
     pub(crate) fn from_value(value: u64) -> Result<Self> {
-        let value_usize =
-            usize::try_from(value).map_err(|_err| Error("step index value exceeds usize"))?;
+        let value_usize = usize::try_from(value)
+            .map_err(|_err| Error::Initialization("step index value exceeds usize".into()))?;
```
LucidSamuel: the path that reaches this (`proof.rs:173`) otherwise uses `MalformedEncoding` for the same byte-parsing ergo `MalformedEncoding` fits better here for consistency.

src/mock/polynomial.rs
```diff
@@ -1,240 +1,70 @@
...
-    /// `commit() = ∑ coeffᵢ·gᵢ` -- the unblinded coefficient commitment.
+    /// `commit() = ∑ coeffᵢ·gᵢ` -- the unblinded coefficient commitment, the
+    /// Vesta point `ragu_arithmetic::mul` returns.
     #[must_use]
-    pub fn commit(&self) -> Commitment {
-        use pasta_curves::group::{Curve as _, Group as _};
-
-        let generators = &*GENERATORS;
+    pub fn commit(&self) -> Eq {
+        let coeffs = self.0.clone();
```
LucidSamuel: `mul` takes `iter()` by reference, so this clone might be unnecessary given `self.0.iter()`/`self.0.len()` would avoid the allocation.

src/mock/polynomial.rs
```diff
@@ -1,240 +1,70 @@
...
-    /// Evaluate via Horner's method: `p(x) = c₀ + x(c₁ + x(c₂ + …))`.
+    /// Evaluate at `x` via the real `ragu_arithmetic::eval` (Horner's method).
     #[must_use]
     pub fn eval(&self, x: Fp) -> Fp {
-        self.0.iter().rev().fold(Fp::ZERO, |acc, &c| acc * x + c)
+        ragu_arithmetic::eval(&self.0, x)
     }
 
-    /// `commit() = ∑ coeffᵢ·gᵢ` -- the unblinded coefficient commitment.
+    /// `commit() = ∑ coeffᵢ·gᵢ` -- the unblinded coefficient commitment, the
```
LucidSamuel: think we use KaTeX for inline unicode math based on conventions

src/mock/constraint.rs
```diff
@@ -0,0 +1,81 @@
...
+    } else {
+        Ok(())
+    }
+}
+
+/// Enforces `a == b` for two curve points. Mirrors `Point::enforce_equal` (real
+/// ragu's derived `GadgetEquals` for `Point`, which constrains the coordinates
+/// equal); returns `Err(Error::InvalidWitness(err))` where those constraints
+/// would be unsatisfiable (`a != b`).
+pub fn enforce_equal_point<C: Group>(a: C, b: C, err: &'static str) -> Result<()> {
```
LucidSamuel, line 80: non blocking:- `enforce_equal_point` accepts `identity == identity`, but the real `Point` gadget cannot represent the identity: `Point::alloc` and `Point::constant` reject points at infinity and since this helper is documented as mirroring `Point::enforce_equal`, maybe we add a short doc note that callers must only pass non-identity affine points when using it as a `Point` mock.

LucidSamuel (commented):
LGTM!
Nothing blocking. Left some doc asks and a few cleanups.

S1nus (approved):

src/mock/sponge.rs
```diff
@@ -0,0 +1,128 @@
+//! Out-of-circuit Poseidon sponge — mirrors `ragu_primitives::poseidon::Sponge`.
+//!
+//! Real ragu's sponge API is inherently in-circuit: every method threads a
+//! driver and operates on `Element` constraint wires, which is useless to a
+//! mock consumer that has no driver. There is also no standalone native
+//! Poseidon permutation — it exists only as an in-circuit routine. So this
+//! module runs the *real* sponge out-of-circuit by driving it through the
+//! wireless emulator (`Emulator<Wireless<Always<()>, Fp>>`), exposing a plain
+//! `Fp`-in / `Fp`-out boundary. The output is bit-identical to the in-circuit
```
alxiong: but by just feeding a wireless driver, it behaves exactly like a out-of-circuit Sponge. 

I'm fine if we add helper wrapper function to safe caller from supplying a wireless driver every time, but creating a mirror/copy feels unnecessary and potential drift. (even though you have tests)

Same thing for the routine already implementing the permutation itself.

TalDerei (approved):
optimistically to unblock things after addressing feedback; there's already API parity divergence, but we can iterate on alignment and cleaning up this crate.

src/mock/sponge.rs
turbocrime: the mock is not emulator-based, so it's not quite ergonomic to just re-export the existing poseidion utilities.

this wrapper allows mock circuits to use a thin abstraction that plays well with the rest of the mock. if it diverges, we can update the mock.

turbocrime: updated the docstring 4fa27e739dcf8d45e722f5e0aec0ab73d77fed9d

src/mock/constraint.rs
turbocrime: i've updated the docstring with a caveat 86c1066ec83968585b48411fa5d2d8cb0a86e2a0 but i think an assertion would be preferred.

if `Point` refuses identity, that might be problematic for the existing step logic. iirc we use the identity point deliberately in some situations, and some times we handle deliberately empty polys, which would commit to the identity point.

i'll investigate consumption sites, but i might want to defer adding anti-identity assertions until that's definitely resolved.

turbocrime: confirmed with bc06020b2e158c7e7e23f65196ffb8dc38072b62 that consumers fail when non-identity points are required.

turbocrime: opened https://github.com/tachyon-zcash/ragu/pull/807

PR #807, opened 2026-07-03
refactor(mock): polys and identity commits

turbocrime:
Ragu can't witness or handle the identity point. The mock accepted it everywhere, letting consumers depend on behavior real ragu forbids. <https://github.com/tachyon-zcash/ragu/pull/802#discussion_r3461366870>

## Changes

- `enforce_equal_point`, the poly-query claim sink, and `derive_challenge` return `InvalidWitness` on an identity point.
- `Header::encode` now diverges to return typed values instead of opaque bytes, so the mock system can reject identity points carried in header data.
- `Polynomial::commit` stays permissive: an empty polynomial still commits to the identity, as is permitted in ragu.
- The mock `Polynomial` now wraps `ragu_circuits::polynomials::sparse::Polynomial` instead of a bare `Vec<Fp>`, so it mirrors the real type (sparse API, `Ring` impl) rather than reimplementing things.

### DIFF
--- a/src/mock/application.rs
+++ b/src/mock/application.rs
@@ -96,6 +96,9 @@ impl Application {
         let left_proof = left.proof;
         let right_proof = right.proof;
 
+        proof::compute_header_hash(S::Left::SUFFIX, &S::Left::encode(&left.data))?;
+        proof::compute_header_hash(S::Right::SUFFIX, &S::Right::encode(&right.data))?;
+
         let mut hooks = FrameworkHooks::new();
         let mut ctx = StepCtx::new(&mut hooks);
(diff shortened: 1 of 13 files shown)


TalDerei (approved):
ACK changes isolated to mock crate; this already being used in tachyon https://github.com/tachyon-zcash/tachyon/pull/172/changes#diff-6adc9c858440eda07d9b53897a33324c0f4ebfbb4e6e31dd306f86af32c4b029R52. noted OOB that we'll eventually do a sweeping refactor of this m,ock crate before we start releasing things on [crates.io](http://crates.io/), and and it's also feature gated so it's fine. cc @ebfull

PR #809, opened 2026-07-06
chore: improve no_std

S1nus:
configure these three deps to enable no_std friendliness, unblocking ragu usage in librustzcash via zcash_tachyon


TalDerei (approved):
ACK, deferring merge to @ebfull.

ebfull (approved):

PR #810, opened 2026-07-06
chore: CI check for thumbv7em-none-eabihf no_std

### LINKED ISSUE #808: CI check for building on thumbv7em-none-eabihf no_std
librustzcash needs to build for this target. tachyon opts for a different strategy than orchard; pulling in ragu types directly rather than orchard's approach of making halo2 an optional dependency. Instead, ragu has no_std friendliness as an [explicit design goal](https://github.com/tachyon-zcash/ragu/pull/540#discussion_r2901209466).

Create a CI check similar to librustzcash for this target

S1nus:
closes https://github.com/tachyon-zcash/ragu/issues/808

after https://github.com/tachyon-zcash/ragu/pull/809

it should pass after 809

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -180,6 +180,28 @@ jobs:
       - name: Run 32-bit tests
         run: cargo test --release --all --locked --target i686-unknown-linux-gnu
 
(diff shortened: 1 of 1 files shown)


TalDerei (approved):
ACK, deferring merge to @ebfull.

ebfull (approved):

PR #813, opened 2026-08-02
feat(primitives): add Element::batch_invert

### LINKED ISSUE #292: Element::batch_invert
`Element` currently has an `invert` method that returns a multiplicative inverse of a nonzero element `a` by witnessing its inverse `b` and proving that `a * b = 1`. This requires the _prover_ to compute $b = a^{-1}$. However, we sometimes need to perform many inversions at the same time in the circuit. It would be convenient if there existed a Montgomery batch inversion API -- one that could merely use `ff`'s [`BatchInverter`](https://docs.rs/ff/0.13.1/ff/struct.BatchInverter.html) or some other API already written -- and internally the batch-inverted witness values could be fed as advice into the underlying inversion logic.

This ties into a more general need for the inversion behavior to be overridden when the caller has already computed the needed inverse somewhere else (perhaps in another circuit!), in which case they should be able to supply advice instead.

CodeMongerrr:
Closes #292.

`compute_v` already had a private `Inverter` that batch inverted its `(u - x_i)` differences with Montgomery's trick and passed the results to `invert_with` as advice. Since #292 asks for exactly that as a general API, this lifts the logic onto `Element` and points the existing caller at it, so `Inverter::invert` becomes a one-line delegation.

The issue also mentions the related need to override inversion when the caller already has the inverse from somewhere else. I think `invert_with` already covers that case, so I left it alone and built the batch API on top of it rather than duplicating the advice plumbing.

### On cost

Worth being precise, since it would be easy to read this as a circuit improvement and it isn't one. Each element still costs one gate and two constraints, exactly as `invert` does. The saving is entirely in the prover's witness computation, which goes from n field inversions to one inversion plus the batch multiplications. The doc comment and the test both say so explicitly.

### On zeros

`BatchInverter::invert_with_external_scratch` skips zero elements via `conditional_select` and leaves them as zero, so the accumulator only ever multiplies nonzero values and its `invert().unwrap()` cannot panic. A zero therefore gets zero as its advice, and the existing per-element constraint is what rejects it. Net behaviour matches `invert`: the batch fails to synthesize. `test_batch_invert` pins this for a zero in first and middle position, and covers the empty batch too.

### Checks

`cargo test --workspace` passes, including the `ragu_pcd` circuit tests that exercise the refactored `compute_v` path. `cargo clippy --workspace --lib --tests --benches -- -D warnings` and the pinned nightly `fmt --check` are both clean.

One judgement call I'd happily change: I kept the `next().unwrap()` idiom from the original code, since the iterator and the slice are built from the same source and are always the same length. If you'd rather it returned an error there, say the word.

### DIFF
--- a/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
+++ b/crates/ragu_pcd/src/internal/native/circuits/compute_v.rs
@@ -43,10 +43,10 @@
 //! [$\mu'$]: unified::Output::mu_prime
 //! [$\nu'$]: unified::Output::nu_prime
 
-use alloc::{vec, vec::Vec};
+use alloc::vec::Vec;
 use core::marker::PhantomData;
 
-use ragu_arithmetic::{Cycle, ff::Field};
+use ragu_arithmetic::Cycle;
 use ragu_circuits::{
     WithAux,
     horner::Horner,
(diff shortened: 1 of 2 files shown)


CodeMongerrr:
Heads up so the red check does not cost anyone time: `cargo audit` is failing here for a reason unrelated to this PR. It is RUSTSEC-2026-0204 in `crossbeam-epoch 0.9.18`, which is currently failing on `main` and on every open PR. I wrote up the full chain in #815, including why the obvious fix is blocked.

This branch touches two `.rs` files and nothing else, no `Cargo.lock` changes. Everything else is green, including all four `test` platforms, `no_std` on thumbv7em, both mock dependency modes, clippy and rustfmt.

TalDerei:
thanks for the PR! we're deferring these kinds of changes for now since some internals are bound to change.

PR #816, opened 2026-08-02
docs: note that circuits are compile-time and cannot be serialized

### LINKED ISSUE #118: Clarify that circuits cannot be serialized to disk
Circuits are defined at compile-time rather than runtime. Consequently, they’re compiled directly into the binary during `cargo build`, and we don’t support serializing or deserializing circuits to or from files. Likewise, we don’t support serializable verifying keys. We should document this this limitation prominently in the readme and user guide section of the book. This same limitation applies in [halo2](https://github.com/zcash/halo2).

CodeMongerrr:
Closes #118.

Two places, as the issue asked:

- **README**, a bullet under Requirements, next to the existing `no_std` note, since that section is already doing the job of "things to know before you use this".
- **The user guide**, a new section at the top of `guide/requirements.md`, above `no_std` so it is the first thing on the page after the version requirements. The core sentence is in an `admonish warning` block, matching how the introduction and the `GadgetKind` page flag their caveats.

I linked the two together with a `#compile-time-circuits` anchor, the same way the README already points at `#no-std`.

Beyond restating the issue, I added one sentence about the practical consequence: the binary carries the circuit, so a verifier has to be built from the same circuit definitions as the prover rather than loading a key file at startup. That is the part that actually bites if you have come from a proving system that ships verifying keys as data, and it seemed worth saying out loud. Happy to cut it if you would rather the docs stay closer to a plain statement of the limitation.

Also noted that halo2 has the same limitation, as the issue mentions.

Markdown only, no code touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,14 @@
 * The minimum supported [Rust](https://rust-lang.org/) version is currently **1.90.0**.
 * Ragu requires minimal dependencies and currently strives to avoid using dependencies that are not already used in [Zebra](https://github.com/ZcashFoundation/zebra).
 * Ragu's library crates are [`no_std`]-compatible and only need a global allocator; the default `multicore` crate feature uses [`maybe-rayon`] and pulls in `std`. See the [book][book-no-std] for details.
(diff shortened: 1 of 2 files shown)


TalDerei:
closing this for now since book will change; think this should be a book only change, i don't think the readme warrants this.

PR #817, opened 2026-08-05
feat(pcd): an application declares hook layout, a step witnesses polynomials

turbocrime:
A step can witness a polynomial commitment, and its proof will carry cleartext coefficients.

## Layout config

`HookConfig` carries per-step budgets as type-level lengths. The budget is
declared rather than discovered, for simplicity. The config is elaborated in
following PRs.

Parameter `J` threads this through `Application`, `ApplicationBuilder`, stages,
and circuit types.

## Step behavior

The `witness_polynomial` hook commits a polynomial to an affine point in the
scalar field, and returns a `PolyHandle`, which is a decomposition of the affine
point into two circuit field elements.

The adapter provides the step body with `StepCtx` and manages `FrameworkHooks`,
and produces padding so each step's behavior is uniform.

## Fuse behavior

Reading a fuse input back, `application_ky` folds the handle region, binding it
into the `k(Y)` claim. `nested::stages::preamble` stashes the commitments as
curve points.

`_10_p` folds each witnessed polynomial into `p(X)` and its commitment into the
endoscaled sum, so what accumulates is the commitment the step pinned.

`Proof` gains the witnessed polys' coefficients and a commitment cache.

## Other changes

- `nested::stages::preamble` gains each polynomial's commitment point
- `native::stages::eval` gains `witness_poly_evals` to challenge each at `u`
- `internal::endoscalar` types use a `Len` parameter instead of const `NUM_POINTS`
- `internal::nested` replaces `NUM_ENDOSCALING_POINTS` with const fns `child_endoscaling_points` and `num_endoscaling_points`
- `internal::nested::register_all` asserts the registry took one slot per `InternalCircuitIndex::all` entry, rather than trusting that the two agree

### DIFF
--- a/crates/ragu_pcd/benches/criterion/fuse.rs
+++ b/crates/ragu_pcd/benches/criterion/fuse.rs
@@ -10,12 +10,12 @@ fn fuse_bench(c: &mut Criterion) {
     let pasta = Pasta::baked();
     let poseidon_params = Pasta::circuit_poseidon(pasta);
 
-    let app = ApplicationBuilder::<Pasta, ProductionRank, 4>::new()
+    let app = ApplicationBuilder::<Pasta, ProductionRank, 4>::new(pasta)
         .register(nontrivial::WitnessLeaf { poseidon_params })
         .unwrap()
         .register(nontrivial::Hash2 { poseidon_params })
         .unwrap()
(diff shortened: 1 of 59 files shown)


TalDerei:

PR #818, opened 2026-08-05
feat(pcd): a step opens a polynomial, and the fuse enforces it

turbocrime:
After #817

The second hook slot. A witnessed polynomial becomes openable: a step claims a
value for it at a point, and the fuse that consumes the proof discharges the
claim.

## Layout config

`HookConfig` now also describes query capacity.

Query padding fills unused query capacity with additional openings.

## Step behavior

The `enforce_poly_query` hook records that the polynomial behind a handle
produces the value `y` when evaluated at `x`. The step proves nothing; the wires
go to the instance and the obligation travels with the proof.

The instance region grows by `poly_queries * (HANDLE_WIRES + 2)`, written after
the handles.

## Fuse behavior

Poly-queries join the batch of openings `compute_f` and `compute_v` already
discharge.

A query names its polynomial by commitment, so `compute_v` resolves it against
polynomial data in the fuse input. A commitment matching no handle is rejected.

## Other changes

- `native::stages::preamble` gains an enforced query region
- `native::circuits::poly_query` resolves each query's evaluation for `compute_v`
- `Proof` gains `application_poly_queries`

### DIFF
--- a/crates/ragu_pcd/benches/criterion/fuse.rs
+++ b/crates/ragu_pcd/benches/criterion/fuse.rs
@@ -10,12 +10,12 @@ fn fuse_bench(c: &mut Criterion) {
     let pasta = Pasta::baked();
     let poseidon_params = Pasta::circuit_poseidon(pasta);
 
-    let app = ApplicationBuilder::<Pasta, ProductionRank, 4>::new()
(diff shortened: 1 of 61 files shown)


TalDerei:

PR #819, opened 2026-08-05
feat(pcd): a step derives a challenge, confirmed later

turbocrime:
After #818

The third hook slot, and the only one that adds a stage to the native chain. A
step derives a challenge value, and a new circuit re-derives it.

## Layout config

`HookConfig` gains `ChallengeDerivations` and `ChallengeWidth`.

Step authors are responsible for constraining challenge inputs in a way that
prevents grinding.

## Step behavior

The `derive_challenge` hook absorbs whatever a gadget writes, and provides a
digest computed out-of-circuit.

Empty input positions are padded with a fixed value, so every derivation absorbs
the layout's full width.

`AdapterLen` gains `challenge_instance_len()`, written after the query region.

## Fuse behavior

The `native::circuits::challenge_binding` circuit re-derives each challenge from
the inputs, and enforces equality against the step's recorded challenge.

## The challenges stage

- `NUM_UNIFIED_CIRCUITS` increases
- `InternalCircuitIndex::NUM` and `RxIndex::NUM` both increase
- `native::stages::challenges` added, the last stage after `native::stages::outer_error`
- `native::circuits::outer_collapse` now ends on the challenges stage
- `native::stages::preamble`'s `application_ky` folds the challenges into `k(Y)`
- `nested::stages::preamble` stashes the challenges and challenge-binding points

### DIFF
--- a/crates/ragu_pcd/benches/criterion/fuse.rs
+++ b/crates/ragu_pcd/benches/criterion/fuse.rs
@@ -10,12 +10,12 @@ fn fuse_bench(c: &mut Criterion) {
     let pasta = Pasta::baked();
     let poseidon_params = Pasta::circuit_poseidon(pasta);
 
-    let app = ApplicationBuilder::<Pasta, ProductionRank, 4>::new()
+    let app = ApplicationBuilder::<Pasta, ProductionRank, 4>::new(pasta)
         .register(nontrivial::WitnessLeaf { poseidon_params })
         .unwrap()
         .register(nontrivial::Hash2 { poseidon_params })
         .unwrap()
-        .finalize(pasta)
+        .finalize()
         .unwrap();
 
     let mut rng = StdRng::seed_from_u64(1234);
--- a/crates/ragu_pcd/benches/criterion/registry.rs
+++ b/crates/ragu_pcd/benches/criterion/registry.rs
@@ -13,7 +13,7 @@ fn registry_bench(c: &mut Criterion) {
 
     // Time finalize separately: build the ApplicationBuilder, then bench only finalize.
     let make_builder = || {
-        ApplicationBuilder::<Pasta, ProductionRank, 4>::new()
+        ApplicationBuilder::<Pasta, ProductionRank, 4>::new(pasta)
             .register(nontrivial::WitnessLeaf { poseidon_params })
             .unwrap()
             .register(nontrivial::Hash2 { poseidon_params })
@@ -23,13 +23,13 @@ fn registry_bench(c: &mut Criterion) {
     c.bench_function("registry::finalize", |b| {
         b.iter_batched(
             make_builder,
-            |builder| builder.finalize(pasta).unwrap(),
+            |builder| builder.finalize().unwrap(),
             criterion::BatchSize::PerIteration,
         );
     });
 
     // Build the finalized app once for evaluation benchmarks.
-    let app = make_builder().finalize(pasta).unwrap();
+    let app = make_builder().finalize().unwrap();
     let registry = app.native_registry();
 
     // Use deterministic "random" field elements.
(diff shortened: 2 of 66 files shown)


LucidSamuel (commented):
I benchmarked the Hooks stack because the new layout turns polynomial, query, and challenge capacity into application-level choices and I wanted to be able answer two practical questions before those choices become defaults:-
what does every application pay even at `NoHooks`, and what does an application pay as it declares more capacity? unequically that should also help with budget guidance and show where optimization work would matter most.

also posting this against the pinned pre-restack heads while the per-layer attribution is still recoverable.

The main results:

- Fixed cost: before the challenge layer, I found no positive overhead of practical size. The cumulative #819 head adds **6.02% to `fuse`, 8.46% to `seed`, 7.48% to rerandomization, and 9.49/9.54% to verification at zero Hook budget. Verification has the largest fixed increase and is paid by every verifier.

- Marginal verification cost: a polynomial costs about **1.54%**, plus about **1.2%** when it crosses an endoscaling-step boundary. Query capacity is much cheaper over the tested range: about **0.09%** for the first slot and **0.07%** for each after.
- Challenges have two costs: at fixed width 2, an added call costs roughly 10M instructions; at one call, each added input slot costs roughly **2.5M**. I did not vary width at multiple call counts, so their interaction is still unmeasured.
-`fuse` is only partly modeled: its challenge-width slope is consistent across the tested intervals, but polynomial and call growth do not fit stable marginal models. ~ I am leaving those unresolved rather than forcing a formula.

the measurements are reproducible: the seven fixed-cost benchmarks and the full 22-case marginal sweep each repeated exactly, to the instruction.

I also found a configuration question rather than a performance result. All tested polynomial and query budgets build, but challenge budgets reach non-obvious limits: `<0,0,1,10>` and `<0,0,3,2>` build, while `<0,0,1,12>` and `<0,0,2,4>` fail with gate-limit errors. this is a sparse probe, not a complete envelope,
and I do not want to assume the limits are bugs.

> Is there an intended supported envelope for challenge calls and width? If so, should it be documented orcvalidated when the application is built, so the failure can name the Hook budget rather than a generic circuit limit?

Separately, I independently confirmed that unused challenge-width padding allocates no wires and that isolated property holds; although it simply is not sufficient to predict end-to-end cost or application constructibility.

Full tables, pinned SHAs, derivations, and scope notes are preserved in `REPORT.md`; I can attach them if useful. the harness is preserved at `bench-harness`/`4fbffe3c`.

~ these are Callgrind instruction-count comparisons from Linux/ARM64, with baseline and candidate built and run under the same container and toolchain.

turbocrime:
@LucidSamuel

there's no constant envelope. the capacity interacts with other configuration
parameters.

`derive_challenge` actually hashes out-of-circuit, and the step just witnesses
the challenge value. that witnessed value is bound by in-circuit derivation
later in `challenge_binding`, which executes at fuse.

challenge cost grows by poseidon rate block. this means additional challenge
width is nearly free until we hit a rate boundary, so 3, 7 and 11 are the
efficient widths.

in `challenge_binding` with domain tag at `RATE=4` the permutation count is
`ceil((width + 1) / 4)` and perms cost 288 gates each. it runs for both fuse
inputs, so double it.

and `challenge_binding` is sharing gate budget with `preamble`, which needs its
own `3*HEADER_SIZE + (2*POLYS + 4*QUERIES) + NUM_WIRES+1` gates where
`NUM_WIRES=29` is some preexisting ragu stuff. so tuning header size, declaring
polynomials, or declaring poly-queries also affects the challenge capacity.

all unused capacity is padded, for uniformity. so no step can execute with an
over-capacity configuration, which is not compile-time but it is a pretty quick
failure from the developer's perspective. we could probably const-assert an
estimated success based on configuration and other parameters, but i didn't do
that during implementation since the characterization was constantly changing.

LucidSamuel:
thanks, this also reconciles the padding result... unused positions add no wires, but the declared layout still determines how many Poseidon rate blocks challenge_binding must process.

TalDerei:

PR #820, opened 2026-08-05
test(pcd): hook tests

turbocrime:
After #819

Step-level tests for hooks.


TalDerei:

PR #821, opened 2026-08-05
feat(mock): update for hooks

turbocrime:
Application hooks, end to end. This PR supersedes and collapses the stack that
was #817, #818, #819 and #820 — those branches were all based on `main`, so each
one's diff already showed the cumulative change. Their descriptions are
preserved below as per-layer sections.

The layers are still the commits, so a commit-by-commit read follows the same
order the split PRs did:

| Layer | Commits | Was |
| --- | --- | --- |
| Polynomial witnessing | `c1b3e369`, `0eeebd1f`, `7f8cbc61`, `05a45a1f` | #817 |
| Polynomial queries | `b5aa7e5b`, `057b0ce3` | #818 |
| Derived challenges | `2b758e17`, `56c832d6` | #819 |
| Hook tests | `8a2b3b1b`, `34c096f3` | #820 |
| Mock support | `8918ef84` | this PR |

---

# Layer 1 — an application declares hook layout, a step witnesses polynomials

A step can witness a polynomial commitment, and its proof will carry cleartext coefficients.

## Layout config

`HookConfig` carries per-step budgets as type-level lengths. The budget is
declared rather than discovered, for simplicity. The config is elaborated in
the layers below.

Parameter `J` threads this through `Application`, `ApplicationBuilder`, stages,
and circuit types.

## Step behavior

The `witness_polynomial` hook commits a polynomial to an affine point in the
scalar field, and returns a `PolyHandle`, which is a decomposition of the affine
point into two circuit field elements.

The adapter provides the step body with `StepCtx` and manages `FrameworkHooks`,
and produces padding so each step's behavior is uniform.

## Fuse behavior

Reading a fuse input back, `application_ky` folds the handle region, binding it
into the `k(Y)` claim. `nested::stages::preamble` stashes the commitments as
curve points.

`_10_p` folds each witnessed polynomial into `p(X)` and its commitment into the
endoscaled sum, so what accumulates is the commitment the step pinned.

`Proof` gains the witnessed polys' coefficients and a commitment cache.

## Other changes

- `nested::stages::preamble` gains each polynomial's commitment point
- `native::stages::eval` gains `witness_poly_evals` to challenge each at `u`
- `internal::endoscalar` types use a `Len` parameter instead of const `NUM_POINTS`
- `internal::nested` replaces `NUM_ENDOSCALING_POINTS` with const fns `child_endoscaling_points` and `num_endoscaling_points`
- `internal::nested::register_all` asserts the registry took one slot per `InternalCircuitIndex::all` entry, rather than trusting that the two agree

---

# Layer 2 — a step opens a polynomial, and the fuse enforces it

The second hook slot. A witnessed polynomial becomes openable: a step claims a
value for it at a point, and the fuse that consumes the proof discharges the
claim.

## Layout config

`HookConfig` now also describes query capacity.

Query padding fills unused query capacity with additional openings.

## Step behavior

The `enforce_poly_query` hook records that the polynomial behind a handle
produces the value `y` when evaluated at `x`. The step proves nothing; the wires
go to the instance and the obligation travels with the proof.

The instance region grows by `poly_queries * (HANDLE_WIRES + 2)`, written after
the handles.

## Fuse behavior

Poly-queries join the batch of openings `compute_f` and `compute_v` already
discharge.

A query names its polynomial by commitment, so `compute_v` resolves it against
polynomial data in the fuse input. A commitment matching no handle is rejected.

## Other changes

- `native::stages::preamble` gains an enforced query region
- `native::circuits::poly_query` resolves each query's evaluation for `compute_v`
- `Proof` gains `application_poly_queries`

---

# Layer 3 — a step derives a challenge, confirmed later

The third hook slot, and the only one that adds a stage to the native chain. A
step derives a challenge value, and a new circuit re-derives it.

## Layout config

`HookConfig` gains `ChallengeDerivations` and `ChallengeWidth`.

Step authors are responsible for constraining challenge inputs in a way that
prevents grinding.

## Step behavior

The `derive_challenge` hook absorbs whatever a gadget writes, and provides a
digest computed out-of-circuit.

Empty input positions are padded with a fixed value, so every derivation absorbs
the layout's full width.

`AdapterLen` gains `challenge_instance_len()`, written after the query region.

## Fuse behavior

The `native::circuits::challenge_binding` circuit re-derives each challenge from
the inputs, and enforces equality against the step's recorded challenge.

## The challenges stage

- `NUM_UNIFIED_CIRCUITS` increases
- `InternalCircuitIndex::NUM` and `RxIndex::NUM` both increase
- `native::stages::challenges` added, the last stage after `native::stages::outer_error`
- `native::circuits::outer_collapse` now ends on the challenges stage
- `native::stages::preamble`'s `application_ky` folds the challenges into `k(Y)`
- `nested::stages::preamble` stashes the challenges and challenge-binding points

## Challenge capacity

Carried over from the review discussion on #819, since it is the only
characterization of how challenge capacity behaves. It belongs in the book
eventually; it is recorded here so it is not lost with that PR.

There is no constant envelope — the capacity interacts with other configuration
parameters.

`derive_challenge` hashes out-of-circuit, and the step just witnesses the
challenge value. That witnessed value is bound by in-circuit derivation later in
`challenge_binding`, which executes at fuse.

Challenge cost grows by Poseidon rate block, so additional challenge width is
nearly free until it crosses a rate boundary. That makes 3, 7 and 11 the
efficient widths.

In `challenge_binding` with a domain tag at `RATE=4`, the permutation count is
`ceil((width + 1) / 4)`, and permutations cost 288 gates each. It runs for both
fuse inputs, so double it.

`challenge_binding` also shares gate budget with `preamble`, which needs its own
`3*HEADER_SIZE + (2*POLYS + 4*QUERIES) + NUM_WIRES+1` gates, where
`NUM_WIRES=29` is preexisting. So tuning header size, declaring polynomials, or
declaring poly-queries all affect the challenge capacity too.

All unused capacity is padded, for uniformity. No step can execute with an
over-capacity configuration — not a compile-time failure, but a quick one from
the developer's perspective. A const-assert on an estimated success from the
configuration and the other parameters is possible, but was not done during
implementation because the characterization kept changing.

On padding specifically: unused positions add no wires, but the declared layout
still determines how many Poseidon rate blocks `challenge_binding` must process.

---

# Layer 4 — hook tests

Step-level tests for hooks.

---

# Layer 5 — mock support

`mock` catches up to the implemented hooks, and re-exports their layout.


TalDerei:
this needs to be rebased.

PR #829, opened 2026-08-13
document step capacity and proving cost

### LINKED ISSUE #824: Extract step capacity and proving costs into book
We can extract the Ragu PCD step's budget and proving cost into book now that it's been calculated more explicitly. References https://github.com/tachyon-zcash/ragu/pull/720. We can also PR some version of this when we fine tune performance and need to benchmark costs.

LucidSamuel:
adds a User Guide page recording what fits in a PCD step and what proving costs. Closes #824 

Hook budgets are excluded until #817 through #821 lands.


TalDerei (approved):
ACK purely additive

PR #831, opened 2026-08-15
book: desc public inputs and stage outputs

LucidSamuel:
- explains public inputs, stage outputs, and staging circuit outputs.
- document the three recursive (k(Y)) evaluations and substitution protection.
- updates stale recursion circuit and stage names.


TalDerei (approved):

TalDerei:
reACK corrects stale code prose in the book.

PR #837, opened 2026-08-17
chore(fuzz): enable routines and allocator pool in patcher driver

LucidSamuel:
This closes the third bullet point in https://github.com/tachyon-zcash/ragu/issues/793. It addresses the routine-support gap only; pooled allocation, automatic advice discovery, and production recursion-circuit targets will be follow-up work.

- removes the `Recorder` and `Playback` routine overrides that panic.
- let both drivers use the existing `Driver::routine()` default implementation.
- adds a regression test covering prediction-derived auxiliary data, constraint recording, honest playback, and rejection of a corrupted routine output.


TalDerei:
pushed an update; this now closes both the driver routine and allocator gaps in our fuzzing recorder driver.

TalDerei (approved):

TalDerei (approved):
reACK

PR #839, opened 2026-08-17
feat(backend): add reference and accelerated backend interfaces

TalDerei:
References https://github.com/tachyon-zcash/ragu/issues/834 (design motivation in https://github.com/tachyon-zcash/ragu/issues/834#issuecomment-5320700832) and implements the first two checklist items in https://github.com/tachyon-zcash/ragu/issues/834#issuecomment-5320749799.

Adds a `Backend` interface and threads a sealed backend selection through `ragu_pcd`, defaulting to `ReferenceBackend` and switchable to `AcceleratedBackend` (currently a placeholder that inherits every default). 

The backend choice covers both proving and verification, but verifier trust stays separate. `SelectableBackend::Verifier` can accelerate the verification kernels, while `AcceleratedProver` speeds up proving and keeps verification on the reference kernels.

Granular breakdown:
  - `Backend` is the monolithic trait with default methods.
  - `ReferenceBackend` inherits every default.
  - `AcceleratedBackend` selectively overrides methods (none yet).
  - `AcceleratedProver` is `AcceleratedBackend` for proving, `ReferenceBackend` for verification.
  - `SelectableBackend` seals which complete ragu backend implementations applications may select.

The backend is currently selected application-wide, it is not application-specific (which seems weird). Later, app specific optimizations can be exposed through [constrained hooks](https://github.com/tachyon-zcash/ragu/pull/839#issuecomment-5405260610), while backend selection remains sealed.

Equivalence tests (reference vs. accelerated vs. canonical) land in the follow-up https://github.com/tachyon-zcash/ragu/pull/841, so this PR is the backend boundary only.

Issue #834, opened 2026-08-16
AI-assisted prover optimization loop


TalDerei:
Inspired by https://blog.zksecurity.xyz/posts/optimizing-cryptography-with-ai/ and this podcast episode on [flock](https://eprint.iacr.org/2026/1329) https://zeroknowledge.fm/podcast/404/, we should _eventually_ freeze our prover implementation and engage in optimization loops that purely optimize the prover. As long as the verifier is correct, we're satisfied with the soundness security of our protocol.

**Workflows:**

1. Freeze the current prover as a reference.
2. Record its RNG consumption and produce canonical proof digests.
3. Let AI propose one optimization at a time.
4. For each dynamically generated case, run the reference and optimized provers with identical replayed RNG input, then require their canonical proof digests to match.
5. Run verifier, completeness, fuzzing, and formal verification checks.
6. Keep the change only if end-to-end benchmarks improve.

**Optimization areas:**

- Cache optimizations
- SIMD vector instructions
- GPU acceleration using Metal
- Assembly optimizations
- Hardware-accelerated instructions for hashing
- Pipelining workloads

TalDerei:
- `ragu_pcd`: shared protocol types and public façade.
  - `ragu_verifier`: minimal, auditable soundness-critical code.
  - `ragu_prover`: replaceable, aggressively optimized proving code.

TalDerei:
References https://github.com/tachyon-zcash/ragu/pull/836. attaching another supplemental brainstorm based on OOB discussion: https://gist.github.com/TalDerei/3c377453058b2995e1a16f1b5e3c8470.

TalDerei:
After ruminating on OOB discussion and distilling my notes:

An application can either select: (1) ragu's reference backend with a stable interface, (2) AI-assisted accelerated backend of that reference implementation. Applications can then provide specialized backend overrides/hooks that the backends can use. 

- `ragu_backend`: Arkworks-style monolithic trait with defaults (eg. MSM, s-polynomial and registry evaluation, out-of-circuit Poseidon, and constrained floor-planning hooks).
- `ragu_acceleration`: implements that trait's associated types and can override inherited methods.
- `ragu_pcd`: canonical protocol orchestration.
- root ragu crate exposes the stable application API and does backend selection.

[...]

PR #840, opened 2026-08-18
feat(backend): route more operations through backends

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/839 and implements the second checklist item in https://github.com/tachyon-zcash/ragu/issues/834#issuecomment-5320749799. It extends #839 to:

- Sparse polynomial evaluation and `revdot`
- $s(X,y)$, $s(x,y)$, and registry evaluation
- Polynomial commitments

The backend is currently selected application-wide, it is not application-specific (which seems weird). It has a sealed backend that defaults to `ReferenceBackend` and can be switched to `AcceleratedBackend`. Later, app specific optimizations can be exposed through [constrained hooks](https://github.com/tachyon-zcash/ragu/pull/840#issuecomment-5334922924), while backend selection remains sealed. 

Granular breakdown:
  - `Backend` is the monolithic trait with default methods.
  - `ReferenceBackend` inherits every default.
  - `AcceleratedBackend` selectively overrides methods.
  - `SelectableBackend` only seals which complete ragu backend implementations applications may select.


TalDerei:
because these are sealed traits, the idea is that an app could _eventually_ implement specific hooks/overrides like through a separate registration API, not by implementing `Backend`:

```ts
  ApplicationBuilder::new()
      .with_backend::<AcceleratedBackend>()
      .register_with_acceleration(
          step,
          StepAcceleration::new()
              .placement(proposal)
              .s_poly_evaluator(my_evaluator)
              .native_poseidon(my_poseidon),
      );
```

crates/ragu_pcd/src/lib.rs
```diff
@@ -51,13 +53,37 @@ use step::{Step, internal::adapter::Adapter};
...
+    /// A Ragu-owned computational backend.
+    ///
+    /// This trait is sealed: applications may select one of Ragu's supported
+    /// implementations, but cannot provide their own backend implementation.
+    pub trait SelectableBackend: Backend + sealed::Sealed {}
+
+    impl<T: Backend + sealed::Sealed> SelectableBackend for T {}
+}
+
+pub use selectable_backend::SelectableBackend;
```
TalDerei, line 73: sealed trait interface.

crates/ragu_pcd/src/backend_tests.rs
TalDerei: these are ephemeral routing tests that sanity-check things are properly routed through the PCD call stack; they'll be removed once we have stronger equivalence tests between the reference and acceleration backends.

PR #841, opened 2026-08-19
feat(backend): expand equivalence test harness

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/840 and implements the third checklist item in https://github.com/tachyon-zcash/ragu/issues/834#issuecomment-5320749799.

Adds property tests comparing the reference, accelerated, and canonical implementations across MSM, polynomial, registry, commitment, and failure behavior (including intentionally planted faults to prove the tests work). It also compares all three PCD backends on seeded leaf / fused proofs, and checks the verifier decisions for valid and corrupted proofs.

PR #842, opened 2026-08-19
feat(acceleration): accelerated override for the MSM backend 

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/841 and implements the fourth checklist item in https://github.com/tachyon-zcash/ragu/issues/834#issuecomment-5320749799.

Currently targets [1.0.0-rc.2](https://crates.io/crates/zakura-pasta-curves/1.0.0-rc.2) to demonstrate using the `zakura_pasta_curves`, and MSM performance will improve by ~40% for 2^13 MSM once the open stack of upstream MSM optimizations lands and we update to another RC. Under current RC, benchmarks are registering ~5% improvement because those other optimizations haven't landed yet.

To demonstrate acceleration backend overrides, we're calling zakura's `zakura-pasta-curves` stack in https://github.com/zakura-core/libraries. 

It uses hand-written AArch64 / x86_64 field assembly, runtime BMI2/ADX selection and other MSM optimizations that match the latest [Zprize](https://github.com/z-prize) performance. It's a CPU/assembly backend, but doesn't have other optimizations like:

- Metal/GPU
- explicit vector SIMD (build even disables AVX)
- persistent cross-call MSM cache
- asynchronous/threaded pipeline


.github/workflows/rust.yml
```diff
@@ -103,6 +103,50 @@ jobs:
       - name: Check runtime crates with legacy dependencies
         run: cargo check -p ragu_arithmetic -p ragu_pasta -p ragu_core -p ragu_primitives -p ragu_circuits -p ragu_backend -p ragu_acceleration -p ragu_pcd --lib --locked --no-default-features --features "alloc legacy-deps"
 
+  # The native MSM is opt-in and target-specific, so the default jobs do not
+  # compile or exercise it. Test every supported host ABI, both dependency
+  # stacks on Linux, and the end-to-end backend boundary on the modern stack.
```
TalDerei: ABI stuff is temporary since zakura libs need to update to`ff/group 0.14` to avoid this incompat, so this will disappear.

LucidSamuel:
Nice work expanding `arb_msm_size()` through `ProductionRank`, it fixes the 255-term cap I initially suspected.

I played around with Scribe by encoding the exact size support of this PR's strategy as a discrimination corpus, then validating the corresponding mutants against HEAD (48433931) using its actual native-MSM acceptance commands.

A narrower deterministic gap remained. above `TestRank::num_coeffs()` (128), `arb_msm_size()` samples powers of two ±1, while Zakura's `best_multiexp` changes Booth-window width at

```
4, 32, 55, 149, 404, 1097, 2981, 8104
```

(⌊e^k⌋+1, since it picks `c = ln(n).ceil()` for n ≥ 32 — [`zakura-halo2-proofs` `src/arithmetic.rs:254`](https://docs.rs/zakura-halo2-proofs/1.0.0-rc.2/src/halo2_proofs/arithmetic.rs.html#254)) ergo none of these are powers of two.

So I planted a one-line sentinel defect that returns an incorrect accelerated result only when `coeffs.len() == 8104` and both the direct Pallas/Vesta MSM suite and the end-to-end PCD backend-equivalence suite remained green.

The controls distinguish this from a dormant override:

- The same defect at n = 8192 is caught by the existing end-to-end proof-digest comparison.
- Instrumented proving reaches MSM sizes 8190 and 8192.
- Supplying n = 8104 directly rejects the mutant immediately.

so the accelerated route is live; the gap is specifically that power-of-two-biased sampling doesn't cover the optimized dependency's actual decision boundaries.

then I added a deterministic test covering both sides of every Zakura transition on both Pasta curves, plus `ProductionRank::num_coeffs()`:

```
3/4, 31/32, 54/55, 148/149, 403/404, 1096/1097, 2980/2981, 8103/8104, 8192
```
https://github.com/LucidSamuel/ragu/commit/46cee3d96a99055eb25ef9e0afdeea6180ce64b0

ebfull (approved):
ACK, but haven't checked for minimal dependency bumps.

Note that whenever I have too many #[cfg(feature(whatever))] I tend to interpret it as "hey maybe i need to refactor the code so that there are fewer of those" (and usually it encourages me to place those attributes on modules, and so other kinds of changes to the code.

LucidSamuel:
took a look at the dep bumps @ebfull mentioned and the good news is nothing we already depend on moves, everything new is just what `zakura-halo2-proofs` drags in, plus `cc/find-msvc-tools/shlex` for building the assembly.

a few things could be improved though but they're all zakura's side rather than this PR, for example, we now have `maybe-rayon` twice, our git fork and zakura's crates.io copy, we could `[patch.crates-io]` to the fork so there's only one, or maybe the fork isn't needed anymore...

zakura also pulls `indexmap 1.9.3` (and an old hashbrown) next to our 2.13 but that should go away whenever they do the ff/group 0.14 update you mentioned and `tracing` should be non-optional over there, which accounts for 4 of the 11 new crates...

~ also happy to do a cleanup as a follow-up once this lands that puts the cfg on `mod msm` with the same-signature fallback module, so the trait impls wouldn't need cfg blocks at all. figured that might be worth doing before the next override copies the pattern.

TalDerei:
this accidentally closed because the base branch was deleted, retargeting main.

TalDerei:
@ebfull bumping to https://crates.io/crates/zakura-pasta-curves/1.1.0.

TalDerei:
> Note that whenever I have too many #[cfg(feature(whatever))] I tend to interpret it as "hey maybe i need to refactor the code so that there are fewer of those" (and usually it encourages me to place those attributes on modules, and so other kinds of changes to the code.

yeah this was always intended to be refactored, it was just kept simple for the pace of the experiment. refactoring now to only have two top-level cfg gates.

PR #843, opened 2026-08-19
chore(fuzz): migrate patcher engine to ragu testing

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/837 and closes the first bullet point in https://github.com/tachyon-zcash/ragu/issues/793. 

It (1) migrates our patcher engine (patcher is a technique from [zksecurity's paper](https://arxiv.org/abs/2504.14881) that lets us catch under-constrained circuit bugs where you basically mutate a witness, then repair the rest through the constraints so it still satisfies them) from `fuzz_target` into `ragu_testing` and (2) adds an entry point so the engine can be pointed at any `Circuit` implementation .

The patcher loop runs like this: the **harness** (`fuzz_target`) picks a circuit and cheats a wire, the **engine** (`ragu_testing::patcher`) records the constraints, repairs the rest of the witness through them, and reports whether they still hold, and the **harness** again compares that against what the values should be and flags the bug if the circuit accepted the lie.

### DIFF
--- a/crates/ragu_testing/src/lib.rs
+++ b/crates/ragu_testing/src/lib.rs
@@ -1,4 +1,14 @@
-//! Test fixtures for Ragu crates.
+//! # `ragu_testing`
+//!
+//! This crate contains test fixtures and harnesses for the Ragu project. This
+//! API is re-exported (as necessary) in other crates and so this crate is only
+//! intended to be used internally by Ragu.
+
+#![forbid(unsafe_code)]
+#![deny(rustdoc::broken_intra_doc_links)]
+#![deny(missing_docs)]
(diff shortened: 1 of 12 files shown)


LucidSamuel (approved):
LGTM just a fit nits: 
`ragu_testing` has no publish = false (unlike qa/fuzz) and no lint header, so this makes the patcher semver-relevant public API in the one crate that enforces neither missing_docs nor forbid(unsafe_code). probably worth deciding now rather than post-release — publish = false or a lint header, either way.

minor: `let pivot_row = rows[r].clone()` in the Gauss–Jordan loop is a per-pivot allocation the old index loop didn't need since `split_at_mut` keeps the iterator form without it.

PR #845, opened 2026-08-20
chore(deps): migrate pasta_curves to zakura library 

TalDerei:
Migrates `pasta_curves` to Zakura's fork, consumed as the published [`zakura-pasta-curves 1.0.0-rc.2`](https://crates.io/crates/zakura-pasta-curves) release.


qa/fuzz/Cargo.toml
```diff
@@ -33,7 +33,7 @@ arbitrary = { version = "1", features = ["derive"] }
 # the repo-root dependency set.
```
ebfull: no more source pins!

ebfull (approved):

qa/fuzz/Cargo.toml
TalDerei: fixed comment

PR #846, opened 2026-08-20
chore(fuzz): patcher determinism oracle

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/843 and closes the second bullet point in https://github.com/tachyon-zcash/ragu/issues/793.

This patcher oracle checks that a circuit's outputs are actually determined by its inputs.

### DIFF
--- a/crates/ragu_testing/src/patcher/circuit.rs
+++ b/crates/ragu_testing/src/patcher/circuit.rs
@@ -237,8 +237,12 @@ mod tests {
     ///   engine records the pooled allocator's auxiliary constraint.
     ///
     /// The healthy `MySimpleCircuit`, with its two private witnesses
-    /// declared, sweeps clean — there the same waste-`b` cheat is rejected,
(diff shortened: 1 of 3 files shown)


LucidSamuel (commented):
Nice.

MySimpleCircuit half: (pinned, rejected) == (0, 1) is vacuous by the module's own docs, not clean as the one cheatable wire was rejected, so maybe keep the assertion; the "sweeps clean" wording could also just say it's vacuous, or use a circuit with pinned > 0.

nits: `Clone` on `ProbeOutcome/SweepReport/Violation` looks unused and only `violation.witness` is cloned. And `determinism_probe`'s d`ebug_assert!(constraints_hold(...))` re-checks the same honest witness every probe, O(wires × events) in debug; fine now, less so when #847 points this at recursion circuits. Hoisting it to determinism_sweep keeps the guarantee.

PR #852, opened 2026-08-22
Msm window boundaries

LucidSamuel:
Stacked on #842 (`accelerated-msm-zakura`), adds one deterministic test on top of it. 
Draft: intended to be merged into that branch or cherry-picked (`46cee3d9`), whichever is easier; not a standalone change.

-the gap

above `TestRank::num_coeffs()` (128), `arb_msm_size()` samples powers of two ±1. Zakura's `best_multiexp` picks its Booth-window width as `c = ln(n).ceil()` for n ≥ 32 ([`zakura-halo2-proofs` `src/arithmetic.rs:254`](https://docs.rs/zakura-halo2-proofs/1.0.0-rc.2/src/halo2_proofs/arithmetic.rs.html#254)), so its behavior changes at

```
4, 32, 55, 149, 404, 1097, 2981, 8104
```

(⌊e^k⌋+1) none of which are powers of two ergo the strategy cannot land on any transition above 128, so a defect confined to one window configuration passes the suite.

Demonstrated, with controls: a planted defect firing only at `coeffs.len() == 8104` keeps the direct MSM suite and the end-to-end PCD backend-equivalence suite green, while the same defect at n = 8192 is caught by the existing proof-digest comparison (so the accelerated route is live and the harness works... the sampling just never reaches the dependency's decision boundaries). Full write-up with the mutation evidence in the #842 thread.

-the fix

One deterministic test covering both sides of every transition on both Pasta curves, plus `ProductionRank::num_coeffs()`:

```
3/4, 31/32, 54/55, 148/149, 403/404, 1096/1097, 2980/2981, 8103/8104, 8192
```

-verification

- Rejects the 8104 sentinel mutant; passes on clean #842 HEAD (`48433931`)
- Full MSM suite, PCD backend tests, `cargo fmt`, and clippy with `-D warnings` all pass
- Runs in ~4.5s (release)


TalDerei:
thanks! since we can't stack this PR, cherry-picked and coauthored in https://github.com/tachyon-zcash/ragu/pull/842.

PR #853, opened 2026-08-23
chore(fv): move lean extraction module under qa/lean

TalDerei:
Consolidating FV under single `qa/fv` path. Also introduces and `pre-flight` skill file from ironwood https://github.com/zcash/ironwood/tree/main/.claude/skills.


TalDerei:
Superseded by #854 — same branch, now pushed directly to tachyon-zcash/ragu instead of the fork.

PR #857, opened 2026-08-23
chore(primitives): poseidon test vectors and parameter consistency

TalDerei:
We originally tested poseidon permutation consistency against Halo2 in https://github.com/tachyon-zcash/ragu/pull/200, and we should now enshrine these test vectors in the codebase as we start formally verifying our poseidon implementation since this is a gap our formal guarantees can't capture (which ironwood also will do in https://github.com/zcash/ironwood/issues/204). 

Some independent tests this runs to check parameter consistency:
- Like #200, we transplant Ragu's permutation against Halo2's published vectors at Halo2's T = 3 parameters,
- Regenerate and check parameters for Ragu's T = 5 with an independent python script,
- A one-time sage run that sanity checks that our local T = 5 parameters are equal to the actual sage output in Daira's sage script https://github.com/daira/pasta-hadeshash.


crates/ragu_primitives/src/poseidon/tests/halo2_vectors.rs
TalDerei, line 1: lifts fixtures from halo2's `poseidon/tests/halo2_vectors.rs` instead of using a dev-dep.

PR #864, opened 2026-08-25
chore(fv): gadget contracts and randomized polynomial checks

TalDerei:
Stacks on https://github.com/tachyon-zcash/ragu/pull/862.

Hardens the composable gadget contracts and fingerprinting via the random polynomial evaluation approach in https://gist.github.com/TalDerei/f00c974c5ec9cd28120409a02841f964.

Today, the Clean framework already has a pretty good compositional functionality for gadgets (reference https://github.com/tachyon-zcash/ragu/issues/759):

- **gadget-level composition**: a parent calls `Child.Circuit` and Clean embeds the child's operations, outputs, local allocation, and offset wiring into the parent;
- **theorem-level composition**: the same `Child.Circuit` carries soundness, and Clean's subcircuit machinery applies it, so the parent gets `Child.Spec`.

Importantly, in Clean the parent proof consumes child `Assumption --> Spec contracts` (what that means is the parent proves the child's assumptions and that the embedded child constraints hold, the child soundness theorem then gives the child spec, and the parent reasons from that spec). There were some gadgets that weren't composing well, so this fixes them.

This also adds an _additive_ polynomial evaluation driver (runs the real Rust gadget code through the normal `Driver` API), and compares the random evaluation that driver produces against Lean, which independently computes the same evaluation from `FormalInstance.reimplementation`. It derives a seed, hashes it, then does a hash-to-field operation to derive the challenge. Again this doesn't replace their extraction driver, it just adds another independent testing path for now.


qa/fv/extraction/src/instance.rs
```diff
@@ -1,86 +1,85 @@
...
+pub trait InstanceDriver<'dr>: Driver<'dr> {
+    /// Allocate `n` verifier-visible input coordinates.
+    fn alloc_input_wires(&mut self, n: usize) -> Vec<Self::Wire>;
+}
+
+impl<'dr, F: PrimeField> InstanceDriver<'dr> for ExtractionDriver<F> {
+    fn alloc_input_wires(&mut self, n: usize) -> Vec<Self::Wire> {
+        ExtractionDriver::alloc_input_wires(self, n)
+    }
+}
```
TalDerei, line 31: wrappers now accept `generic D: InstanceDriver` and return `D::Wire` which lets the same real gadget invocation run under both `ExtractionDriver` and `EvaluationDriver`.

qa/fv/extraction/src/instance.rs
```diff
@@ -134,13 +129,71 @@ pub trait CircuitInstance {
...
         crate::fingerprint::digest_hex::<Self::Field>(trace.input_len, &trace.ops, &trace.outputs)
     }
+
+    /// Evaluate the production four-slot gate relation directly at `points`
+    /// independently derived challenge vectors.
+    fn polynomial_record(
+        instance: &str,
+        seed: [u8; 32],
+        points: usize,
+    ) -> core::result::Result<Record, String> {
```
TalDerei, line 144: old fingerprint hash with new polynomial evaluation.

qa/fv/extraction/src/main.rs
```diff
@@ -69,7 +70,21 @@ struct ExportTarget {
...
+    Exact,
+    Polynomial {
+        seed: [u8; 32],
+        points: usize,
+    },
+    #[cfg(test)]
+    Differential {
+        seed: [u8; 32],
+        points: usize,
+    },
```
TalDerei, line 87: all 3 modes run in CI.

qa/fv/Ragu/PolynomialFingerprint/Main.lean
TalDerei, line 1: all gadget comparison happens in CI by running this and diffing its 53 records against Rust.

qa/fv/Ragu/PolynomialFingerprint.lean
TalDerei, line 1: independent Lean impl of the randomized evaluation protocol.

PR #868, opened 2026-09-02
chore: use release of maybe-rayon fork

S1nus:
avoid using dependencies from git by pulling the maybe-rayon fork from crates.io


TalDerei (approved):
reviewed vendored https://github.com/tachyon-zcash/maybe-rayon and release in https://crates.io/crates/tachyon-maybe-rayon/0.1.2. this is safe to land, cc @ebfull

PR #869, opened 2026-09-03
mock: less

turbocrime:
this PR deletes a good chunk of the mock.


Cargo.toml
```diff
@@ -24,31 +24,24 @@ default = ["multicore"]
...
   "dep:ragu_core",
   "dep:ragu_pasta",
-  "dep:ragu_pcd",
-  "dep:ragu_primitives",
-]
-multicore = [
-  "ragu_arithmetic?/multicore",
-  "ragu_circuits?/multicore",
-  "ragu_pcd?/multicore",
 ]
```
TalDerei, line 37: why are we modifying this?

TalDerei: oh also because we're removing their crates as root deps.

TalDerei (approved):
optimistic ACK

PR #872, opened 2026-09-06
rename workflows

TalDerei:


crates/ragu_core/src/parameters/pasta/mod.rs
TalDerei: should most of `crates/ragu_core/src/parameters/pasta/` move into udon? migrated some things into core temporarily here to justify removing the `ragu_pasta` crate.

Issue #748, opened 2026-06-01
Tracking: Recursion Implementation (Split-Accumulation)


TalDerei:
Split accumulation tasks:

- [x] Dual pipeline #749
- [x] Challenge smuggling circuit #750
- [x] Mirror circuits #751
- [x] Endoscaling on native curve to close #752
- [x] Verifier gaps to close https://github.com/tachyon-zcash/ragu/issues/298

cc @ebfull, @alxiong

TalDerei:
closed by #873.

PR #873, opened 2026-09-10
chore(pcd): close recursion

TalDerei:
**Human Summary:**

Split-accumulation tracking issue in https://github.com/tachyon-zcash/ragu/issues/748. 

**Experimental** draft i've been working through to trace the remaining pieces of the construction, and this may be split up and replaced. This is my understanding of the remaining pieces left to sufficiently close recursion (folding and PCS batching on nested side in the prover, mirror nested circuits to actually check that, challenge smuggling the remaining challenges using the cheap endomorphism, endoscaling on native side, and the closing verifier decision gaps).

It's easier to review this commit by commit.

---------

**AI Summary:**

1. [`8d4d8194`](https://github.com/tachyon-zcash/ragu/commit/8d4d81946638134d95942539f1ffe020211d32a9) **Fold the children's nested claims prover-side** — The parent collects both children's nested claims and folds them into `(a_n, b_n)`. The fold's error terms use the existing bridge stages, and its challenges are endoscalar lifts of the native challenges. At this point, this is prover computation only; later circuits check it.

2. [`7fd4539a`](https://github.com/tachyon-zcash/ragu/commit/7fd4539aba8b555e27f519e53884fa6f313d8f85) **Batch the nested polynomial openings prover-side** — Defines one ordered batch of nested polynomial openings, then uses it to construct the quotient polynomial `f_n`, the batched polynomial `p_n`, and its commitment `P_n`. The query and evaluation bridge stages carry the values needed to check this later.

3. [`6bdf96a1`](https://github.com/tachyon-zcash/ragu/commit/6bdf96a1009e0a05821a52fabfe077647280afce) **Bind the lifted nested challenges to the native transcript** — Adds nested stages for the lifted challenges and native circuits that recompute their commitments from the transcript challenges. Beta is handled separately by a parent's `bind_beta` circuit because it is sampled after the evaluation stage is committed.

4. [`cff7b3d3`](https://github.com/tachyon-zcash/ragu/commit/cff7b3d371787b36efa8d046054b56335dbb5db1) **Export the nested unified instance and fold it in the parent** — Introduces a nested unified instance containing `c_n`, `v_n`, the lifted challenges, and exported native commitments. The export circuit ties that instance to the current proof's stages, and a parent copies the instance into its preamble. This replaces the old copying circuits and retained child-stage polynomials; the next commit checks those copies through the recursive fold.

5. [`6168fc0f`](https://github.com/tachyon-zcash/ragu/commit/6168fc0fd7a8d9c29cb01bb11bb9c04af7cfa148) **Verify the nested fold and batch evaluation in-circuit** — Adds a nested `collapse` circuit that recomputes each child's `k(y_n)` from the unified-instance values copied into the parent's preamble, checks both folding layers, and enforces `c_n` outside the base case. A nested `compute_v` circuit recomputes `v_n = p_n(u_n)` and connects the endoscalar bits to the lifted beta.

6. [`dac7b30c`](https://github.com/tachyon-zcash/ragu/commit/dac7b30c1d4a66c678e4c78498165f518169a47c) **Walk the nested batch's commitments on the native side** — Adds native endoscaling circuits that perform the Horner walk over the nested batch's commitments, producing the endpoint intended to equal `P_n`. `bind_endoscalar` ties the walk's bits to `pre_beta`, and the number of steps is fixed from the batch sizes on both sides. Later commits bind the endpoint and the recursively supplied commitment inputs.

7. [`0d1cd29a`](https://github.com/tachyon-zcash/ragu/commit/0d1cd29a504894854d0d21cea2e9c5c1b3df1f1c) **Make the verifier rederive proof data it previously trusted** — The verifier now recomputes cached commitments, transcript challenges, derived bridge data, and staged registry evaluations instead of trusting the values stored in the proof. It also makes the trivial proof's `P` and `P_n` genuine commitments to their polynomials.

8. [`7f1196b7`](https://github.com/tachyon-zcash/ragu/commit/7f1196b7826626ae71798f9ee78110fb8f2c0ece) **Enforce curve membership and bit constraints for both endoscaling walks** — Checks that every point used by either endoscaling walk lies on its curve and that each endoscalar-stage wire is a bit. These checks live in nested `export` and native `bind_endoscalar`, where they can be enforced directly.

9. [`56b05c10`](https://github.com/tachyon-zcash/ragu/commit/56b05c109b6ff57f4ba6e07d661db78ac51c25cd) **Bind the nested batch's commitments before their transcript challenges** — Splits the native point data into stages committed before the challenges that depend on it. It also merges beta into the nested challenge stage and has the parent's `bind_beta` complete and check the exported challenge commitment.

10. [`425e408f`](https://github.com/tachyon-zcash/ragu/commit/425e408f3ad9d414ca0f6bdb343f7279ff472682) **Pin the walked `P_n` to the unified instance** — Makes the child's exported `P_n` equal the final point produced by its native endoscaling walk. The parent's `bind_beta` circuit then checks that it consumes that same exported point.

11. [`01138b5d`](https://github.com/tachyon-zcash/ragu/commit/01138b5de035c5850a67a30cfbf664f50508433b) **Bind nested walk commitments through recursion** — Extends the parent's `bind_beta` circuit so the bridge commitments it walks match the child's transcript-bound bridge polynomials, while the walked `A_n`, `B_n`, and `registry_xy` commitments match the child's exported native-instance slots. The decider also compares the corresponding points in the native stages with the verified polynomial commitments.

12. [`60d825a2`](https://github.com/tachyon-zcash/ragu/commit/60d825a27c01da74b0b8bdbd3265ff8d6f7b89ae) **Support one-point endoscaling walks while retaining the final equality** — Allows the Horner walk to contain only its initial point while still enforcing that the final output equals that point, with a regression test for the equality check.


TalDerei:
@v12sec review this

TalDerei:
@v12sec review this

TalDerei:
my human summary of the commits:

**1. Folding the children’s nested claims prover side:** (1) the prover is folding children nested revdot claims (endoscaling circuit SAT, wellformedness and consistency stage masks, and accumulator claims) from step i - 1 in step i (we don't enforce nested fold in circuit yet), (2) it introduces nested revdot error terms that are carried in the existing nested bridge stages, and (3) derives the nested folding challenges from the native challenges using the same endoscalar lift pattern as beta.  

**2. Batch the nested openings prover side**: (1) opening claims over polynomials (some of the polynomials there are child B’s own bridge polynomials) produce quotients, batched with alpha into a batched quotient polynomial f_n. Then those polynomials and f_n are batched with beta into p_n. 

**3. Bind nested challenges to the transcript:** the prover derived nested challenges from native transcript challenges using endoscalar lifts, and now we’re adding the corresponding nested stages and native circuits checks that check the nested challenges were derived correctly, and we do that by taking a pallas commitment to the nested stage, witnessing that in the circuit, and then recomputing the commitment in the circuit and checjing that they match. 

**4. Give the nested circuits shared public inputs:** creates nested unified instance that contains accumulator, batched evaluation, lifted challenges, native commitments all produced in the current step i (and child data feed the computations). Unified instances are created in the current proof step and coped to the proof at the end. Also, there’s a new export circuit connecting things stored in nested stages in step i to the unified instance created in step i who's fields eventually get stored in the proof. We also removed the copying circuits (which checked the fields stored in the parents preamble stage match what the child actually stored in their stages) in favor if export circuit which checks that the child B’s unified instance with the values stored in B’s own nested stages. The next commit has circuit checks that compute k(y) from child unified instance. So to clarify about the copying circuits, (1) before there were child stage polynomials (grandchildren from perspective of parent) that the parent stored and copying circuit compared every field against what it stored in it's own before, and (2) now in the child stage, export circuit checks that fields against the child's unified instance, parent copies it, and then a collapse circuit copies those instance fields and computes k_y and checks the nested fold using that k_y. 

**5. Verify the nested fold and batch evaluation in circuit:** two new nested circuits, collapse that checks revdot folding and computes expected k_y and c_n, and compute_v which checks batch evaluation by computed v = P(u). It also performs nested base case detection. 

**6. Walk the nested batch’s commitments on the native side:** adding 24 native endoscaling circuits that check MSM / Horner walk over the nested batch pallas commitments, mirroring the existed nested endoscaling circuits of vesta commitments, also adds a circuit that checks that the native commitment walk uses the current steps transcript-derived beta. Adding 24 native endosclaign circuits also adds 14 additional nested endoscaling circuits. 

**7. Redrive what the decider used to trust:** verifier not rederives commitments against polynomials, absorbed bridge commitments and squeezes transcript challenges, reconstructs bridge polunomials from native commitments, etc. rather than trusting the proof. 

**8. Enforce the walks' point and bit requirements:** nested export and native bind_endoscalar circuits checks that endoscaling walk’s points lie on the curve via membership checks and booleanity that it’s endoscalar wires are bits. 

**9. Commit the nested points before their challenges:** we had native point stages holding nested batch’s pallas commitments, but we generated those pallas commitments too late in terms of the transcript binding, so we need to separate them into multiple point stages and their Vesta commitments need to enter the appropriate earlier bridges along the native batch’s commitments. 

**10. Pin the walked P_n to the unified instance:** ensures P_n a parent consumes it eh point computed by the child’s native endoscaling walk.

**11. Bind nested walk commitments through recursion:** connect more child’s nested commitments to the copies the parent consumes. These are pallas commitments to nested Fq polynomials which are consumed by native endoscaling circuits. 

**The rest of the commits below are regression testing.**

TalDerei:
not going to force push to this branch anymore, just appending commits (ideally just doc comments and regression tests from audit runs).

TalDerei:
need to fix some of these proptests, they're taking way too long in CI.

crates/ragu_pcd/src/internal/stage_wires.rs
```diff
@@ -0,0 +1,211 @@
...
+//! polynomial.
+//!
+//! A stage's rx polynomial carries its values unblinded: `StageExt::rx` puts
+//! the blinding `alpha` at the SYSTEM gate alone and the values, two per
+//! gate, at the $a$ and $d$ wires of the stage's reserved gates in
+//! allocation order. So a value can be named by its reservation index (the
+//! order the stage's output gadget traverses its wires) and read straight
+//! off the polynomial at the degree that index maps to. The verifier uses
+//! this to hold what the stages claim against what it can recompute, and
+//! the patcher harness to declare a circuit's outputs.
```
ebfull: Curious why this isn't inside of `ragu_circuits`.

ebfull (approved):

crates/ragu_pcd/src/internal/stage_wires.rs
TalDerei: fixing, this was actually organizationally intended to live in `ragu_circuits` since they're generic helpers.

TalDerei: https://github.com/tachyon-zcash/ragu/pull/873/commits/a71a9c4eafe903672962855d9d4048c23c09525d

